PR #24, opened 2017-12-24
Serbian translation (sr and sr@latin)

Faenriis:
Please merge, thank you.


andyholmes:
Cool, thanks for the first translation. Using your translation in the extension should be pretty easy until I get a release posted, but I'll try and push an update to the WebExtension so you can start using that.

PR #39, opened 2018-02-02
Correctly handle Seek in MPRIS plugin

getzze:
Due to a bug (I don't know who is to blame, DBUS or the players), the `Position` property is always 0, but it can be retrieved from the more annoying `org.freedesktop.DBus.Properties` interface.


andyholmes:
Thanks, this great. I wondered why that wasn't working but hadn't looked into it. I'm going to merge this  and #40 into master, although I'm working on branch for v10 that's diverged a lot, I'll manually implement these there.

PR #46, opened 2018-02-07
Use Caribou to generate keyevents on X11

getzze:
Using Caribou is better than atspi, although it also only work on X. You can use modifiers and non-ascii characters (there is a bug in caribou with gnome, so not all unicode characters are doable, though. I hope they can solve it).
It would be great to add a Preference option to use or not Caribou and test if the library is available in this preference dialog and not in mousepad.js


andyholmes:
This is great stuff, thanks. Sorry for the slow response, my free time has been a bit fragmented and I don't like reading PR's on my phone.

PR #50, opened 2018-02-20
Added italian translation

amivaleo:
I hope you will appreciate this!

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -1,3 +1,4 @@
 sr
 sr@latin
 pl_PL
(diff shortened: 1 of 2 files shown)


andyholmes:
Thanks, that's great. Currently GSConnect is in a rewrite cycle, but much of the translation will still work when v10 is released. Until then, I hope you can atleast use it locally :)

amivaleo:
I'll keep using your code. I like to have my phone connected to my computer.
I will continue to help with the italian translation. ;)

Issue #48, opened 2018-02-16
Unable to browse SD card


didierga:
My phone is a Motorola Moto G 3nd Gen running Android 6.0.1 and KDE Connect 1.7.1

Using MConnect and KDE-Connect I can browse both internal storage and SD card

Unfortunately using GSConnect 9 I can browse internal storage but  I cannot browse SD card.

andyholmes:
Hi, I don't have a device with an SDCard myself, however all browseable directories are reported by the app itself.

Could you tell me in what way you can't browse? I the SDCard not visible in the list of directories the shell extension offers? Or does it just fail to browse when you open it in nautilus?

didierga:
I have 5 choices in GSConnect "browse files" menu (translated from french)
- SD Card
- Pictures from camera (SD Card)
- Internal storage
- Pictures from camera (Internal storage)
- Unmount

When I click on "Internal storage" or "Pictures from camera (Internal storage)" I open a Nautilus window

When I click "SD Card" or "Pictures from camera (SD Card)" nothing happen but I have messages in log. See attached file. 

Command "tree -d /run/user/1000/gsconnect/"' shows only internal storage and no the SD card.

[gsconnect#48.log](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/1733723/gsconnect.48.log)

andyholmes:
Sorry I haven't had time to come back to this. I have a feeling this may be related to PR #40 which was submitted by @getzze, although I'm unsure. Something that might help is if you could post the debug message from the app which would look something like this:

```js
{
    "id": 1518956413092,
    "type":"kdeconnect.sftp",
    "body": {
        "ip": "192.168.1.71",
        "port": 1743,
        "user": "kdeconnect",
        "password": "UzcNCrI7T668JyxUFjOxQncBPNcO",
        "path": "/storage/emulated/0",
        "multiPaths": ["/storage/emulated/0","/storage/emulated/0/DCIM/Camera"],
        "pathNames":["All files","Camera pictures"]
    }
}
```

didierga:
I hope it's that you need.....


JS LOG: [gsconnect@andyholmes.github.io]: SFTP: mount()
JS LOG: [gsconnect@andyholmes.github.io]: LanChannel.send(adb7ba6ff2123456, {"id":0,"type":"kdeconnect.sftp.request","body":{"startBrowsing":true}})
[...]

PR #54, opened 2018-02-27
Allow sftp from sdcard (from kdeconnect code)

getzze:
This is the same code as implemented in c++ by kdeconnect. You can browse the internal and sdcard. Tested on Android N.
It improves PR #40 and it should solve issue #48.


andyholmes:
Thanks so much for the quick turn around on this, I haven't a device with an SD card so I've never seen that output before.

PR #55, opened 2018-03-03
Update Italian translation

amivaleo:
Hi! I've just updated the italian translation.
Only one line was missing respect to the previous version.
Now everything should be up to date.


andyholmes:
Cool, thanks a lot for the quick response!

PR #57, opened 2018-03-04
Add French translation

mkey-cdx:
This is a full French translation. I hope you'll enjoy it :)


andyholmes:
Thanks so much for the quick translation!

PR #79, opened 2018-04-10
Allow tel: URIs to be shared

mavit:
This allows `tel:` URIs to be passed to a phone from the Share File/URL dialog.  This is a first step towards allowing `tel:` URIs clicked in a browser to be passed directly to the phone.


andyholmes:
Thanks

PR #87, opened 2018-05-02
cs_CZ.po

psimacek:
Czech localization for GSConnect.


psimacek:
Hi Andy, 

not repo, I'm confused :-) I don't know, how to do it. Github is very confusing for me. I'm able to translate on something like Transifex etc., which is "foolproof". I'll welcome, if you help me a bit more.

Thanks for superb **GSConnect**!

andyholmes:
That's okay, I think it's easier for me to fix. Basically what happened was the template file `org.gnome.Shell.Extensions.GSConnect.pot` was renamed to `cs_CZ.po` instead of being copied. If you're trying to doing this in the Github website interface it's a little bit confusing, so not really your fault :)

PR #177, opened 2018-08-25
Fix mousepad plugin on HiDPI displays

### LINKED ISSUE #160: Remote Input broken on HiDPI screens
I have KDE installed on my laptop and phone, they are paired and options are enabled.
When using remote control, i can move the mouse, but when clicking or scrolling it doesn't work as expected.
clicking behavior: pointer jumps to top left then clicks (checked on xev as well)
scrolling: pointer moves to the right top corner

same behavior with different phones.

laptop used: chromebook pixel 2013
OS: Ubuntu 18.04

phone1: Samsung Note 8
phone2: OnePlus 3

flipflop97:
This fixes remote input functionality as discussed in #160.

At-spi uses absolute positions and therefore requires its arguments to be multiplied with the current monitors scale factor.

closes #160

### DIFF
--- a/src/service/plugins/mousepad.js
+++ b/src/service/plugins/mousepad.js
@@ -130,7 +130,9 @@ var Plugin = new Lang.Class({
 
         try {
             let [screen, x, y] = this._pointer.get_position();
-            Atspi.generate_mouse_event(x, y, event);
+            let monitor = this._display.get_monitor(screen.get_monitor_at_point(x, y));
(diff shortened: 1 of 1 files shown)


andyholmes:
I'm going to merge this, so I remember and you can get credit, but I'll be merging the wip-rewrite branch soon which has a lot of changes so I'll have to forward-port your changes after that.

PR #179, opened 2018-08-26
Fixed the issue with mouse clicks moving on high dpi screens, also fi…

danieldeng2:
Fixed the issue with mouse clicks moving on high dpi screens (issue #160), also fixed the issue of not being able to control VLC media player when paused.

### DIFF
--- a/src/service/plugins/mpris.js
+++ b/src/service/plugins/mpris.js
@@ -344,7 +344,7 @@ var Plugin = new Lang.Class({
                 pos: Math.round( pos / 1000),
                 isPlaying: (player.PlaybackStatus === "Playing"),
                 canPause: (player.CanPause === true),
-                canPlay: (player.CanPlay === true),
+                canPlay: (player.CanPlay === true || player.PlaybackStatus === "Paused"),
                 canGoNext: (player.CanGoNext === true),
                 canGoPrevious: (player.CanGoPrevious === true),
                 canSeek: (player.CanSeek === true)
@@ -387,4 +387,3 @@ var Plugin = new Lang.Class({
         PluginsBase.Plugin.prototype.destroy.call(this);
     }
 });
-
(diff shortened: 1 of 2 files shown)


flipflop97:
We've fixed the same thing in exactly the same way apparently.

#177

danieldeng2:
@flipflop97 I saw your original post, didn't know you also fixed it lol

src/service/plugins/mpris.js
```diff
@@ -344,7 +344,7 @@ var Plugin = new Lang.Class({
                 pos: Math.round( pos / 1000),
                 isPlaying: (player.PlaybackStatus === "Playing"),
                 canPause: (player.CanPause === true),
-                canPlay: (player.CanPlay === true),
+                canPlay: (player.CanPlay === true || player.PlaybackStatus === "Paused"),
```
andyholmes: I think this should be filed as an upstream bug with VLC. `CanPlay` is a [mandatory property](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Property:CanPlay) of the MPRIS specification and implementing workarounds for specific players is not a path I want to go down. KDE Connect also uses [MPRIS in the almost the exact same way](https://github.com/KDE/kdeconnect-kde/blob/master/plugins/mpriscontrol/mpriscontrolplugin.cpp#L137-L207) (`CanPlay` means the method `Play()` can be invoked) so I think it's best to stick with that.

andyholmes:
Aside from the note I left, I'm going to merge #177 since it came first (I'll also be leaving a note on that PR)

src/service/plugins/mpris.js
danieldeng2: Ok thanks I respect your decision

danieldeng2:
@andyholmes Also, the seek bar on android defaults back to 0:00 after setting position.

PR #184, opened 2018-09-03
Fix Dutch language code

piotrdrag:


Vistaus:
No, it should be nl_NL to differentiate it from nl_BE (Flemish Dutch is different at some points than Dutch from the Netherlands).

piotrdrag:
Are you going to provide nl_BE translations? Do you know of any project that includes nl_BE? GNOME doesn’t.

Vistaus:
I could name plenty of projects that use nl_BE. I could provide an nl_BE translation of gnome-shell-extension-gsconnect as I also have experience with the Flemish language. That's actually a good idea, 'cause else someone from Belgium could modify the general nl translation to adapt it for Belgian speakers, but then Dutch people will have a hard time understanding.

Vistaus:
@piotrdrag I've added a Flemish (nl_BE) translation now: https://github.com/andyholmes/gnome-shell-extension-gsconnect/pull/187

PR #189, opened 2018-09-13
Fix removal of widget property

ofuangka:
Fixes "no property widget" error when opening preferences in Gnome 3.30.  Hopefully there are no side effects...


andyholmes:
Thanks, I'm going to merge both your PR's so they're in git but unfortunately I won't be packaging another release until v13 is merged and stable.

Issue #186, opened 2018-09-10
unable to load up in gnome 3.30


Tids:
GSConnect is not able to start when using the latest gnome 3.30

```
Gio.DBusError: Fehler beim Aufruf von StartServiceByName für org.gnome.Shell.Extensions.GSConnect: GDBus.Error:org.freedesktop.DBus.Error.Spawn.ChildExited: Process org.gnome.Shell.Extensions.GSConnect exited with status 1

Stack trace:
  _init@/home/fabi/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/client.js:33:9
  wrapper@resource:///org/gnome/gjs/modules/_legacy.js:82:22
  _parent@resource:///org/gnome/gjs/modules/_legacy.js:39:12
  _init@/home/fabi/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/client.js:695:9
  wrapper@resource:///org/gnome/gjs/modules/_legacy.js:82:22
  _init@/home/fabi/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/prefs.js:100:23
  wrapper@resource:///org/gnome/gjs/modules/_legacy.js:82:22
  buildPrefsWidget@/home/fabi/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/prefs.js:215:23
  _selectExtension@resource:///org/gnome/shell/extensionPrefs/main.js:92:22
  wrapper@resource:///org/gnome/gjs/modules/_legacy.js:82:22
  _onCommandLine@resource:///org/gnome/shell/extensionPrefs/main.js:244:17
  wrapper@resource:///org/gnome/gjs/modules/_legacy.js:82:22
  main@resource:///org/gnome/shell/extensionPrefs/main.js:398:5
  @<main>:1:43
```
is what I get

Kerrung:
I have the same issue

andyholmes:
I believe these issues may have been solved by @ofuangka latest contributions. Unfortunately there won't be another formal release until after v13 has been merged and stabilized. Here is a Zip built from master which may work with Gnome Shell 3.30, although I have not tested it:

[gsconnect@andyholmes.github.io.zip](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/2377746/gsconnect.andyholmes.github.io.zip)

andyholmes:
[...]

PR #193, opened 2018-09-20
Correct usage of toString on Uint8Array

tingvarsson:
According to warning when using Gnome 3.30:
"Some code called array.toString() on a Uint8Array instance. Previously this would have interpreted the bytes of the array as a string, but that is nonstandard. In the future this will return the bytes as comma-separated digits. For the time being, the old behavior has been preserved, but please fix your code anyway to explicitly call ByteArray.toString(array)."

Related to #186

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -3,6 +3,7 @@
 const Lang = imports.lang;
 const Format = imports.format;
 const Gettext = imports.gettext;
+const ByteArray = imports.byteArray;
 
 const Gdk = imports.gi.Gdk;
 const Gio = imports.gi.Gio;
@@ -13,7 +14,10 @@ const Gtk = imports.gi.Gtk;
 ext.app_id = "org.gnome.Shell.Extensions.GSConnect";
 ext.app_path = "/org/gnome/Shell/Extensions/GSConnect";
 
-ext.metadata = JSON.parse(GLib.file_get_contents(ext.datadir + "/metadata.json")[1]);
+ext.metadata = JSON.parse(
+    ByteArray.toString(
+        GLib.file_get_contents(ext.datadir + "/metadata.json")[1]
+    ));
 ext.localedir = GLib.build_filenamev([ext.datadir, "locale"]);
 Gettext.bindtextdomain(ext.app_id, ext.localedir);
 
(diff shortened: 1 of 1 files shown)


src/common.js
```diff
@@ -13,7 +14,10 @@ const Gtk = imports.gi.Gtk;
 ext.app_id = "org.gnome.Shell.Extensions.GSConnect";
 ext.app_path = "/org/gnome/Shell/Extensions/GSConnect";
 
-ext.metadata = JSON.parse(GLib.file_get_contents(ext.datadir + "/metadata.json")[1]);
+ext.metadata = JSON.parse(
+    ByteArray.toString(
+        GLib.file_get_contents(ext.datadir + "/metadata.json")[1]
+    ));
```
andyholmes: Pretty much what should happen here is what gnome-shell is doing for byteArray:
```js
let metadata = GLib.file_get_contents(ext.datadir + "/metadata.json")[1];
gsconnect.metadata = JSON.parse(
    (metadata instanceof Uint8Array) ? imports.byteArray.toString(metadata) : metadata.toString()
);
```

Basically check each returned "bytearray" to see if it's a `Uint8Array` or not. Probably the backwards compatible version is faster since it's called automatically in the C++, but this is the only way to be backwards and forwards compatible I think.

src/common.js
```diff
@@ -53,10 +57,12 @@ ext.resource._register();
...
-        ext.app_path + "/" + ext.app_id + ".xml",
-        0
-    ).toArray().toString()
+    ByteArray.toString(
+        Gio.resources_lookup_data(
+             ext.app_path + "/" + ext.app_id + ".xml",
+             0
+        ).toArray()
+    )
 );
```
andyholmes: You could just include a function since it's used a few places:

```js
gsconnect.get_resource = function(rel_path) {
    let array = Gio.resources_lookup_data(
        GLib.build_filenamev([gsconnect.app_path, rel_path]),
        Gio.ResourceLookupFlags.NONE
    ).toArray();

    if (array instanceof Uint8Array) {
        array = imports.byteArray.toString(data);
    } else {
        array = array.toString();
    }

    return array.replace('@DATADIR@', ext.datadir);
};
```

tingvarsson: That was easy enough with a if/else, good to know. And good thinking with the function.
Want me to cleanup the PR or scrap it? Either way works for me. Since you're confident in v13 I dont feel huge need to rely on legacy branch anyway. :)

andyholmes: it's up to you, they're just warnings anyways and I think anyone with 3.28+ will end up updating before it's completely deprecated.

tingvarsson: True, will scrap it since I've moved on to master/v13 regardless :)

PR #197, opened 2018-09-22
Make GSchemas path configurable

jtojnar:


andyholmes:
Great, thanks.

jtojnar:
I think this needs a post-install script as well.

PR #201, opened 2018-09-22
NixOS compatibility fixes

jtojnar:


andyholmes:
Looks good. `post-install.sh` is causing some problems for other distros, so I'm also going to push a change to make it opt-in (eg `-Dpost_install=true`)

PR #209, opened 2018-09-25
Correct keyboard modifiers

getzze:
The switch block was not working, it seems every case was evaluating to true.
Tested with the new debugger feature that is awesome!

### DIFF
--- a/src/service/plugins/mousepad.js
+++ b/src/service/plugins/mousepad.js
@@ -116,16 +116,19 @@ var Plugin = GObject.registerClass({
                     // Set Gdk.ModifierType
                     let mask = 0;
 
-                    switch (true) {
-                        case input.ctrl:
-                            mask |= Gdk.ModifierType.CONTROL_MASK;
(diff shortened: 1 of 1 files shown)


andyholmes:
Ah good catch. Do you remember specifically what Caribou does? It's been deprecated in Gnome so it should be replaced if possible. I'll be raising the minimum gnome-shell version to 3.28 which is the first without it I think.

getzze:
It's a virtual keyboard for X11. It works with unicode and it can emulate
key modifiers like Ctrl, Super, Alt. The On-Screen keyboard for gnome >3.28
cannot and last time I tried it was not working with unicode either.
Caribou does not work on Wayland and it will probably never, I guess that's
why they deprecated it.

Le mer. 26 sept. 2018 à 04:07, Andy Holmes <notifications@github.com> a
écrit :

getzze:
Maybe the Caribou optional dependency could be added in the panel Other/Additional Features

PR #261, opened 2018-10-25
Update French translation

mkey-cdx:
Hello Andy, I'm a bit late for v13 release but there an update for the french translation.


andyholmes:
That's great, and no problem. I think we'll be doing another release quite soon and there will be another 3-4 strings to translate, but I'll merge this now and ping you all on the translations thread.

Issue #258, opened 2018-10-25
Configuration menu out of sync with enabled plugins


RenWal:
I just updated to the newest version available via the official Gnome Extensions plugin. There seems to be an issue with the plugin selection.

Steps to reproduce:

- Go to the settings for a device and hit "Advanced", then enable the notifications plugin if not already enabled and restart the service.
- Now go back into "Advanced" and disable the notifications plugin. Notice that the notifications configuration tab in the selection on the left did not disappear.
- Re-enable the notifications plugin. Notice that the notifications configuration tab has now disappeared.

To me it looks like the show/hide logic for the plugin configuration tabs is inverted, so enabling a plugin actually disables its settings tab.

Restarting the service or rebooting resolves the issue.

andyholmes:
Good catch. You right about the cause and it's just a missing `!`:

https://github.com/andyholmes/gnome-shell-extension-gsconnect/blob/48e89d862b6a83cb2acbc4a768e913c0e0a71e41/src/service/ui/settings.js#L1264

Should be:

```js
           this[name].visible = !disabled.includes(name); 
```

I have some pending commits locally, but I'll try and get this in by the end of day. If it's a problem you now, you can obviously just add that exclamation point and restart again. Thankfully this should be somewhat rare to encounter.

ferdnyc:
I've submitted the above PR with that fix.

PR #271, opened 2018-10-27
Fix visibility of plugin settings

ferdnyc:
Invert the logic for plugin-settings visibility, as described by @andyholmes in #258.

Fixes: #258


andyholmes:
Thanks, I goofed my local repo so I couldn't push this one until I finish some work one the menu UI.

PR #275, opened 2018-10-27
Update PackageKit req's for nautilus

ferdnyc:
I already updated the wiki, but figured I'd submit the source change as well:

* Fedora doesn't even have a `nautilus-python`, that's provided by `python2-nautilus`.
* Gentoo doesn't appear to have a `nautilus-extensions`, seems that's Fedora-only.
* (Also, it _seems_ like [Gentoo's package is called `pygobject`][1], not `python2-gobject`, so I also added that just in case.)

[1]: https://packages.gentoo.org/packages/dev-python/pygobject


andyholmes:
Thanks, I was using pkgs.org to try and get most of those and apparently some of them are dated or not correct.

PR #282, opened 2018-10-28
Rework PackageKit dependency management

### LINKED ISSUE #263: Does not detect if dependencies are already installed
## What happens

When you try to install one thing, it tells you in the title text it is already installed:
grafik

And in my case, everything is installed.

## What should happen 

Show a checkmark or so for "already installed".

## Possibly solution
AFAIK querying the current installation state of a package is possible without root access, so AFAIK it should work fine.

## System 

v13 (upgraded from before)
GNOME 3.28
Fedora 28

### LINKED ISSUE #273: PackageKit install buttons unfocus to "on" state
Like so many of the issues I open, an admittedly-minor look & feel issue.

On Fedora 28, when the settings window loses focus, the install buttons in Additional Features defocus to an "enabled" / positive state, rather than a deactivated / temporarily-disabled state as one might expect. 

Since the install button is a control rather than a switch, it doesn't make sense to show this kind of state:

screenshot from 2018-10-27 05-05-07

(Compare to the controls in other panes, especially the Service pane's Restart Service button):

screenshot from 2018-10-27 05-04-50

screenshot from 2018-10-27 05-04-59

Note: It's possible #263 is in play here, since the detection of installed packages doesn't seem to be working properly on Fedora. I'd be curious to learn how the unfocused state of these controls is rendered on systems where the PackageKit integration is working as expected.

ferdnyc:
* Remove the install buttons' `suggested-action` styling
* Refactor `_getInstallable()` / `_getAvailable()` into `_query(names, filter)`
* Update the installed-packages discovery logic, as outlined [in this comment](https://github.com/andyholmes/gnome-shell-extension-gsconnect/issues/263#issuecomment-433662336).

Fixes #263, fixes #273

### DIFF
--- a/src/service/ui/packagekit.js
+++ b/src/service/ui/packagekit.js
@@ -265,8 +265,6 @@ var DependencyButton = GObject.registerClass({
     }
 
     _done(icon_name=null) {
-        this._button.get_style_context().remove_class('suggested-action');
-
         this._button.visible = false;
         this._spinner.visible = false;
 
@@ -276,8 +274,6 @@ var DependencyButton = GObject.registerClass({
     }
 
     _warning(e) {
-        this._button.get_style_context().remove_class('suggested-action');
-
         this._result.visible = false;
         this._spinner.visible = false;
 
@@ -294,7 +290,6 @@ var DependencyButton = GObject.registerClass({
 
         this._button.visible = true;
         this._button.image.icon_name = 'folder-download-symbolic';
-        this._button.get_style_context().add_class('suggested-action');
         this.tooltip_markup = this._packages.map(pkg => {
             return `<b>${pkg.get_name()}</b> - ${pkg.get_summary()}`
         }).join('\n');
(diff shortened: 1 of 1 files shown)


andyholmes:
Nice.

ferdnyc:
Whoops! You snuck it in before I added one last commit. :laughing: I'll just push that one directly to master, it's just cleanup.

ferdnyc:
Hmm, maybe I can't do that. Opened #283 instead.

ferdnyc:
> Hmm, maybe I can't do that.

...Because I should check my email more often.

andyholmes:
That's weird, push access didn't work? Oh maybe you were on your fork? I wonder if I can make it so you can open forks somehow.

andyholmes:
Since you have push access like Bertrand, I think it makes sense to add you to Authors too; do you want your e-mail omitted?

ferdnyc:
Nah, I just hadn't accepted the invitation yet. It's Sunday, I'm slow.

Authors is cool, thanks. Email can go in, no worries.

Issue #268, opened 2018-10-26
Must not link to Google Play (only)


rugk:
I see v13 now links to Google Play (labelled with "Android") and Salfish OS.

So now two problems:
* You may not want to promote Google Play.
* KDEConnect is also available on F-Droid.

Solution:
Of course you could just add two links, but really a more visually pleasing solution would be to just add these two **download buttons** the same as you did for the browser add-ons.

Both Google Play and F-Droid have these badges, so you may use them. :smiley:

ferdnyc:
Fixed in #285

PR #285, opened 2018-10-29
Settings: Update add-on/app links

ferdnyc:
Taking @rugk's points in #268 to heart, I added a link to F-Droid as an app source, alongside Play Store. While I was in there I did a bit of a refresh:
* Added a "Browser Add-Ons" heading over the store buttons
* Changed "KDE Connect" to a heading
* Replaced the `GtkLinkButton` links with a single `GtkLabel` that uses Pango markup to create the hyperlinks
* Changed "Android" link to "Play Store"
* Added "F-Droid" link as an alternative Android source
* Increased default height to `580px`, to accomodate

The window now looks like this:

image

The markup has one down side, which is that the translatable string includes that markup. The _escaped_ markup. The string that translators will receive is, in full:
```html
Android (&lt;a href="https://play.google.com/store/apps/details?id=org.kde.kdeconnect_tp"&gt;
Play Store&lt;/a&gt; / &lt;a href="https://f-droid.org/packages/org.kde.kdeconnect_tp/"&gt;
F-Droid&lt;/a&gt;), &lt;a href="https://openrepos.net/content/piggz/kde-connect"&gt;
Sailfish OS&lt;/a&gt;
```

If that's a problem (too ugly for translators), an [oft-suggested alternative](https://stackoverflow.com/questions/4640084/django-templates-best-practice-for-translating-text-block-with-html-in-it) is to construct the label from a placeholder-laced string instead, via `.format()`. (This would involve moving the label text into the JavaScript source.) So, e.g.:
```js
widget.set_label(_('Android (%sPlay Store%s, %sF-Droid%s), %sSailfish OS%s').format(...));
```
However, this will break if the order of the HTML fragments changes, which I have to believe it would when dealing with translations in RTL languages. It'd be better if more expressive placeholder strings could be used.
```js
// i.e.
_('Android ({play-link}Play Store{lend}, {fdroid-link}F-Droid{lend}...').format(...);
// or even...
_('Android ({0}Play Store</a>, {1}F-Droid</a>), {2}Sailfish OS</a>').format(...)
```
 I don't know the translation system in use, is that possible?

(With all due respect to @rugk, I think using app-link _buttons_ would be overkill, and a waste of screen real estate because **nobody** will actually be using these — they'll search for "KDE Connect" in their device's applications directory, where they should be getting the app.)

### DIFF
--- a/data/settings.ui
+++ b/data/settings.ui
@@ -5,7 +5,7 @@
   <template class="GSConnectSettingsWindow" parent="GtkApplicationWindow">
     <property name="can_focus">False</property>
     <property name="default_width">640</property>
-    <property name="default_height">440</property>
+    <property name="default_height">580</property>
     <child type="titlebar">
       <object class="GtkHeaderBar" id="headerbar">
         <property name="visible">True</property>
@@ -845,7 +845,7 @@
                           </object>
                           <packing>
                             <property name="left_attach">0</property>
-                            <property name="top_attach">2</property>
+                            <property name="top_attach">3</property>
                           </packing>
                         </child>
                         <child>
@@ -873,66 +873,7 @@
                           </object>
                           <packing>
                             <property name="left_attach">1</property>
-                            <property name="top_attach">2</property>
-                          </packing>
-                        </child>
-                        <child>
-                          <object class="GtkGrid">
-                            <property name="visible">True</property>
-                            <property name="can_focus">False</property>
-                            <property name="column_spacing">12</property>
-                            <child>
-                              <object class="GtkLinkButton">
-                                <property name="label" translatable="yes">Android</property>
-                                <property name="visible">True</property>
-                                <property name="can_focus">True</property>
-                                <property name="receives_default">True</property>
-                                <property name="halign">start</property>
-                                <property name="relief">none</property>
-                                <property name="uri">https://play.google.com/store/apps/details?id=org.kde.kdeconnect_tp</property>
-                              </object>
-                              <packing>
-                                <property name="left_attach">1</property>
-                                <property name="top_attach">0</property>
-                              </packing>
-                            </child>
-                            <child>
-                              <object class="GtkLinkButton">
-                                <property name="label" translatable="yes">Sailfish OS</property>
-                                <property name="visible">True</property>
-                                <property name="can_focus">True</property>
-                                <property name="receives_default">True</property>
-                                <property name="halign">start</property>
-                                <property name="relief">none</property>
-                                <property name="uri">https://openrepos.net/content/piggz/kde-connect</property>
-                              </object>
-                              <packing>
-                                <property name="left_attach">2</property>
-                                <property name="top_attach">0</property>
-                              </packing>
-                            </child>
-                            <child>
-                              <object class="GtkLabel">
-                                <property name="height_request">32</property>
-                                <property name="visible">True</property>
-                                <property name="can_focus">False</property>
-                                <property name="halign">start</property>
-                                <property name="valign">center</property>
-                                <property name="label" translatable="yes">KDE Connect</property>
-                                <attributes>
-                                  <attribute name="weight" value="bold"/>
-                                </attributes>
-                              </object>
-                              <packing>
-                                <property name="left_attach">0</property>
-                                <property name="top_attach">0</property>
-                              </packing>
-                            </child>
-                          </object>
-                          <packing>
-                            <property name="left_attach">0</property>
                             <property name="top_attach">3</property>
-                            <property name="width">2</property>
                           </packing>
                         </child>
                         <child>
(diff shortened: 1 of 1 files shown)


andyholmes:
Actually I came across the guidelines for [GNOME Translations](https://wiki.gnome.org/TranslationProject/DevGuidelines) and looked a few other places before. I believe that it's common practice not translate trademarks or product names unless the "owner" has their own translation.

A quick grep of current translations shows that no one has been translating these (Android, Sailfish) anyways so maybe in this case we should just drop the translation strings here. Another reason this might be a good idea is that if these links are exposed to translators there's a possibility someone could maliciously change them, and I don't generally make a habit of reviewing the content of translations.

Thoughts?

ferdnyc:
Hm. Interesting! Well, that would make life easier. Curious to hear if any users familiar with interfaces in other languages have any input on that, though. It's completely outside my experience.

While I kind of hate the `GtkLinkButton` setup from a _layout_ perspective (and hate how Glade handles them even more), maybe in this case it is the better way to go. Then those links are just three separate (invisible) buttons; avoids the formatting hassle of a single string. Still not sure how to cleanly group the first two links together as separate Android-related alternatives, vs. the Sailfish link which is separate from them both.

andyholmes:
Yeah, other than the size difference, there's also no official OpenRepos.net badge for the Sailfish link. I think the way you did it works pretty well, and like you said most people probably know where/how to get the Android app anyways.

Another thing I noticed is that some themes make GtkLinkButton's look like actualy buttons or they take a lot of space that isn't visibly associated with the link itself, until maybe you hover it.

ferdnyc:
> Another reason this might be a good idea is that if these links are exposed to translators there's a possibility someone could maliciously change them, and I don't generally make a habit of reviewing the content of translations.

\*nod\* Which is another advantage to either of these formats,
```js
// i.e.
_('Android ({play-link}Play Store{lend}, {fdroid-link}F-Droid{lend}...').format(...);
// or even...
_('Android ({0}Play Store</a>, {1}F-Droid</a>), {2}Sailfish OS</a>').format(...)
```
The URL (and the link tags in general) aren't in the translation file, just their placeholders. Sure, I suppose someone could still maliciously _mislabel_ one of the links (say, make "Play Store" link to F-Droid, and vice versa), but if a translator decides to start maliciously mistranslating app strings they could do plenty of damage regardless. Malicious URL injection would be a bigger concern (and a more likely occurrence), and placeholders for markup solve that.



> Another thing I noticed is that some themes make GtkLinkButton's look like actualy buttons or they take a lot of space that isn't visibly associated with the link itself, until maybe you hover it.

That was one of the reasons I switched to a label with markup, I was initially trying to "assemble" the same basic text out of separate labels and buttons in a horizontal grid, like
> `Android (` `[Play Store]` ` / ` `[F-Droid]` `), `&nbsp;`[Sailfish OS]` 

It was just... bad. And even more untranslatable.

At one point I'd also changed the "**KDE Connect**" heading over those links to "**KDE Connect Device App**", to make it a bit clearer what exactly is being linked to. (Especially since "KDE Connect" is also the name of the _desktop_ app.) But then I second-guessed myself and changed it back, because it's a translatable string. Wondering if you have any thoughts on that?

andyholmes:
> Wondering if you have any thoughts on that?

I think at one point I was using a phrase like "KDE Connect for Other Devices" which might do it. It's also pretty important that _their_ dev team is clearly given credit. The Play Store badges or links without proper context seem like "Get GSConnect for Android" and I think a lot of people are now being introduced to KDE Connect through GSConnect or don't really understand the distinction.

> The URL (and the link tags in general) aren't in the translation file, just their placeholders.

Another reason to avoid the translations here, especially for using templates and especially is a note I buried in CONTRIBUTING:

> JavaScript template literals are not handled well by gettext, like in the case
> `` `<b>Breaks!</b>` `` [...] Note that error messages might be incorrect and the line number earlier in the file than is claimed.

There's a couple situations where doing even minorly-complex things with gettext really doesn't work well with js and can cause hard to find errors. Now that I look a bit deeper even the non-latin Serbian or (now out of date) Chinese translation wasn't translating these proper nouns:

```
#~ "要连接Android设备，请安装 KDE Connect 这个 Android应用 <a href=\"%s"
#~ "\">Google Play Store</a> or <a href=\"%s\">F-Droid</a>."
```

We should probably just hardcode it and let someone open a translation bug if it's an issue in some language.

rugk:
This is a quick "bugfix" here, but as I also pointed out in https://github.com/andyholmes/gnome-shell-extension-gsconnect/issues/268 a badge like "Get it on <addon store>" would be preferable.

Here you can get the [FDroid badge](https://gitlab.com/fdroid/artwork/blob/master/badge/README.md):
fdroid badge

ferdnyc:
Right, it's Sailfish that doesn't have such a button.

And as I said a few comments back:
> (With all due respect to @rugk, I think using app-link buttons would be overkill, and a waste of screen real estate because nobody will actually be using these — they'll search for "KDE Connect" in their device's applications directory, where they should be getting the app.)

rugk:
Ah okay skimmed over the comments here and did not saw this comments. So two problems:
* "Salfish does not have such a button" – well… okay, then you can place it below as a usual link. Also think this is not too bad, considering this OS is quite seldom used and with FDroid and GPlay we would have 4 buttons then, which aligns quite good into the layout instead of 5 as an odd number.
* The use case: "they'll search for "KDE Connect" in their device's applications directory" –
   I admit I have not thought about this and this makes quite much sense, however, actually at least for Google Play this is not the only use case. Actually, in Google Play, you can – if you are logged in – trigger an app installation even from the website on any device.
* "a waste of screen real estate" IMHO even the visual appearance is already important/enough reason to include them. Users e.g. are unlike to read these small texts and may not even notice there are apps for this extension. Also, it is quite some kind of equal "advertisement" for the different app stores – even across the OSes/systems, I think it's not so good to promote browser add-ons more than mobile apps.
  Finally, it looks much nicer… :wink:

andyholmes:
Well, in general I'm happy with the way it is in the screenshot, just with dropping the translatable hook. So if you're on board with that change @ferdnyc I'm happy to merge.

I think most Android users are going to be using Play Store, and promoting Sailfish is about as important as pushing F-Droid since it's just an alternative source of the same app. Generally, this is something most people will look at once, if it all.

> I think it's not so good to promote browser add-ons more than mobile apps

Well, one difference here is that the browser extensions are a part of GSConnect, but the Android/Sailfish apps are not.

ferdnyc:
Sorry, I caught a cold and I've been lying in bed moaning for the past day. I'm such a baby.

Why don't I flip the text to `translatable=no`, then, so the URLs don't end up in the translation file? Then we can merge it that way.

ferdnyc:
Oh, you said that. So, yeah, we're on the same page. I'll add that commit now.

andyholmes:
Yeah, I think i better turf it, it is not working well at all.

ferdnyc:
Seems fine to me... it's not speedy, but then again what's the rush? The question is whether it's doing anything useful to us with that time. If it is, I say keep it. If it's just grinding away for no benefit, then maybe it's not worth it.

Issue #246, opened 2018-10-19
Notifications plugin won't load: Argument 'object' (type interface) may not be null


EvilJimJafar:
Plugin won't load:

_init@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/notification.js:77:9
loadPlugin@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:783:26
loadPlugins@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:808:19
async*vfunc_startup/<@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:746:13
vfunc_startup@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:743:9
_init@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:97:9
@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:863:2

treba123:
Same here, fedora 29, gnome-shell 3.30.1.1

andyholmes:
Could you try with this ZIP:

[gsconnect@andyholmes.github.io.zip](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/2496021/gsconnect.andyholmes.github.io.zip)

See [Installing from Zip](https://github.com/andyholmes/gnome-shell-extension-gsconnect/wiki/Installation#installing-from-zip)

treba123:
Solves the problem for me

AGitBoy:
I'm using that ZIP on Ubuntu 18.10, and this same problem occurs for me as well. I'm on GNOME 3.30.1

andyholmes:
You're using the Zip posted in this thread? If so, what has the error message changed to?

EvilJimJafar:
> What version of GNOME Shell are you running?

3.30.1 here running on Manjaro.

AGitBoy:
No change in error message. I'll try rebooting and reninstalling the plugin.

AGitBoy:
After rebooting and reinstalling the extension, it no longer gives me the error message.

pauloxnet:
It worked for me on Ubuntu 18.04.1 LTS and Gnoem 3.28.2

[...]

PR #286, opened 2018-10-29
Sound Component: refactor libcanberra/gsound checks

### LINKED ISSUE #276: Unhelpful error when Sound Effects support is missing
I am using pop os.

@/home/mkamran/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/components/sound.js:17:15
_loadComponents@/home/mkamran/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:672:21
vfunc_startup@/home/mkamran/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:736:9
_init@/home/mkamran/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:97:9
@/home/mkamran/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:864:2

andyholmes:
Currently this is only used in one case; incoming "Find My Phone"
requests, but it's always checked at startup and only once. This caches
the found (or not) backend to be used, while checking if/and when the
action is actually invoked.

This still isn't ideal, because it means we might not find out we can't
use an audible alert until the request happens, but still better and it
should be rare that at least canberra-gtk-play is not availale in GNOME.

fixes #276

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,124 @@
+{
+    "env": {
+        "es6": true
+    },
+    "extends": "eslint:recommended",
+    "rules": {
+        "array-bracket-newline": [
+            "error",
+            "consistent"
+        ],
+        "array-bracket-spacing": [
+            "error",
+            "never"
+        ],
+        "brace-style": "error",
+        "comma-spacing": [
+            "error",
+            {
+                "before": false,
+                "after": true
+            }
+        ],
+        "indent": [
+            "error",
+            4,
+            {
+                "MemberExpression": "off"
+            }
+        ],
+        "key-spacing": [
+            "error",
+            {
+                "beforeColon": false,
+                "afterColon": true
+            }
+        ],
+        "keyword-spacing": [
+            "error",
+            {
+                "before": true,
+                "after": true
+            }
+        ],
+        "linebreak-style": [
+            "error",
+            "unix"
+        ],
+        "no-empty": [
+            "error",
+            {
+                "allowEmptyCatch": true
+            }
+        ],
+        "no-implicit-coercion": [
+            "error",
+            {
+                "allow": ["!!"]
+            }
+        ],
+        "nonblock-statement-body-position": [
+            "error",
+            "below"
+        ],
+        "object-curly-newline": [
+            "error",
+            {
+                "consistent": true
+            }
+        ],
+        "object-curly-spacing": "error",
+        "prefer-template": "error",
+        "quotes": [
+            "error",
+            "single",
+            {
+                "avoidEscape": true
+            }
+        ],
+        "semi": [
+            "error",
+            "always"
+        ],
+        "semi-spacing": [
+            "error",
+            {
+                "before": false,
+                "after": true
+            }
+        ],
+        "space-before-blocks": "error",
+        "space-infix-ops": [
+            "error",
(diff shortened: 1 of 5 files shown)


ferdnyc:
Hmm, my #285 changes (which are now merged) seem to have snuck into your branch. Not sure what effect that's going to have.

andyholmes:
I know, I'm just awful at git. I'm not sure how to untangle this. I think I'll just wipe the pr/branch and redo it.

andyholmes:
@ferdnyc 

Here's a good a place to discuss as any. I'm not opposed to a smaller version bump, but there are a number of fixes I'm not sure we can untangle, and some of them are reasonably important. There have also been one or two translation updates and I'm not sure how those work out if we start cherrypicking.

What's you feeling of `master` right now? We could just realign the translations and do a release. I have a pretty good feeling they didn't *really* review v13 since it was 13k lines of new code and passed in a day or so. They might be letting me off the hook for good past behaviour.

ferdnyc:
That's a good question. Other than my testing of the updated Settings > Other pane, I haven't really run from `master` at all. But in that limited testing, it _seems_ like it's in good shape... if a release were impending, though, I think I'd want to try running with it for a day or two, see how it shakes out. (And I can certainly start doing that now...)

If nothing else, I figure there are probably some outstanding translations that it'd be good to give the translators a little time with, no?

My thinking in #246 was to do a bugfix/maintenance release, branched off v13, with literally _just_ 41c96a5 applied, and no other changes. That change is pretty self-contained and independent of all the other work that's gone on, so it should be safe to cherry-pick. That could go to e.g.o as v14, probably get through pretty quickly (even if _anything_ would get through pretty quickly, sounds like), and then there'd be no rush to get v15 out, however ready or not ready it may be. 

The bigger the next "major" release is going to be, the more possibility new issues come up from the new changes. It'd be nice to have the upgrade fix released and widely adopted, before hitting people with all those other changes.

andyholmes:
That's fair enough, I'll just roll one out now and then maybe we can tag a release candidate for v15. I don't upload those to e.g.o but I think a handful of people give them a try.

andyholmes:
Alright, I wasn't quite sure how to tag a release pointing to a cherry picked commit, so I just made a phony branch, cherry picked and tagged off of that. I'll roll out a v15-rc1 in a bit here, and then I guess notify the translators and consider ourselves in a sort of feature/string freeze?

ferdnyc:
Very cool! Is v14 going in to e.g.o? I'll downgrade myself to v12 if so, in anticipation. (Or maybe I'll run with v15-rc1 a bit first, kick the tires, and then downgrade to v12.)

andyholmes:
Yep, but I think only I can see it until it's [reviewed](https://extensions.gnome.org/review/8581). Oops, looks like I forgot to bump version in the repo, so e.g.o did it :/

Right now I'm just trying to adjust the eslintrc from GJS's repo to the project, then I'll pop up a v15-rc and notify the translators. That codacy CI was brutally slow (probably their "free" server). Any recommendations for super lightweight linting CI's? It's actually kind of annoying Github doesn't have it's own pipelines like GitLab does.

PR #297, opened 2018-11-04
 Update French translation 

mkey-cdx:
Here is the update for the french translation (v15) + minor changes to be more consistent with gnome shell.


ferdnyc:
Mercí! (That one I didn't even have to look up in Google Translate. :wink: )

Issue #298, opened 2018-11-04
Translatable string with a mysterious variable


piotrdrag:
`%d Connected` in src/extension.js:246 has no translator comment, so I don’t know what %d stands for and I can’t translate it.

ferdnyc:
Good point; it is a bit cryptic. `%d` in printf format strings is a number (a base 10 integer, specifically — the "d" indicates signed "d"ecimal notation). Looking at the source, the context is the number of devices currently connected.

ferdnyc:
That's a good point, would this work as the generated output?
```
#. TRANSLATORS: The number of devices connected
#: src/extension.js:247
#, javascript-format
msgid "%d Connected"
msgid_plural "%d Connected"
msgstr[0] ""
msgstr[1] ""
```
(The result of modifying that section of code to
```js
if (this.available.length > 1) {
  //TRANSLATORS: The number of devices connected
  this._item.label.text = gsconnect.ngettext('%d Connected', '%d Connected',
  this.available.length).format(this.available.length);
```

(As you can see, the reality is even more cryptic. `%d` will always be >= 2 when that code path is followed, so the singular form will never actually be used. But that's not a detail translators really need to concern themselves with.)

I'll look around for other strings that involve countable values, and could benefit from `ngettext()`. But let me know if you notice any. There's currently only one other instance of plural translation in the source, the time duration in hours set at `src/shell/donotdisturb.js:263`. But I feel like there are probably other candidates.

piotrdrag:
`TRANSLATORS: %d is the number of devices connected` would be more clear. Apart from that nitpick, it looks good to me.

As for other strings that need ngettext, there is this one:

```
#. TRANSLATORS: Time duration in minutes (eg. 15 minutes)
#: src/service/ui/messaging.js:88 src/service/ui/messaging.js:120
#: src/shell/donotdisturb.js:266
#, javascript-format
msgid "%d minutes"
msgstr ""
```

PR #299, opened 2018-11-04
Use ngettext for countable strings

ferdnyc:
@piotrdrag pointed out that some countable strings weren't being translated using `ngettext`. This fixes that, and adds a translator comment on one of the strings. The `.pot` file is updated accordingly.

Fixes: #298

### DIFF
--- a/src/service/ui/messaging.js
+++ b/src/service/ui/messaging.js
@@ -85,7 +85,7 @@ function getTime(time) {
         // Under an hour
         case (diff < GLib.TIME_SPAN_HOUR):
             // TRANSLATORS: Time duration in minutes (eg. 15 minutes)
-            return gsconnect.ngettext('%d minute', '%d minutes', (diff / GLib.TIME_SPAN_MINUTE)).format(diff / GLib.TIME_SPAN_MINUTE);
+            return ngettext('%d minute', '%d minutes', (diff / GLib.TIME_SPAN_MINUTE)).format(diff / GLib.TIME_SPAN_MINUTE);
 
         // Yesterday, but less than 24 hours ago
         case (diff < GLib.TIME_SPAN_DAY && (now.get_day_of_month() !== time.get_day_of_month())):
@@ -117,7 +117,7 @@ function getShortTime(time) {
 
         case (diff < GLib.TIME_SPAN_HOUR):
             // TRANSLATORS: Time duration in minutes (eg. 15 minutes)
-            return gsconnect.ngettext('%d minute', '%d minutes', (diff / GLib.TIME_SPAN_MINUTE)).format(diff / GLib.TIME_SPAN_MINUTE);
+            return ngettext('%d minute', '%d minutes', (diff / GLib.TIME_SPAN_MINUTE)).format(diff / GLib.TIME_SPAN_MINUTE);
 
         case (diff < (GLib.TIME_SPAN_DAY * 7)):
             return time.format('%a');
--- a/src/service/ui/messaging.js
+++ b/src/service/ui/messaging.js
@@ -85,7 +85,7 @@ function getTime(time) {
         // Under an hour
         case (diff < GLib.TIME_SPAN_HOUR):
             // TRANSLATORS: Time duration in minutes (eg. 15 minutes)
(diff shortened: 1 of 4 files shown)


src/service/ui/messaging.js
```diff
@@ -117,7 +117,7 @@ function getShortTime(time) {
 
         case (diff < GLib.TIME_SPAN_HOUR):
             // TRANSLATORS: Time duration in minutes (eg. 15 minutes)
-            return _('%d minutes').format(diff / GLib.TIME_SPAN_MINUTE);
+            return gsconnect.ngettext('%d minute', '%d minutes', (diff / GLib.TIME_SPAN_MINUTE)).format(diff / GLib.TIME_SPAN_MINUTE);
```
andyholmes: This is a weird
```suggestion
            return ngettext('%d minute', '%d minutes', (diff / GLib.TIME_SPAN_MINUTE)).format(diff / GLib.TIME_SPAN_MINUTE);
```
This is a non-obvious situation, but when under the "GSConnect process" the gettext functions are directly on the `window` (global) variable. When under the gnome-shell process they're on the `window.gsconnect` variable because otherwise we would clobber gnome-shell's (since they already have _their's_ on the global object).

src/service/ui/messaging.js
```diff
@@ -85,7 +85,7 @@ function getTime(time) {
         // Under an hour
         case (diff < GLib.TIME_SPAN_HOUR):
             // TRANSLATORS: Time duration in minutes (eg. 15 minutes)
-            return _('%d minutes').format(diff / GLib.TIME_SPAN_MINUTE);
+            return gsconnect.ngettext('%d minute', '%d minutes', (diff / GLib.TIME_SPAN_MINUTE)).format(diff / GLib.TIME_SPAN_MINUTE);
```
andyholmes:
```suggestion
            return ngettext('%d minute', '%d minutes', (diff / GLib.TIME_SPAN_MINUTE)).format(diff / GLib.TIME_SPAN_MINUTE);
```
Same here as below

src/shell/donotdisturb.js
```diff
@@ -263,7 +263,7 @@ var Dialog = class Dialog extends ModalDialog.ModalDialog {
             return gsconnect.ngettext('One hour', '%d hours', hours).format(hours);
         } else {
             // TRANSLATORS: Time duration in minutes (eg. 15 minutes)
-            return _('%d minutes').format(this._time / 60);
+            return gsconnect.ngettext('%d minute', '%d minutes', (this.time / 60)).format(this._time / 60);
```
andyholmes: This one *is* correct, because all the shell extension stuff is under gnome-shell's proc.

andyholmes:
One thing to note here, is once we merge this I think everyone will have to resubmit translations. I doubt gettext can "fuzzy" handle singular->plural changes.

No biggy, I'm totally fine putting off release until everyone has that chance, because it's a bummer when people update translations *right* after release :P

src/service/ui/messaging.js
ferdnyc: Aha! Well, that's the reason I submitted this as a pull request instead of directly committing it. 😃

(Actually, the reason is mostly that I didn't actually _run_ these changes, I only tested the `.pot` file generation. I'm still squatting on v12 waiting for v14 to get reviewed, and feeling more stupid every day that doesn't happen.)

andyholmes: Yeah, that's why I decided might as well get v15 on the go, too :P

ferdnyc:
That was the _other_ reason I submitted this as a PR, I wanted to leave the option open to hold off until after v15 is released. Which I still think would be fine, actually. It's only two strings. Then again, because the "Connected" one is so new, almost none of the translations have it even in its old form.

piotrdrag:
I would rather wait for this PR to be merged before I update the translation.

andyholmes:
@ferdnyc do you want to do the review changes, or have you given me edit permissions for the PR, in which case I can just do them in the web interface?

ferdnyc:
Sorry, I was replying from a subway car earlier, didn't see that last bit. :grin: But, sounds like it got sorted without me.

PR #302, opened 2018-11-05
Update Brazilian Portuguese translation

ricvelozo:
Some strings turned comments after update from POT... Can you check if is ok?


ferdnyc:
Everything looks good with this, thanks! 

The commented strings are ones that we removed from the source or edited their (English) meaning since the last time the master `.pot` file was generated. They're stored in comments that way to make it clear that was intentional. (And also so any translations provided for those strings will be preserved, in case they're put back into the code at a later date.)

PR #309, opened 2018-11-05
Update german translation

taaem:


andyholmes:
Thanks, been awhile since we had a german update :)

PR #317, opened 2018-11-08
Fix type error in about dialog creation

scde:
Hi,

I noticed that the about dialog was not working and threw this error:
```
JS ERROR: Error: Wrong type integer; string expected
_init/Gtk.Widget.prototype._init@resource:///org/gnome/gjs/modules/overrides/Gtk.js:48:9
_aboutAction@/home/user/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:399:27
@/home/user/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:881:2
```
It turns out it was a type error with the `version` field of the AboutDialog object. It should be a string (see [here](https://devdocs.baznga.org/gtk20~2.24.31/gtk.aboutdialog#property-version)) but `gsconnect.metadata.version` is an integer.

Thanks for all the work and have a nice day =)

### DIFF
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -413,7 +413,7 @@ const Service = GObject.registerClass({
                 program_name: _('GSConnect'),
                 // TRANSLATORS: eg. 'Translator Name <your.email@domain.com>'
                 translator_credits: _('translator-credits'),
-                version: gsconnect.metadata.version,
+                version: gsconnect.metadata.version.toString(),
                 website: gsconnect.metadata.url,
                 license_type: Gtk.License.GPL_2_0
             });


ferdnyc:
Confirmed, the bug prevents the About window from displaying at all. This fix corrects that, LGTM.

(The weird thing is, it seems like that code's been like that for a while now — I _thought_ the About window worked in v13, but in light of this bug I don't see how it could have.)

scde:
@ferdnyc I just today updated from v13 to v15 (from the GNOME shell extension website) and I can confirm that the dialog was not working in v13 either.

andyholmes:
That's strange, because it was definitely that way for awhile, *but* I actually made the exact same mistake before the rewrite; there was even a special release to fix it ([v10](https://github.com/andyholmes/gnome-shell-extension-gsconnect/releases/tag/v10)) :stuck_out_tongue:

andyholmes:
It should bug everytime actually! It totally worked for me, because I double check to see I've bumped the version. It's also weird because JavaScript usually coerces everything, even when you don't want it to.

Issue #320, opened 2018-11-09
No SMS history for Signal, Samsung Messages


stratus-ss:
I have recently installed GS Connect on a fully up to date Arch Linux system.
The Phone is an up to date Honor 6x (or as up to date as the manufacturer allows)

I can send SMS from the desktop but the history is lost. On the phone I only have one side of the conversation (the received SMS). In the application, I can see my contacts but if I send an sms from the desktop and then go to the main contact list in the messaging window the history is lost.

On the phone I use signal as my SMS app. Is this an expected limitation?

andyholmes:
I have heard of similar problems, even with KDE Connect, although I don't know if it's because of the SMS app or not. Only the most recent version of the Android app (v1.10) has support for SMS history, but that is read from Android's [SMS database and API](https://developer.android.com/reference/android/provider/Telephony.Sms.Conversations).

Whether Signal uses that database, or it's own I don't know. I know there is a newer and older API for this and Signal may not support the newer API, possibly so it can support older phones. If you wanted to test, you can try using the stock app and send a few test messages.

This would be an upstream bug in the Android app (especially if the message is being "sent" properly), so the best thing to do is to open a bug in [KDE Connect's bugzilla](https://bugs.kde.org/buglist.cgi?quicksearch=kdeconnect&list_id=1505828), including a link to _this_ bug, and then add a link here to _that one_ so both projects can track it. You might find that it's a problem with Signal and you could see if they have a bug tracker too, and file a bug there :)

It might be a bit of work and waiting, but that's the best way to get things moving.

andyholmes:
Signal: [#5729 Signal doesn't store incoming SMS messages to Android's message store](https://github.com/signalapp/Signal-Android/issues/5729) [WONTFIX]

pwd-github:
It would seem the only solution would be to apply to Signal for accept to their encrypted message storage.

andyholmes:
Firstly, there are four relevant "packets" involved here:

* **Notifications** (*notification* plugin)
* **SMS send** & **SMS event** (*telephony* plugin)
* **SMS messages** (*sms* plugin)

*Long* before upstream KDE Connect had two-way SMS conversations or history of any kind, GSConnect used to fake it. The way this was done was by (1) collecting **SMS send** (outgoing) and (2) using flaky tests to match **SMS event** and **Notification** packets (incoming) to each other, because they were essentially duplicates.

[...]

PR #332, opened 2018-11-14
Messaging Interface Rewrite

### LINKED ISSUE #288: [DEVEL] Messaging interface
The Gnome Shell (GMenu, mostly) interface has gotten a lot of love in #205, and has come a long way for it, so I wanted to open a conversation (post-v15, I'm thinking) about the SMS plugin's contacts window, which is feeling sort of "legacy" at this point, especially compared to the rest of the (updated & refreshed) UI.

First, one observation and one question:

* **Observation: The packed-too-tight header bar makes it _really_ hard to move the window** 
    There's no place to grab for dragging. I'm thinking the text entry widget should have spacers on both sides, to provide a little titlebar-handle real estate for the window.

* **Question: Why are contacts selected using _checkboxes_?**
     I'm _guessing_ the idea was, originally, that multiple contacts would be selectable for group texting... but that doesn't actually exist as a feature, does it? (I know it doesn't work on my phone, but I'm on a CDMA network and I know legacy SMS can be weird there, compared to GSM.) 
    
    If there's no ability to multi-select contacts, then those checkboxes just don't make any sense, and it'd be way more user friendly to just make the entire row clickable from the start of the number to the end of the label.

Actually, I lied, a second observation:
* **Observation: The right-justified elements are _waaay_ too far away from the associated numbers.**
    They're very hard to tie together visually, given the width of the window. (Which is excessive, for the purposes of a contacts list, though it's a good size for the conversation interface.) I imagine if my contacts had more phone numbers, it'd start to get hard to track which label/checkbox corresponds to which number.

image

### LINKED ISSUE #339: First message gray (like unsend)
When I write a message, the first one is marked as gray, like hasn't sent (but it is sent) and next, messages are normal. Why is that? There is a screenshot ("test" message is the first one).
I've installed the newest version of extension from **extensions.gnome.org**.

test

Sorry for my english. Thank you for help.

### LINKED ISSUE #341: Contacts updates on the device can cause mistargeted SMS sessions
**Describe the bug**
This is kind of a weird one. It appears that when contacts are updated on the device, it can trick GSConnect into using the wrong number for a contact.

**To Reproduce**
Steps that triggered the behavior:
1. I was  having an SMS conversation with an unknown number
2. During the conversation, I added their number to an _existing_ contact
3. After that point, GSConnect's notifications of texts mirrored from that same number started being labeled with the contact's name
4. However, the new _number_ (the one the SMS messages are actually coming from) hasn't been added in the GSConnects contacts list
5. As a result of that (presumably), when I clicked on the latest SMS notification from GSConnect, it opened an SMS window to the **wrong** number (the contact's old number, which isn't the number the text message came from)

**Expected behavior**
SMS conversation windows always open with the recipient set to the number which sent the message in the notification

 - GSConnect version: v15
   - Installed from: extensions.gnome.org
 - GNOME/Shell version: 3.30
 - Distro/Release: Fedora 29

 - Paired Device(s): Galaxy S6
 - KDE Connect app version: latest beta from Google play
 - Messaging App: Android Messages

andyholmes:
This is the rewritten messaging UI. It's about 99% functional, although there are still places where it could be improved. One big pro is there's only one window per device, which saves (relatively) a lot of memory. This PR closes #288 

- [x] Move the "floating" entry into the contact list widget and use Gtk template
- [x] Inline the *555-5555・NumType* label instead of concatenating
- [x] The placeholder widget label, when no contacts are available, should also be updated
- [x] Maybe add a panel when we have a pending message (eg. Share SMS from browser plugin) and a contact is yet to be selected or, alternatively, we could re-use the above panel.
 - [x] For sure a rarity, but we don't support multi-recipient conversations at all anymore so if a `sms://` URI with multiple recipients is handled, we only take the first phone number
- [x] Remove the legacy `telephony` and `sms` packets and behaviour

  Since a number of SMS apps don't support SMS history, we now have no way to tell if we should manually inject messages using the `kdeconnect.telephony` packet, so those devices can't get feedback about sent or received SMS in the window (See Issue #320).

  I spoke with @sredman about this, and we agreed it's better to get SMS working for well-behaved apps first and worry about the rest later. The legacy packets haven't been removed from `kdeconnect-android`, and there are other ways we might fake partial support, when the time comes.
- [x] Add some method of feedback for messages waiting to be sent by Android, fixes #339
- [x] Fix weird hidden window bug when migrating from "Share by SMS" dialog
- [x] Don't open conversations directly based on a contact name, fixes #341 

**Shortcomings**

* Pop-out conversations are not supported for now, since it will complicate receiving messages
* Additionally closing SMS notifications is spotty at best.

**Testing**

Pull from the `messaging-interface` branch or install the latest posted Zip (translations are not included).


andyholmes:
Updated Zip for 39f07c3: [gsconnect@andyholmes.github.io.zip](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/2592653/gsconnect.andyholmes.github.io.zip)

PR #338, opened 2018-11-16
Add issue templates

ferdnyc:
I thought it might be helpful, when users open new Issues, to prompt for a little more structure and give them an idea of what details would be most helpful.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,43 @@
+---
+name: Bug report
+about: Report an issue to help us improve GSConnect
+labels: 
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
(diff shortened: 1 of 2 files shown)


ferdnyc:
(The templates mean that clicking "New Issue" will first show this screen, then open the edit window pre-populated with the selected template for editing...)

image

ferdnyc:
I almost wonder if, along the same lines, the bug report template should try to delineate what kind of reports are really KDE Connect app issues that we won't have any ability to address. (That's a bit harder to define cleanly without any context, though...)

andyholmes:
That's probably okay, it's good to hear about them anyways because we can ask them file upstream and link the two bugs.

ferdnyc:
Note added to bottom of template:

>#### Please delete this line and everything below before submitting your request
>
>GSConnect works in concert with the [KDE Connect][1] app on paired Android devices. Communication between GSConnect and the device is controlled by **KDE Connect**, _not_ GSConnect.
>
>Any request that would require features not supported in KDE Connect, or would require GSConnect to deviate from the KDE Connect protocols, cannot be implemented. Any such requests should be made to the KDE Connect project. The requested functionality can only be implemented in GSConnect after support is added to KDE Connect.
>
> [1]: (https://community.kde.org/KDEConnect)

ferdnyc:
If there are any "greatest hits" requests of that type that keep coming up again and again (or if any emerge), we can probably even add them to the bottom of the template in a sort of preemptive "Don't bother" list.

PR #347, opened 2018-11-20
Updated italian translation

amivaleo:
:)


amivaleo:
"conflicting files" 🤔

andyholmes:
does that look right now?

po/it.po
```diff
@@ -220,6 +220,7 @@ msgstr "Messaggistica"
 
 #: data/messaging.ui:110
 msgid "Select a conversation"
+msgstr "Aggiungi persone per iniziare una conversazione"
```
amivaleo: The italian for that is "Seleziona una conversazione".

po/it.po
```diff
@@ -314,8 +315,8 @@ msgstr "GSConnect"
 #, javascript-format
 msgid "%d Connected"
 msgid_plural "%d Connected"
-msgstr[0] "%d% Connesso"
-msgstr[1] "%d% Connessi"
+msgstr[0] "%d Connesso"
+msgstr[1] "%d Connessi"
```
amivaleo: This is perfect. ;)

po/it.po
andyholmes: Got it. Should have seen that because I actually chose Italian in high school ;)

amivaleo: Oh... That means you don't really need my help in translating. :)

It must be alright now.

By the way, I liked the battery info. The latest update removed that from everywhere. :(

andyholmes: Yeah, me too, but there were some complaints about it, especially for accessible users :/

I hope to come back to that part of the UI in future. I think it would be neat if people could make simple UI files to change it and that they could share. There's a thing called [ClutterScript](https://developer.gnome.org/clutter/stable/ClutterScript.html) I want to look into that might work for that.

amivaleo: What people had to complain about that? o.o

I can't help with coding in any way.
I tried to understand something about Glade, for example, but it doesn't do what people get in their yt videos... -.-

I really would like to help with (G)UIs, but they're a puzzle to me.

andyholmes: Ah, UIs are actually a lot more complicated and a lot less "fun" than regular coding :) It was something I found surprising when started writing more code.

Maybe I'll take a day off soon (from my spare time doing this) to write a How-to on GJS/Glade/Gtk Templates. Those are always a bit of fun.

amivaleo: "UIs are actually a lot more complicated and a lot less "fun" than regular coding"
And that's bad. :(
It could be just a matter of drag&drop ( & tweaking).

Please, do write some guide or record a video.
I really really really want to learn, but I got stuck already with the basics, since it looks like that "my" glade" works differently from everybody else's. :|

amivaleo: Can't wait to get it.
Thank you. :)

PR #350, opened 2018-11-21
tr.po update

orhanenginokay:
[tr.zip](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/2603431/tr.zip)

I update tr.po ^^


andyholmes:
I think you pulled the wrong branch - you used `legacy` instead of `master` :P 

It's okay, I got your translation from the Zip file and committed for you :)

PR #353, opened 2018-11-22
Add fallback launching for debug logs

### LINKED ISSUE #329: Remote Input not working
The remote input plugin does not work on my desktop PC. I can neither move the cursor nor produce mouse clicks.

The other plugins work fine, so it's definitely not the connection. In particular I can see the battery levels, switch between songs and view notifications no problem. I have the same phone connected to two laptops with similar configurations and remote input works there.

If there are any log files I can provide let me know.

---

Desktop & laptops:

- OS: Archlinux
- GNOME: 3.30.1 (running on X)
- GSConnect: v15

Phone:

- KDE Connect 1.10 (installed from F-Droid)
- Android 7.1.2

ferdnyc:
If launching `gnome-terminal` directly fails, we fall back to launching the two `journalctl` commands via Gio.AppInfo's "NEEDS_TERMINAL" application launcher. Should automatically bring up whatever recognized terminal emulator is installed.

In my testing, after installing `rxvt` and uninstalling `gnome-terminal` completely, hitting the terminal-launcher in the extension's Debug window got me two terminal windows, each showing one of the two GSConnect debug log views.

(To be honest, I wasn't sure what `debug.js` was for, but I noticed that terminal launching was also present there, so I added the same fallback logic to that code as in `daemon.js`. It'll never be used unless `gnome-terminal` isn't installed, so I figure it was safe enough to add anyway.)

Fixes #329

### DIFF
--- a/data/debug.ui
+++ b/data/debug.ui
@@ -179,7 +179,7 @@
                     <property name="receives_default">True</property>
                     <property name="tooltip_text" translatable="yes">Debug Terminal</property>
                     <property name="valign">center</property>
-                    <property name="action_name">app.log</property>
+                    <property name="action_name">win.log</property>
                     <child>
                       <object class="GtkImage">
                         <property name="visible">True</property>
@@ -876,3 +876,4 @@
     </child>
   </template>
 </interface>
+
--- a/src/service/components/debug.js
+++ b/src/service/components/debug.js
@@ -48,6 +48,10 @@ var Window = GObject.registerClass({
             visible: true
         });
 
+        // Log & Debug Mode actions
+        let logAction = new Gio.SimpleAction({name: 'log'});
+        logAction.connect('activate', this._logAction);
+        this.add_action(logAction);
         this.add_action(gsconnect.settings.create_action('debug'));
 
         // Watch for device changes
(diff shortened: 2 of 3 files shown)


andyholmes:
Good idea. Made a few changes there, figured might as well rip this stuff out of `daemon.js`, since it doesn't really belong there anyways.

If my changes work for you, you can go ahead and merge this :+1:

andyholmes:
Also, isn't a bit weird I could just push to your fork without asking, or did you allow that explicitly? :stuck_out_tongue_closed_eyes:

ferdnyc:
> Also, isn't a bit weird I could just push to your fork without asking, or did you allow that explicitly?

I didn't in the sense that it's the default, but there is a checkbox displayed in the right column of the PR page indicating that the branch is set to "Allow edits from maintainers."**\*** So I have the option to disable that access, though I can't imagine why I ever would. It only allows you to push to the specific topic branch for the PR, and that branch is code that I'm proposing go into _your_ repo after all! :laughing: 

**\*** - (There's also a "Learn more" popup which explains):
> If checked, users with write access to **andyholmes/gnome-shell-extension-gsconnect** can add new commits to your `terminal-fallback branch`. You can always change this setting later.

Changes LGTM, though I see I was too slow on the merge anyway. :+1:

andyholmes:
> Changes LGTM, though I see I was too slow on the merge anyway. ￼￼￼:+1:

Ah now worries, I was mostly being lazy. I spent all day on the WebExtension, which was a total mess. Another thing I "ripped off" someone else and never put much time into, but I had setup that auto-translation script awhile ago and it was a little bizarre to only have English and Serbian as languages :stuck_out_tongue_closed_eyes:

PR #363, opened 2018-11-28
Implement Gio backed SFTP connections

andyholmes:
This is the first step to properly integrating into Nautilus. Once it's confirmed there are no regressions with `sshfs` and Gio backed connections are stable enough, we can merge this and remove `sshfs` as an official dependency of the SFTP plugin.

* SFTP Plugin: implement Gio connections as a fallback when sshfs is missing
* SFTP Plugin: allow private key authentication for Gio-backed sftp connections
* SFTP Plugin: [HACK] clean up stale mounts in _sftp_mount() after success
* SFTP Plugin: [HACK] catch a special case where mounting fails because a
  GMount is not unmounted on service shutdown, but uses the same port so can be
  reused by applying the new password
* Settings UI: remove checks and widgets for sshfs
* Device: amend openPath GAction to accept any URI
* Service: increase RSA key length to 4096
* meson: add build-time option for ssh-add


andyholmes:
Zip as of 04b8500: [gsconnect@andyholmes.github.io.zip](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/2623168/gsconnect.andyholmes.github.io.zip)

ferdnyc:
I hadn't been following this as it was ongoing (the deluge, again), but it sounds pretty awesome. Does this mean that phone connections will show up in the Nautilus sidebar / as desktop icons (if enabled), as well? Or, does it pave the way to potentially making that happen? I'd be interested in going down that road if we're not already there. (I should really just check out the latest changes and see for myself, but this is a week of other projects so that'll probably have to wait a few days at least.)

andyholmes:
> Does this mean that phone connections will show up in the Nautilus sidebar / as desktop icons (if enabled), as well? 

If you're running from `master` you can uninstall `sshfs` to have it fallback to Gio. Later once it's better tested I think we'll drop it entirely, unless NixOS wants it.

Unfortunately, Android exports the root directory and I don't think there's a way around that from our end. So it will show in Nautilus, but you won't have read or write access and have to change the path manually (or use the regular device menu method). I plan on trying to port primitive-ftpd's SAF implementation to kdeconnect-android, but it'll obviously have to wait until I have some free-free time :P

PR #384, opened 2018-12-08
French translations (issue with hour ?)

gautierlabarre:
My first contribution, 

My POEditor told me there is an issue with One hour / %d hours.
I checked for minutes and it's displayed differently : 
```
msgid "One hour"
msgid_plural "%d hours"
msgstr[0] "Une heure"
msgstr[1] "%d heures"
```
```
msgid "%d minute"
msgid_plural "%d minutes"
msgstr[0] "%d minute"
msgstr[1] "%d minutes"
```

To me, msgId for hour should be "%d hour" to be more consistent with minutes.

Hope this helps.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: org.gnome.Shell.Extensions.GSConnect\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2018-11-19 11:14-0800\n"
-"PO-Revision-Date: 2018-12-08 11:23+0100\n"
+"PO-Revision-Date: 2018-12-08 11:32+0100\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: fr\n"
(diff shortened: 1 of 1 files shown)


andyholmes:
> My first contribution

Thanks!

> My POEditor told me there is an issue with One hour / %d hours.

I think it's usually preferable to use a plain language word for cases like this, but it also shouldn't cause a warning in translation editors. Unfortunately if the `msgid` is `%d hour` I don't think it will let anyone use a string/word since it will always expect a digit.

> I checked for minutes and it's displayed differently

That wasn't me :stuck_out_tongue:. Actually I think `1 minute` will never happen, because it will be superseded by `Just now` in the code. 

There's some notes about this plural situation here, which seems to say there shouldn't be a problem:
https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

@piotrdrag do you have any insights about this?

piotrdrag:
Good catch! You shouldn’t use `One hour` for singular, because many languages use the first form also for e.g. 21, 31, 41 etc.

PR #386, opened 2018-12-09
Improve readme.md

gautierlabarre:
For a better grasp of what GSConnect is and can do, i updated this file to : 

- Show a picture (that needs updating, I know)
- List all features 
- Add a direct link to installation


ferdnyc:
Thanks! The update was sorely needed, and I think this looks great. I only have two notes:

1. It looks like some changes to `po/fr.mo` snuck in with one of your commits, and since it's a binary file I can't easily see what's in there. So I'm a little wary of merging this with that change included.

2. I wonder if it's possible for the the "To install..." part to give more of a summary of what's required/involved directly, then point to the Wiki install instructions as a further resource for those who need it. 
   I'm not even sure myself exactly what I'm thinking, something like a link to the [extensions.gnome.org][ego] page directly from the README, and a note about requiring the KDE Connect Android app, and **_then_** "For step-by-step instructions and links to troubleshooting resources, see the **[Installation Instructions][install]** in the Wiki." or something like that. 

...Well, it can always be further edited afterwards, so point 2 isn't really anything worth holding up this PR for. So really the only issue is that translation file.

[ego]: https://extensions.gnome.org/extension/1319/gsconnect/
[install]: https://github.com/andyholmes/gnome-shell-extension-gsconnect/wiki/Installation#standard

ferdnyc:
Oh! I just realized that `po/fr.mo` isn't actually in the repo currently, so that's not a problem. I can just merge this, then delete the file. No big deal. Comitting now, thanks!

gautierlabarre:
Oh yeah, sorry i forgot this file from a previous pull request. Thanks for the rectification, i'll be more careful next time.

For your second point, i'll add more stuff on this file later, i didn't want to add too much, because it would duplicate some text so... But you're right, some bullet point on what is mandatory and the link for a more step-by-step could be nice.

Thanks.

andyholmes:
Thanks for the update! I'd really rather keep the readme as small as possible though and use the Wiki as the source of all information.

The readme used to be an exact copy of the Wiki home page, but it required constant tiny pushes to keep it up to date, whereas the Wiki can be updated without affecting the main repo.

ferdnyc:
Aha, gotcha. Well that's my fault, I should've checked with @andyholmes before merging. Whoops.

ferdnyc:
Reverted, all back to normal. Sorry 'bout that.

andyholmes:
Oh no worries, I think we _can_ improve it a bit with some quick links (and this git after all, we can always change our mind). But anything that's likely to change, like a list of features, should probably be in the Wiki especially since _everywhere_ links there now.

A list of features is actually a good idea, I often lose track of the Wiki. I kind of stuffed a bunch of junk in the Device page that should really be elsewhere. Maybe a whole page for Features is worth doing, especially since there are Service features (libfolks/Gnome Online Accounts, UPower, etc) and Device features; some of which could probably benefit from explanations. I wish nothing needed explanation, but I think that's not the case.

On related note, I'm doing *another* settings UI spree (including lots of ideas from @amivaleo), including removing the debugger from user-view. Instead I'm going to do a "Generate Support Log" somewhat like GNOME Builder, that turns on debugging until the dialog is closed and lets the user review the log. After the 3rd person uploaded a JS engine heap dump, it became obvious this was confusing for users (and I totally don't maintain that UI between releases or test for data loss/corruption).

\</hijack-thread\>

PR #398, opened 2018-12-20
Add Belarusian translation

DeadMetaler:


po/be.po
```diff
@@ -0,0 +1,937 @@
...
+"POT-Creation-Date: 2018-12-19 02:37+0300\n"
+"PO-Revision-Date: 2018-12-19 13:33+0300\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=3; plural=(n%10==1 && n%100!=11 ? 0 : n%10>=2 && n%10<=4 && (n%100<12 || n%100>14) ? 1 : 2);\n"
+"X-Generator: Poedit 2.2\n"
+"Last-Translator: Максім Крапіўка <metalomaniax@gmail.com>\n"
+"Language: be_BY\n"
```
ferdnyc: I just have one question before merging this. It's only because I'm clueless about `poedit` and about translations in general, so I'm being over-cautious.

I noticed that the language in `LINGUAS` and the filename is `be`, but the metadata generated by `poedit` shows `"Language: be_BY\n"` here. 

I checked, and **all** of the other translations list the same code in `Language:` that's used for the filename, so I'm wondering if it should be `"Language: be\n"` just to be safe? Or does it not matter?

ferdnyc:
Thanks, BTW! Did you get that Nautilus-extension translation issue you were asking about in #1 sorted out?

andyholmes:
> Also it doesn't affect to my issue with not translatable Nautilus-extension. Yes, issue still present for me.

You'll probably have to restart Nautilus with `nautilus -q`. If that doesn't fix it, then it may be the locale domain should be `be`.

po/be.po
```diff
@@ -0,0 +1,937 @@
...
+#: src/service/ui/settings.js:616
+msgid "translator-credits"
+msgstr "Maksim Krapiŭka <metalomaniax@gmail.com>"
+
+#: src/service/ui/settings.js:646
+msgid "Debug Logging Enabled"
+msgstr "Журнал адладкі ўключаны"
+
+#: src/service/ui/settings.js:647
+msgid "Reproduce the problem and press “Done” to review the log."
```
andyholmes: I think this is still from the previous POT file. Maybe after you've fetched and rebased you should update with ninja?

    ninja -C _build org.gnome.Shell.Extensions.GSConnect-update-po

andyholmes:
I'll probably release v18 tomorrow sometime, so if you don't get a chance to fix those last bits I'll just merge what's there and you can fix them for next release.

andyholmes:
Going to merge this as-is. I'll mark the changed strings as fuzzy, since they're close enough, and it'll be easier to update it later I think.

DeadMetaler:
Oh.. Thanks for merging and realign my localization file because I was not able to do it myself. 
I found some commits with mentions Crowdin. Can I translate using Crowdin now? It's much simpler for me for using this service.

andyholmes:
> Can I translate using Crowdin now? It's much simpler for me for using this service.

It will definitely make it easier, I just have to sort out some issues that Crowdin in causing. I've been emailing with tech support today to try and get them fixed, so hopefully by tomorrow or so.

PR #422, opened 2019-01-01
Extension: replace GDBusObjectManager with in-house rewrite

andyholmes:
Among general performance improvements during startup, this also makes
it possible to truly use the Wayland backend for Gdk, which allows for
Multi-DPI support (at the expense of clipboard).

Additionally, the battery state is now exposed as a stateful GAction,
which is easy to work with and compresses changed signals into a single
emission per update.

* export battery state as a GAction
* use raw DBus call for service settings
* new streamlined manager class

TODO:

* [ ] Fix corner-case where device menu doesn't short on service restart
* [ ] Disable forcing Gdk X11 backend
* [ ] Disable Clipboard on Wayland


ferdnyc:
@andyholmes TODO list:
>  Fix corner-case where device menu doesn't short on service restart

...Que?

andyholmes:
Por alguna razón, si se reinicia el servicio, los menús del nuevo dispositivo no tienen ningún elemento.

ferdnyc:
Hmm, kinda screwed myself with that joke, since I don't speak a lick of Spanish beyond the stuff you can't _help_ but pick up (like "Que?"). :wink: (GTranslate to the rescue!)

ofuangka:
Do you plan on re-implementing Clipboard for Wayland somehow?  Or is it gone indefinitely?

ferdnyc:
@ofuangka
At the risk of speaking for @andyholmes, I'm sure that he'd very much _like_ clipboard functionality to work in Wayland as well, it's just a matter of finding a way to do that.

**Note for reference**, on that front: Though I haven't tried it myself (still on Xorg), the bugaevc/wl-clipboard project sounds like it may be very enlightening as a reference implementation.

andyholmes:
> Do you plan on re-implementing Clipboard for Wayland somehow? Or is it gone indefinitely?

The problem is Wayland really has no support for a global clipboard. Among other reasons, because allowing applications to access each other's data in that way is exactly what Wayland aims to prevent. The same is true of mouse and keyboard synthesis (remote input).

Previously we worked around this by forcing the [XWayland](https://wiki.gnome.org/Initiatives/Wayland/XWayland) backend, essentially making GSConnect a X11-only application, but this caused a number of other problems requiring other workarounds to prevent stalling or crashing Wayland.

There is a [conversation in progress](https://lists.freedesktop.org/archives/wayland-devel/2018-November/039623.html) about a Wayland extension for clipboard management (started by Albert Vaka of KDE Connect), but it has yet to be implemented by anyone (except maybe [SwayWM](https://github.com/swaywm/wlr-protocols/pull/25)). Probably implementing this for GNOME will require interacting with PermissionStore or something so that users can grant permission before allowing applications to access clipboard data.

The only possible workaround without breaking Wayland again would be to run another mini XWayland server that communicates clipboard changes over pipes or something. If you want to open a bug titled "XWayland Clipboard Proxy" and link to this pull request, I'll try and come back to it when I get some time and see if there's a non-ugly way to do that. I haven't had much time for GSConnect lately, but the next release when Clipboard is being removed is a long way off, so maybe I'll have time before then.

ofuangka:
@ferdnyc @andyholmes
Thanks for the context.  I could tell the removal was a deliberate decision but was curious about the details.

PR #425, opened 2019-01-03
Add Experimental switch to force GVFS use

ferdnyc:
Thought I'd formalize my previous hack, since the Experimental section was added to device settings. Reversed the user-facing logic to "use (force) GVFS" instead of "disable sshfs", even though conceptually it's really still doing the latter.

### DIFF
--- a/data/device.ui
+++ b/data/device.ui
@@ -1555,6 +1555,59 @@
                                 </accessibility>
                               </object>
                             </child>
+                            <child>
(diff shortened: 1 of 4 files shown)


andyholmes:
Not sure about this one tbh. I don't really like the idea of experimental features to begin with, but also the only real benefit of GVFS is that it works when `sshfs` is not installed. I'll have a think about it.

ferdnyc:
Fair enough!

Issue #340, opened 2018-11-16
Presentation remote support


CBiX:
In a recent update of KDE Connect for Android a feature for remote controlling presentations has been added. You can even lock the phone and control a presentation with volume buttons!

However, it doesn't work with GSconnect and applications like LibreOffice or Evince. Only the menu item "exit presentation" works. It would be awesome to see a full implementation of this feature in GSconnect and I'd be glad to test it.

andyholmes:
Hi, thanks for the report.

The presentation remote is actually a convenient UI that re-uses the `mousepad` plugin to send keystrokes. If you are using Wayland it won't be possible to use this remote due to limitations synthesizing mouse/keyboard input.

If you aren't using Wayland, let me know and I might need you test the input. It should just show up in a terminal as `^[[5~` and `^[[6~`, but you could also run this command for more detailed info:

    $ xinput test-xi2 --root

I don't really use any presentation software myself, but I'll try and give it a test myself when I get a few minutes.

CBiX:
No wayland here, xinput test works with normal input and input from the App's remote input, but doesn't show anything for "presentation remote" input except for the (working) "Exit presentation" button.

GSconnect debugger output:

Presentation remote:
> DEBUG: [core.js:receive/<:404]: FloS5: {
>   "id": 1542562335302,
>   "type": "kdeconnect.mousepad.request",
>   "body": {
>     "specialKey": 9
>   }
> }
> DEBUG: [mousepad.js:pressKeySym:262]: Mousepad: pressKeySym(65366, 0)
> DEBUG: [core.js:receive/<:404]: FloS5: {
>   "id": 1542562348014,
>   "type": "kdeconnect.mousepad.request",
>   "body": {
>     "specialKey": 8
>   }
> }
> DEBUG: [mousepad.js:pressKeySym:262]: Mousepad: pressKeySym(65365, 0)
> DEBUG: [core.js:receive/<:404]: FloS5: {
>   "id": 1542562349603,
>   "type": "kdeconnect.mousepad.request",
>   "body": {
>     "specialKey": 9
>   }
> }
> DEBUG: [mousepad.js:pressKeySym:262]: Mousepad: pressKeySym(65366, 0)
> DEBUG: [core.js:receive/<:404]: FloS5: {
>   "id": 1542562358827,
>   "type": "kdeconnect.mousepad.request",
>   "body": {
>     "specialKey": 25
>   }
> }
[...]

PR #436, opened 2019-01-07
Mousepad plugin - generate unicode with composition

getzze:
In Gnome, unicode character can be output by pressing Control+Shift+u, then hexadecimal code, then Enter (https://help.gnome.org/users/gnome-help/stable/tips-specialchars.html.en). This mimics this composition, both using At-spi and Caribou.

It has Left Control and Left Shift keycodes hardcoded. It should be fine for most users but not if these keys are used for another purpose (composing key for example).

It requires testing and should probably be optional and activable in the plugin settings.

Modifiers can be implemented like this too, so we can eventually get rid of Caribou (#340).

### DIFF
--- a/src/service/plugins/mousepad.js
+++ b/src/service/plugins/mousepad.js
@@ -38,6 +38,11 @@ var Metadata = {
 
 const _ASCII = /[\x20-\x7E]/;
 
+const XKEYCODE_Alt_L = 0x40;
+const XKEYCODE_Control_L = 0x25;
+const XKEYCODE_Shift_L = 0x32;
+const XKEYCODE_Super_L = 0x85;
+
 
 /**
  * A map of "KDE Connect" keyvals to Gdk
@@ -188,7 +193,7 @@ var Plugin = GObject.registerClass({
     }
 
     _handleInput(input) {
-        let mask = 0;
+        let mods = false;
 
         // These are ordered, as much as possible, to create the shortest code
         // path for high-frequency, low-latency events (eg. mouse movement)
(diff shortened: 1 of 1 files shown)


andyholmes:
Cool, I'll try and give this a test asap, but I think hold of until after v20 before merging. That should be in a just a few days, then I'll have closer look. Seems pretty straight forward though.

andyholmes:
By the way, the development tool is still there, it's just hidden from users since many were causing problems for themselves. You can launch it manually:

    gapplication action org.gnome.Shell.Extensions.GSConnect devel

andyholmes:
I pushed some changes; minor changes I guess was a bad commit title :P 

I added some phony `mode_lock()` functions for doing modifiers with keycodes. I'm not sure I understand this as well as you, but it seemed to work perfect for me. I guess that if this works Caribou is still required for Unicode with modifiers, but I couldn't find a way to do that with Android even with Hacker's Keyboard, so I'm not too worried.

Is there a reasonable way to detect if a keycode is being used as the compose key? I don't use this feature so I'm unfamiliar with it.

ferdnyc:
> Is there a reasonable way to detect if a keycode is being used as the compose key? I don't use this feature so I'm unfamiliar with it.

Probably not, but it also shouldn't be necessary since Compose isn't a modifier. Compose is a normal key that initiates a special input mode. (Handy because you don't have to hold it down while pressing the other keys; _especially_ handy since some of them have to be shifted.) That mode terminates when the Compose sequence is complete, which can involve anywhere from two to four (theoretically more, even) keypresses following the key bound to Compose itself.

I have my <kbd>Menu</kbd> key bound to Compose, so I type...
* <kbd>Menu</kbd><kbd>.</kbd><kbd>=</kbd>: •
* <kbd>Menu</kbd><kbd>-</kbd><kbd>-</kbd><kbd>.</kbd>: –
* <kbd>Menu</kbd><kbd>C</kbd><kbd>C</kbd><kbd>C</kbd><kbd>P</kbd>: ☭ (that one's still my favorite) :grin: 

Point is, whatever key is bound to Compose requires no special input handling.

ferdnyc:
I guess the only thing you _might_ have to detect is the fact that the Compose key is NOT being used as a modifier, even if it normally would be.

Like I said, I use <kbd>Menu</kbd>, which normally _is_ a modifier, I think? `mod3`, looks like, from the output of `xmodmap -pm`. But because I have it set to Compose...

```console
$ # While xmodmap -pk stills shows it as Multi_key:
$ xmodmap -pk |grep 135
    135    	0xff20 (Multi_key)	0xff20 (Multi_key)	0xff20 (Multi_key)	0xff20 (Multi_key)	

$ # It's not listed as bound to any modifier in xmodmap -pm:
$ xmodmap -pm
xmodmap:  up to 4 keys per modifier, (keycodes in parentheses):

shift       Shift_L (0x32),  Shift_R (0x3e)
lock        Caps_Lock (0x42)
control     Control_L (0x25),  Control_R (0x69)
mod1        Alt_L (0x40),  Alt_R (0x6c),  Meta_L (0xcd)
mod2        Num_Lock (0x4d)
mod3      
mod4        Super_L (0x85),  Super_R (0x86),  Super_L (0xce),  Hyper_L (0xcf)
mod5        ISO_Level3_Shift (0x5c),  Mode_switch (0xcb)
```

andyholmes:
Ok I see now. Well, I guess really anyone using a compose key will think of that key as a compose key, so maybe it's not a problem at all. The only corner case would be if they used eg Alt_L as their compose key and Alt_R as their "regular Alt", whereas for us Alt === Alt_L.

I just pushed an amendment to read keycodes from Gdk during `_init()` so hopefully that's one less corner case to deal with.

andyholmes:
I'm going to go ahead and merge this for now. I think it'll be awhile before the next release, but maybe we'll do another series of RC's so this and GVFS get some good testing.

PR #443, opened 2019-01-12
Allow sshfs to connect with old crypto protocols

pothos:
A mismatch between a new ssh installation on the desktop and
an older ssh on Android may lead to a broken sshfs connection.

Maybe you can consider even other protocols to allow, but this already works for my case where the annonced options of the Android device were:

```
HostKeyAlgorithms ssh-dss
KexAlgorithms diffie-hellman-group14-sha1,diffie-hellman-group1-sha1
MACs hmac-md5,hmac-sha1,hmac-md5-96,hmac-sha1-96
```

### DIFF
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -322,6 +322,10 @@ var Plugin = GObject.registerClass({
                 '-f',
                 // Do not use ~/.ssh/config
                 '-F', '/dev/null',
+                // Allow older crypographic protocols
+                '-o', 'HostKeyAlgorithms=+ssh-dss',
+                '-o', 'MACs=+hmac-sha1',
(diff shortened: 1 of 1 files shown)


andyholmes:
Support for older protocols like DSA/DSS was dropped on purpose in 9bad1a5 since upstream kdeconnect-android upgraded sshd-core to 0.14 in KDE/kdeconnect-android@526d500.

Is there some reason this doesn't work in kdeconnect-android v1.10+? In future sshfs support will be dropped in favour of GVFS in GSConnect, which won't support adjusting the key algorithms.

pothos:
Hi,

thanks for the response. In fact, the problem was F-Droid which did not update any apps anymore until I disabled and reenabled the repository…
So I'm closing this.

Regards,

Kai

PR #480, opened 2019-03-07
Add Ukrainian translation

kotyhoroshko2:
Hi. I tried to translate GSConnect to Ukrainian. It would be nice if this translation was uploaded to Crowdin and then reviewed by someone else.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -17,4 +17,5 @@ sk
 sr
 sr@latin
 tr
+uk
 zh_CN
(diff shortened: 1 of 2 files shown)


andyholmes:
I've added Ukrainian to crowdin and uploaded your existing translations. Crowdin uses different plurals form than most glibc projects, so those may need to be updated. I don't believe there are any other Ukrainian translators, but I will leave your translations unreviewed in case one volunteers.

PR #506, opened 2019-03-28
Create a symlink to FUSE dir with a stable name

mavit:
GVFS mounts filesystems on `/run/user/.../gvfs/sftp:host=...,port=...`, where the IP address and port number will vary.  Create a symbolic link in `/run/user/.../gsconnect/by-name/` so that the filesystem can be found in a predictable place.

### DIFF
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -350,7 +350,7 @@ var Plugin = GObject.registerClass({
         // Check for and remove any existing stale link:
         try {
             let link_stat = link.query_info(
-                Gio.FileAttributeType.symlink_target,
+                'standard::symlink-target',
                 Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
                 null
             );
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -337,8 +337,7 @@ var Plugin = GObject.registerClass({
         let safe_device_name = this.device.name.replace('/', '∕');
         if ( safe_device_name == '.' ) {
             safe_device_name = '·';
-        }
-        else if ( safe_device_name == '..' ) {
+        } else if ( safe_device_name == '..' ) {
             safe_device_name = '··';
         }
 
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -173,6 +173,7 @@ var Plugin = GObject.registerClass({
                 if (this._uri === uri) {
                     this._gmount = mount;
                     this._gmount.connect('unmounted', this.unmount.bind(this));
+                    this._addSymlink(mount);
 
                 // Or if it's a stale mount we need to cleanup
                 } else if (this._uriRegex.test(uri)) {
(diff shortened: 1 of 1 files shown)


src/service/plugins/sftp.js
```diff
@@ -317,6 +318,54 @@ var Plugin = GObject.registerClass({
         }
     }
 
+    /**
+     * Create a symbolic link referring to the device by name
+     */
+    _addSymlink(mount) {
```
andyholmes: Since this function does IO, make it an `async` function and wrap the whole body in a single try-catch.

src/service/plugins/sftp.js
```diff
@@ -317,6 +318,54 @@ var Plugin = GObject.registerClass({
...
+        let by_name_dir = Gio.File.new_for_path(
+            gsconnect.runtimedir + '/by-name/'
+        );
+        try {
+            by_name_dir.make_directory_with_parents(null);
+        } catch (e) {
+            if ( ! e.matches(Gio.IOErrorEnum, Gio.IOErrorEnum.EXISTS) ) {
+                throw e;
```
andyholmes: Since this function is not critical to the mount succeeding, it's probably best not to re-throw errors in this function. Instead of `throw e;` just use:

```js
debug(e, `${this.device.name}: ${this.name}`);
```
...or `warning()` if you think it's an error a user might want to have in the log.

src/service/plugins/sftp.js
```diff
@@ -317,6 +318,54 @@ var Plugin = GObject.registerClass({
...
+        }
+
+        let link_target = mount.get_root().get_path();
+        let link = Gio.File.new_for_path(
+            by_name_dir.get_path() + '/' + safe_device_name
+        );
+
+        // Check for and remove any existing stale link:
+        try {
+            let link_stat = link.query_info(
```
andyholmes: When `_addSymlink()` is made async, you can wrap this and `g_file_delete()` in small promises:

```js
let link_stat = await new Promise((resolve, reject) => {
    link.query_info_async(
        "standard::symlink-target",
        Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
        GLib.PRIORITY_DEFAULT,
        null,
        (link, res) => {
        try {
            resolve(link.query_info_finish (res));
        } catch (e) {
            reject(e);
        }
    );
});
```

In the case of `g_file_delete_async()` below, you can just use `debug()` on a NOT_FOUND error and call `resolve()` from the catch.

src/service/plugins/sftp.js
```diff
@@ -317,6 +318,54 @@ var Plugin = GObject.registerClass({
...
+                link.delete(null);
+            }
+        } catch (e) {
+            if ( ! e.matches(Gio.IOErrorEnum, Gio.IOErrorEnum.NOT_FOUND) ) {
+                throw e;
+            }
+        }
+
+        link.make_symbolic_link(link_target, null);
+    }
```
andyholmes: In the catch of the try-catch that wraps this function, you can just log the error as described above.

src/service/plugins/sftp.js
```diff
@@ -317,6 +318,54 @@ var Plugin = GObject.registerClass({
...
+                throw e;
+            }
+        }
+
+        // Replace path separator with a Unicode lookalike:
+        let safe_device_name = this.device.name.replace('/', '∕');
+        if ( safe_device_name == '.' ) {
+            safe_device_name = '·';
+        }
+        else if ( safe_device_name == '..' ) {
```
andyholmes: This needs to be on the same line as the bracket above or the linter will fail:

```js
if (cond) {
} else if (cond2) {
}
```

src/service/plugins/sftp.js
```diff
@@ -317,6 +318,54 @@ var Plugin = GObject.registerClass({
...
+        );
+
+        // Check for and remove any existing stale link:
+        try {
+            let link_stat = link.query_info(
+                'standard::symlink-target',
+                Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
+                null
+            );
+            if ( link_stat.symlink_target != link_target ) {
```
andyholmes: Missed this first time through. This should a method call, since it's not a GProperty: `link_stat.get_symlink_target()`

mavit:
Thanks for the review.  I think I've addressed all of your comments.

PR #594, opened 2019-08-18
LAN: do a better job ensuring IPv6 and IPv4 support

### LINKED ISSUE #593: Re-confirm simultaneous IPv4 & IPv6 support
Just wanted to say that this was my exact issue with connecting through a Wireguard VPN. I couldn't figure out why I wasn't getting a connection but it turned out that GSConnect on the desktop wasn't listening on IPv4 and therefore the VPN wouldn't facilitate the connection. Disabling IPv6 fixed the issue. Thank you!

_Originally posted by @decaby7e in https://github.com/andyholmes/gnome-shell-extension-gsconnect/issues/69#issuecomment-522130271_

andyholmes:
Since we don't use some of the helper functions in GSocketListener,
we need manually ensure we get socket bound to at least one of IPv4/6,
preferrably one that speaks both.

closes #593

### DIFF
--- a/src/service/protocol/lan.js
+++ b/src/service/protocol/lan.js
@@ -149,53 +149,92 @@ var ChannelService = class ChannelService {
     }
 
     _initUdpListener() {
-        this._udp = new Gio.Socket({
-            family: Gio.SocketFamily.IPV4,
-            type: Gio.SocketType.DATAGRAM,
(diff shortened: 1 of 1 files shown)


andyholmes:
Seems to work for me, so I'm merging.

Issue #620, opened 2019-09-12
Ship a desktop file & an icon for gsconnect


bilelmoussaoui:
**Describe the bug**
Currently, gsconnect can send notifications; which means it get displayed in the notifications settings 
![Screenshot from 2019-09-12 12-30-46](https://user-images.githubusercontent.com/7660997/64776852-4781b800-d559-11e9-8862-0e115a7908f9.png)

Without a desktop file & an icon, the GSconnect entry will look out of the place

**Steps To Reproduce:**
1. Open GNOME Control Center
2. Go to notifications
3. No Icon for GSConnect

**Expected behavior**

An icon would be displayed there. 

**System Details (please complete the following information):**

 - **GSConnect version:** 24
   - **Installed from:** GNOME Extensions Website
 - **GNOME/Shell version:** 3.32.2
 - **Distro/Release:** Fedora Silverblue

Add any addtional information about the problem or your system.

The desktop file doesn't necessarily need to have Display=true; it could be hidden of course.

PR #621, opened 2019-09-12
Switch .desktop file to symbolic icon

ferdnyc:
The only real use for the `.desktop` file is the Notifications source list, as pointed out in #620. There's precedent there, with the notifications from Control-Center panels (Color, Date & Time, Power), for using the icon that users will actually see and be familiar with.

Addresses #620


andyholmes:
I'm not really against this, but doesn't GNOME Shell decide whether to use a symbolic icon or not? I seem to recall different themes passing `-st-icon-stlye` in CSS.

I can't find a good reference, but there's [this](https://github.com/numixproject/numix-icon-theme/issues/485#issuecomment-291870434). It seems like falling back between symbolic and regular is the usual course of action. There are some other future concerns, but do you have any more thoughts on this?

ferdnyc:
> I'm not really against this, but doesn't GNOME Shell decide whether to use a symbolic icon or not? I seem to recall different themes passing `-st-icon-stlye` in CSS.
> 

For _apps_, I didn't **think** so... but I don't think there's really a concept of a "symbolic app icon" at all, and after looking at it I'm not sure exactly _how_ all that's supposed to work. `gnome-control-center` seems to just display symbolic icons for _its own_ items in its Notifications list, because those are the icons it displays elsewhere in the interface. But those entries do indeed have corresponding `.desktop` files with _non_-symbolic icons set, so yeah I guess it is left to the system to choose.

I'm uncertain enough about how all that's supposed to work that I'm going to close this, better to leave things as they are I think.

> I can't find a good reference, but there's [this](https://github.com/numixproject/numix-icon-theme/issues/485#issuecomment-291870434). It seems like falling back between symbolic and regular is the usual course of action. There are some other future concerns, but do you have any more thoughts on this?

andyholmes:
Okay, no worries, we can always revisit this. The newer symbolic icon doesn't actually match the full color one anymore, and the only other place it's used at all is the *About* dialog, so maybe we could just drop it one day and consider the symbolic icon the official icon.

PR #690, opened 2019-11-02
Update POT / PO files

### LINKED ISSUE #689: Acronyms and proper nouns should not be translated.
**Describe the bug**

MPRIS is an acronym for "The Media Player Remote Interfacing Specification" a DBus interface spec and should not be translated nor marked as translatable. It should be treated as a proper noun in this sense.

ferdnyc:
<strike>As was pointed out, `'MPRIS'` shouldn't be a translatable string, so I removed the `_()` wrapping.</strike>

I've rethought that, I think I was steered wrong. Ultimately it doesn't _matter_, since NONE of the translation files have `'MPRIS'` translated into anything else. But I've removed the commit that excluded it from the translation master.

But while I was in there, I regenerated the POT and updated the PO files anyway.

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: gsconnect\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-10-10 07:07-0400\n"
+"POT-Creation-Date: 2019-11-02 21:17-0400\n"
 "PO-Revision-Date: 2019-10-10 11:13\n"
 "Last-Translator: Andy Holmes (andyholmes)\n"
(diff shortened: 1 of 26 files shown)


ferdnyc:
Hmm. I wonder if I should insert a comment indicating _why_ the label isn't translated, now?

ferdnyc:
I've rethought that, I think I was steered wrong. Ultimately it doesn't _matter_, since NONE of the translation files have `'MPRIS'` translated into anything else. But I've removed the commit that excluded it from the translation master.

andyholmes:
Is this one all good to merge, or do you have more to push?

ferdnyc:
Nah, I wasn't planning on fiddling with it anymore. I'll merge it now.

Issue #697, opened 2019-11-06
Proposed re-ordering of items in the system-menu


amivaleo:
Hi!

This is how the gsconnect-items are displayed in the system-menu:

![Schermata da 2019-11-06 10-24-53](https://user-images.githubusercontent.com/24316547/68285516-b3772d00-007f-11ea-92e2-9be11fe26451.png)

I translate them:
- Messages
- Ring
- Photos
- Files >
- Send file

To me, they could ordered in a more convenient way, according to their "meaning".

This is what I suggest:
- Ring
- Messages
- Photos
- Send file
- Files >

I pretend the "Jimmy" label for this "issue", because it is a veeeery pedantic suggestion! ☺

ferdnyc:
Hmm. Prioritizing "Ring" doesn't make a lot of sense to me, it's a function I suspect only a tiny percentage of users _ever_ use even once. Messaging as the first item listed makes a lot of sense to me, as it's one of the extension's more popular features. I'd probably put either Photo or Share next, and stick either Ring or Mount way at the bottom (only because Mount is another small-percentage feature, _and_ it's the one that alters the menu itself).

...The biggest issue, though, is that I suspect most users are used to the menu in the order it's currently in, and arbitrarily shuffling it around on them might cause some annoyance. Nothing worse than when some software just decides to move everything around simply because it can.

amivaleo:
I feel like that "Ring" and "Files/Mount" are somewhat misplaced. "Ring" in particular, also for the reason you said: it's almost never used. And all the others entries belong to the same context, in some sense.

> Nothing worse than when some software just decides to move everything around simply because it can.

Yep, definetely. That's definitely a point, indeed.

🤔

I think this "issue" is maybe way too much pedantic maybe.
Gonna close this. 💪

Thank you, Frank, for being well-disposed at any time. ^_^

andyholmes:
FYI, you can actually re-order these by changing the dconf key for menu items:
https://github.com/andyholmes/gnome-shell-extension-gsconnect/blob/9b54aad174f97a012d8abe8fded8b377bee9aa71/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml#L54-L56

This was eventually intended to be exposed in some menu editor, it's just a bit complicated because some items are dynamic.

amivaleo:
Settings! I love settings.

But... Can't find that key in dconf-editor. 🤔

ferdnyc:
[...]

PR #722, opened 2019-11-30
Fix typo

Huluti:


ferdnyc:
@Huluti Ack! Yeah, that'll cause issues.

I was trying to figure out why it didn't bite me and I could still see Keyboard in the GSConnect menu, but then I realized that because of the discussion in #697 I still have a custom `menu-actions` defined!

Once I removed that, sure enough Keyboard was missing for me as well. Good catch, thanks! Squashing and merging.

ferdnyc:
(#697 also presents a workaround for anyone looking to fix this themselves so they don't have to wait fo the next version to be released. Because you'd have to not only edit the schema file but compile a new schema, to correct this value in the code, it's probably easier to just fix it in dconf:
```sh
# Get the identity of the device (or devices) paired to your computer:
$ dconf read /org/gnome/shell/extensions/gsconnect/devices
['1234567890abcdef', '9876543210fedcba']
# (You'll *probably* only have one, if so you can skip the loop
$ for d in (list bare device IDs here with spaces separating); do \
    dconf write /org/gnome/shell/extensions/gsconnect/device/${d}/menu-actions \
    '["sms", "ring", "mount", "commands", "share", "keyboard"]'; \
   done
# Then trigger a restart of the GSConnect daemon...
$ touch $HOME/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js
```

ferdnyc:
@andyholmes Looks like we're gonna need at least a `v29-rc3`...

ferdnyc:
@andyholmes Oops, never mind, I was behind on my notifications. Make that a bugfix `v30`, I guess.

andyholmes:
Did this actually throw an error or cause problems, or did it just not work?

ferdnyc:
It just didn't work, Keyboard doesn't appear in the menu. No harm other than missing functionality.

andyholmes:
Hmm, I guess I could release v30, but if anyone asks all these mistakes I keep making in odd-numbered releases are actually a homage to Star Trek movies :wink:

I'll give it until the morning to see if any other last minute bugs show up.

ferdnyc:
I'd offer to run point on `v31`, but I'm worried that would be the equivalent of letting Shatner direct.

PR #742, opened 2020-01-01
New translations org.gnome.Shell.Extensions.GSConnect.pot (German)

BjoernDaase:
Just a few random corrections and one addition to the German translation.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -3,8 +3,8 @@ msgstr ""
 "Project-Id-Version: org.gnome.Shell.Extensions.GSConnect\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2019-11-02 21:17-0400\n"
-"PO-Revision-Date: 2019-11-04 07:36\n"
-"Last-Translator: Andy Holmes (andyholmes)\n"
+"PO-Revision-Date: 2020-01-01 12:15+0100\n"
+"Last-Translator: Björn Daase (BjoernDaase)\n"
 "Language-Team: German\n"
 "Language: de_DE\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


po/de.po
```diff
@@ -1034,7 +1034,7 @@ msgstr "GSConnect"
 #. TRANSLATORS: Chrome/Firefox WebExtension description
 #: webextension/gettext.js:29
 msgid "Share links with GSConnect, direct to the browser or by SMS."
-msgstr "Verweise mit GSConnect direkt an den Browser oder per SMS freigeben."
+msgstr "Link mit GSConnect direkt an den Browser oder per SMS freigeben."
```
ferdnyc: (Disclaimer, up front: I don't speak German. I really have no idea what I'm talking about here. But that's never stopped me before...)

I wanted to ask about this one, only because Google Translate makes your version into this:
> Share the link with the GSConnect directly to the browser or via SMS.
whereas the previous translation reverses to:
> Share references with GSConnect directly to the browser or via SMS.

The English version is a description of GSConnect's features for operating on browser pages (or, links to them), in fact a more precise English description might be,
<blockquote>Share links <em><b>using</b></em> GSConnect, directly to the browser or in an SMS message.</blockquote>

(Google Translate makes <i>that</i> sentence, in German:)
> Teilen Sie Links mit GSConnect, direkt mit dem Browser oder per SMS.

I just wanted to double-check whether that would affect the translation at all.

BjoernDaase: > (Disclaimer, up front: I don't speak German. I really have no idea what I'm talking about here. But that's never stopped me before...)

No worries, feedback is always appreciated :)

> The English version is a description of GSConnect's features for operating on browser pages (or, links to them), in fact a more precise English description might be,
> 
>     Share links using GSConnect, directly to the browser or in an SMS message.

Ah, that definitely helps. Context is always good :+1: 

>     Share the link with the GSConnect directly to the browser or via SMS.
>     whereas the previous translation reverses to:
>     Share references with GSConnect directly to the browser or via SMS.

So both translations do not look like the original one after translating them back to English, correct?  
I actually did not look into the detail for this translation (because the sentence made sense) and just replaced "Verweis" with "Link"

> (Google Translate makes that sentence, in German:)

    Teilen Sie Links mit GSConnect, direkt mit dem Browser oder per SMS.

That translation is actually pretty good. I would make the following out of it:  

> Teilen Sie Links mitttels GSConnect, direkt an Browser oder per SMS.

PR #745, opened 2020-01-03
Fix notification not shown when payloadTransferInfo is empty

R1tschY:
Notification is not shown when send from KDE Connect Desktop / Sailfish Connect and there exists no icon for the notification.

Bug is caused by wrong detection of empty payload in notification plugin. KDE Connect Desktop / Sailfish Connect sends an empty payloadTransferInfo dict when there is no payload and `!{} === false`. Than strange things happen: binary gibberish is send in a new TCP connection to the other side.

### DIFF
--- a/src/service/plugins/notification.js
+++ b/src/service/plugins/notification.js
@@ -397,7 +397,7 @@ var Plugin = GObject.registerClass({
         let file, path, stream, success, transfer;
 
         try {
-            if (!packet.payloadTransferInfo || !packet.payloadTransferInfo.port) {
+            if (!packet.hasPayload()) {
                 return null;
             }
 
--- a/src/service/plugins/notification.js
+++ b/src/service/plugins/notification.js
@@ -397,7 +397,7 @@ var Plugin = GObject.registerClass({
         let file, path, stream, success, transfer;
 
         try {
-            if (!packet.payloadTransferInfo) {
+            if (!packet.payloadSize) {
                 return null;
             }
 
--- a/src/service/plugins/notification.js
+++ b/src/service/plugins/notification.js
@@ -397,7 +397,7 @@ var Plugin = GObject.registerClass({
         let file, path, stream, success, transfer;
 
         try {
-            if (!packet.payloadSize) {
+            if (!packet.payloadTransferInfo || !packet.payloadTransferInfo.port) {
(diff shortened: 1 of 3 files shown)


ferdnyc:
Re: Your second commit,
> `!payloadSize` cannot be used, because missing payloadSize means an payload stream without known size.

**_Does_** it? My understanding is that a stream with an unknown size will have a `payloadSize` of `-1`, but from what I can tell it shouldn't ever be _missing_. Is it? And, if it is... why?

In fact, _their_ implementation for `hasPayload()` returns the results of  testing `m_payloadSize != 0`, ensuring that a value of `-1` will produce a true result, along with any positive integers. (For `hasPayloadTransferInfo()` they get to cheat, because `QVariant` comes with an explicit `isEmpty()` method ready-made, so they can just use that for free.)

...Any reason we can't use a `payloadSize !== 0`  test, though, (philosophically) the same as the reference implementation? I'm not super wild about tying the determination about whether there's a  `payloadTransferInfo` attached to the presence of `payloadTransferInfo.port`, specifically, only because it ultimately seems like an arbitrary pick. Theoretically `payloadTransferInfo` could be redefined for some packets to make `port` optional, and then we'd be in trouble.

(I'll defer to @andyholmes for the final call on this, of course.)

src/service/plugins/notification.js
```diff
@@ -397,7 +397,7 @@ var Plugin = GObject.registerClass({
         let file, path, stream, success, transfer;
 
         try {
-            if (!packet.payloadTransferInfo) {
+            if (!packet.payloadSize) {
```
ferdnyc: Actually, since it's in `_downloadIcon()` specifically, I suppose this could even be
```js
if (packet.payloadSize === 0) { return null; }
```
Since you can't download a 0-byte payload.

ferdnyc: (Hmm. Would that do the right thing for streams with a `payloadSize` of `-1`, though?)

R1tschY:
@andyholmes Thanks for the function code. Added a check for `payloadSize == 0` like @ferdnyc said. 
I also found an additional code location with the wrong payload check.

ferdnyc:
Updated version LGTM!

Issue #218, opened 2018-10-03
branch wip-rewrite not working at all (for me)


mirko:
I know it's work-in-progress, but apparently this branch appears to work to some degree at least for some people - while it doesn't work for me at all.

I git-pull every couple of weeks, build it, install it, enable it in gnome-tweaks and.. nothing is happening. Also clicking the settings-icon for the extension just nothing is happening.

From what I read, it appears to work for some people at least - that's why I allow myself to open this ticket, even though explicitly declared as wip.

I'm fairly familiar with linux/programming and happy to help out / provide more information / debug output if requested.

Current version: 1d98a752c1207b699bfc8fa1f7efc4294f8a8709
OS: Debian testing/buster
Gnome: ii  gnome-shell    3.30.0-1     amd64

andyholmes:
> From what I read, it appears to work for some people at least - that's why I allow myself to open this ticket, even though explicitly declared as wip.

That's okay, this is a good time for testers to jump in, especially 3.30 users since I'm not there yet. Because of the commit 1d98a75 (about Aug. 22) I'm thinking you might be stuck on an old branch now.

A few weeks ago wip-rewrite was merged into master and then deleted, so I'm thinking your local clone still has the branch `wip-rewrite` which isn't been getting any new commits now. Could you try checking out the master branch and building from that?

mirko:
Ahhh, I missed the merge to master!
So I already used the new version for a while, without knowing (always switched back to master when I noticed the wip-rewrite branch doesn't work for me).
Thanks!

PR #781, opened 2020-03-09
WIP: Messaging: add URL previewer from Polari

andyholmes:
Add preview images for links in SMS messages.

See also: https://gitlab.gnome.org/GNOME/polari/-/merge_requests/134

### DIFF
--- a/src/utils/uri.js
+++ b/src/utils/uri.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const GLib = imports.gi.GLib;
+const Gio = imports.gi.Gio;
+const thumb = imports.service.ui.thumbnailer;
 
 
 /**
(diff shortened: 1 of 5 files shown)


andyholmes:
cc @jtojnar 

There's still a few more things I'll add tonight before I merge, but this is pretty much it for the script execution. Line `#218` of `uri.js` is where the call is made.

src/utils/uri.js
```diff
@@ -166,3 +167,72 @@ var SmsURI = class URI {
...
+
+        let nextData = this._urlQueue.shift();
+        if (nextData)
+            this._processData(nextData);
+    }
+
+    _generateThumbnail(data) {
+        let {filename, uri} = data;
+        this._subProc = Gio.Subprocess.new(
+            ['gjs', `${gsconnect.extdatadir}/utils/thumbnailer.js`, uri, filename],
```
jtojnar: Yeah, we prefer shebangs, as they are easier to patch automatically to a concrete path.

jtojnar: Though, why not just import that as a module? It should be possible to make it sync as well, should not it?

andyholmes: That's true, I've been playing with inlining it a bit. The catch is gjs still doesn't deal well with garbage collection complex objects like GtkWindow, so so far it has some pretty big memory increases if I run it in-process.

I'll work more at it and ping you again closer to merge-time.

andyholmes:
Time to close this one down, it's not going anywhere

PR #786, opened 2020-03-12
Remove deprecated public-name

ferdnyc:
For GSconnect v25, the `public-name` GSettings key was deprecated in favor of `name`.

As part of that deprecation, code was added to `daemon.js` to migrate `public-name` settings to `name`, with a TODO comment saying "remove after a few releases". It's been 9, seems like time.

This PR removes the migration code, and drops `public-name` from the schema, eliminating the last traces of that string from the GSConnect codebase.

### DIFF
--- a/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml
+++ b/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml
@@ -10,9 +10,6 @@
     </key>
 
     <!-- Service Settings -->
-    <key name="public-name" type="s">
-      <default>""</default>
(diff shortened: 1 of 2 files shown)


andyholmes:
Oops, totally forgot about that :laughing:

ferdnyc:
I actually spotted it in DConf Editor (because setting it to `""` still leaves the key in place), and was about to just delete it there when I noticed it was still part of the schema as well.

andyholmes:
Heh, I just realized I have a whole butt-load of `TODO` notes in GSConnect :P

PR #805, opened 2020-03-31
restore pulseaudio on 'nothing' volume option

koter84:
i'm forwarding call-audio to my fedora machine, so i want to be "notified" of an incoming call by lowering the audio-out, but when i answer the call i want the audio-out to be restored.

i replied to another issue with a bit more detailed description of how i'm forwarding the call-audio
https://github.com/andyholmes/gnome-shell-extension-gsconnect/issues/575#issuecomment-606514700

i'm not sure if this would we a logical thing to do, or that i should add an option to 'restore', instead of using the 'nothing' option for this.

### DIFF
--- a/data/ui/device-preferences.ui
+++ b/data/ui/device-preferences.ui
@@ -2476,6 +2476,13 @@
         <attribute name="target">nothing</attribute>
       </item>
       <item>
+        <!-- TRANSLATORS: Restore the system volume -->
+        <attribute name="label" translatable="yes">Restore</attribute>
+        <attribute name="icon">audio-volume-medium-symbolic</attribute>
+        <attribute name="action">settings.ringing-volume</attribute>
+        <attribute name="target">restore</attribute>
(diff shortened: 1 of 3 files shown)


andyholmes:
Seems like a reasonable feature I won't ever use myself :D 

Yeah, I think a `restore` option would be less confusing than re-using `nothing`.

You'll also want to add a menu item UI definition [here](https://github.com/andyholmes/gnome-shell-extension-gsconnect/blob/master/data/ui/device-preferences.ui#L2495). The gettext parser doesn't work properly right now, but you can just add a section like this to the [POTFILE](https://github.com/andyholmes/gnome-shell-extension-gsconnect/blob/master/po/org.gnome.Shell.Extensions.GSConnect.pot) to get it translated:

```
#. TRANSLATORS: Restore the system volume
msgid "Restore"
msgstr ""
```

Make those changes and I'll happily merge :)

PR #831, opened 2020-04-26
SFTP Plugin: refactor for better stability

### LINKED ISSUE #478: Mount problem
So I just installed GSConnect and everything works perfect, except for the mount option. I am able to mount my phone, but after some time it unmounts automatically, often when I'm trying to transfer some files from my phone to my pc. If this is the case (I'm copying files) I get "error splicing file: input/output error", which I guess is just the consequence of the phone being unmounted. I have all the dependencies installed. Any ideas?

andyholmes:
Refactor to watch the global GVolumeMonitor instead of the result of
GMount operations and the eventual GMount::unmounted signals.

This takes advantage of a number of quirks that occur with the Android
SFTP server, allowing mounts to survive device connection breaks and be
reused in several cases.

closes #478

### DIFF
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -34,6 +34,9 @@ var Metadata = {
 };
 
 
+const MAX_MOUNT_DIRS = 12;
+
+
 /**
  * SFTP Plugin
  * https://github.com/KDE/kdeconnect-kde/tree/master/plugins/sftp
@@ -46,26 +49,51 @@ var Plugin = GObject.registerClass({
     _init(device) {
         super._init(device, 'sftp');
 
+        this._gmount = null;
+        this._mounting = false;
+
         // A reusable launcher for ssh processes
         this._launcher = new Gio.SubprocessLauncher({
             flags: (Gio.SubprocessFlags.STDOUT_PIPE |
                     Gio.SubprocessFlags.STDERR_MERGE),
         });
 
-        this._mounting = false;
+        // Watch the volume monitor
+        this._volumeMonitor = Gio.VolumeMonitor.get();
+
+        this._mountAddedId = this._volumeMonitor.connect(
+            'mount-added',
+            this._onMountAdded.bind(this)
+        );
+
+        this._mountRemovedId = this._volumeMonitor.connect(
+            'mount-removed',
+            this._onMountRemoved.bind(this)
+        );
     }
 
-    get info() {
-        if (this._info === undefined) {
-            this._info = {
-                directories: {},
-                mount: null,
-                regex: null,
-                uri: null,
-            };
+    get gmount() {
+        if (this._gmount === null && this.device.connected) {
+            let host = this.device.channel.host;
+
+            let regex = new RegExp(
+                `sftp://(${host}):(1739|17[4-5][0-9]|176[0-4])`
+            );
+
+            for (let mount of this._volumeMonitor.get_mounts()) {
+                let uri = mount.get_root().get_uri();
+
+                if (regex.test(uri)) {
+                    this._gmount = mount;
+                    this._addSubmenu(mount);
+                    this._addSymlink(mount);
+
+                    break;
+                }
+            }
         }
 
-        return this._info;
+        return this._gmount;
     }
 
     handlePacket(packet) {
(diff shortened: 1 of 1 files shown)


andyholmes:
This seems much better now, but I'd still want some user testing before I merge it, since I don't really use this feature much.

ferdnyc:
@andyholmes 

I've just installed the PR branch and will kick the tires some as soon as I can. The only barrier is that I've got a bunch of things open that I need to finish up and close out first, since I need to reboot (or at least log-out-log-in) to get the extension working again.

(Restarting the `gnome-shell` process isn't an option currently, as there's some sort of bug I've recently discovered with the combination of X.org session and Nvidia binary drivers. After I trigger an <kbd>F2</kbd>+<kbd>r</kbd> restart of the Shell process, somehow I guess the lockscreen blur effect gets stuck in an enabled state, and constantly keeps blurring all of the running applications' windows into unreadability. So, shell restarts are useless even in non-Wayland sessions now, yay!)

Aaanyway, I'll get around to restarting at **some** point this afternoon.

andyholmes:
> I've just installed the PR branch and will kick the tires some as soon as I can.

Much appreciated! You can always do it the old-fashioned way too :wink: 

```sh
$ cd ~/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service
$ gjs daemon.js
```

I've been starting it that way a lot lately, actually, as you can do fun things like:

```sh
$ gdb --args gjs daemon.js
$ sysprof-cli --gjs --no-perf -- gjs daemon.js
$ psrecord --interval 1 --plot plot.png "gjs daemon.js"
```

I have some (eventual) plans to revive the old devel tool, as an repo-only tool for nerds & devs.

ferdnyc:
> ways do it the old-fashioned way too 
> 
> ```shell
> $ cd ~/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service
> $ gjs daemon.js
> ```

Well, it was actually the _extension_ that's in an Error state in GNOME Shell (still is, actually)*, which means I don't have any of the user menu controls.

But, I finally did realize that I could run the daemon, as you say, and then use `gnome-extensions prefs gsconnect@...` to get to the UI, and then I was able to use the hamburger menu inside the device settings to trigger a mount.

Overall it seems... better? I think better, anyway.

Still didn't stop me from getting disconnected even though I was right in the middle of a long file transfer (PC -> Mobile), unfortunately, but that seems to be just kind of the norm, Android is definitely not a server platform.

But browsing the remote filesystem feels nice and stable, and when I do get disconnected GSConnect notices and reverts back to the unmounted state immediately. At least so far, no more bouncing or thinking the mount is up when it isn't, both of which have been an issue in the past. :+1: 

\* – (Still haven't gotten around to rebooting. I've still got a shell open inside a Docker image of Python's manylinux2014, and Qt is busy compiling... it's a whole thing.)

ferdnyc:
After bookmarking the device root in Nautilus, I was also able to reconnect to it just by clicking on the bookmark, and GSConnect picked up the mounted state as soon as the connection came back up. I don't recall that working very well in the past, so that's cool.

(Obviously, that's kind of fragile as it relies on both the IP and port# not changing. But it is what it is. ...Unless we come up with a way to register `gsconnect:$DEVICE` or w/ever URLs with Gio/Gvfs, so we can address the connection logically instead of having to expose the underlying SFTP link, I suppose...)

andyholmes:
> Overall it seems... better? I think better, anyway.

Much appreciation for the testing. I think as long as it's not worse, the approach is more efficient so I'll merge :)

> Still didn't stop me from getting disconnected even though I was right in the middle of a long file transfer (PC -> Mobile), unfortunately, but that seems to be just kind of the norm, Android is definitely not a server platform.

Definitely not a server platform :) I think the feature is obviously convenient, but also better tools exist for this job.

> After bookmarking the device root in Nautilus, I was also able to reconnect to it just by clicking on the bookmark, and GSConnect picked up the mounted state as soon as the connection came back up. I don't recall that working very well in the past, so that's cool.

There is that code that @mavit wrote which dumps a named symlink in the runtime directory. We might be able to spruce that up, possibly to add an automatic bookmark or something? Unfortunately I don't think Gvfs allows "out of tree" extension modules.

We can register a `gsconnect://` URI scheme with Gvfs, but that only works in *our* process so I don't think it's really that useful. We could also register as a URI *handler* for `gsconnect://` scheme, but that would basically look like:

    Nautilus address bar -> gsconnect://foo/bar/ -> GSConnect service -> Default file manager

Kind of the opposite of seamless :)

PR #874, opened 2020-06-12
Improve Date format in messaging UI.

dmeltzer:
This makes two changes:
1) The year is now shown in messages that are not from the current year.
2) The tooltip when hovering over a message now shows a more detailed
locale-formatted date+time string.


andyholmes:
LGTM, thanks :)

PR #875, opened 2020-06-12
Group Message Improvements.

dmeltzer:
1) Show a list of all group participants in conversation list tooltip.
2) Show name of message sender in conversation view if there are
multiple participants in conversation.


andyholmes:
Would you mind giving me a screenshot for this one? I think I see what's happening here, but don't have any group messages to have a look at the result.

PR #879, opened 2020-06-17
Fix incorrect call in sort method.

### LINKED ISSUE #877: Message history messed up since 3392fbd
Since 3392fbd3360cdf34ba79ceeed6c747759fa0d366, the SMS history as well as message scrolling is broken and displays what looks like unsorted messages.

What I saw in the log:
```
JS WARNING: [/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/ui/messaging.js 712]: reference to undefined property "date"
```
Also this while scrolling:
```
Drawing a gadget with negative dimensions. Did you forget to allocate a size? (node row owner GtkListBoxRow)
```

dmeltzer:
This was overlooked when we moved from a date property on the row to the
date being part of the message property on the row.

Fixes #877

### DIFF
--- a/src/service/ui/messaging.js
+++ b/src/service/ui/messaging.js
@@ -709,7 +709,7 @@ const ConversationWidget = GObject.registerClass({
     }
 
     _sortMessages(row1, row2) {
-        return (row1.date > row2.date) ? 1 : -1;
+        return (row1.message.date > row2.message.date) ? 1 : -1;
     }
 
     /**


ferdnyc:
@dmeltzer The `_sortMessages()` change looks good, but isn't `_sortThreads()` in the same boat?

https://github.com/andyholmes/gnome-shell-extension-gsconnect/blob/fcaee34097bc433fda5ac4a26c4cc071be31f514/src/service/ui/messaging.js#L1046-L1048

dmeltzer:
> @dmeltzer The `_sortMessages()` change looks good, but isn't `_sortThreads()` in the same boat?

I looked at this because it confused me as well.  The ThreadRow class has a date getter on it that delegates to the message internally.  I wonder if it makes more sense to add a date getter to the conversation that does the same thing to keep the code consistent across both places.

ferdnyc:
@dmeltzer That sounds like it makes sense.

The really confusing thing is, `this.message()` is a getter for the `ThreadRow`'s `_message` property, so it looks like it could really go either way when it comes to threads.

But I guess it makes sense to mirror that same redundancy in the message rows, especially if those rows are going to be sorted by `date`. If `message.date` is a sort key for the list, seems a bit weird to have it buried down in a sub-sub-property.

andyholmes:
LGTM

>  I wonder if it makes more sense to add a date getter to the conversation that does the same thing to keep the code consistent across both places.

You're can to do that if you want :) You're always welcome to refactor stuff like that and include it as a separate commit in a PR. Much of this code hasn't had a top-to-bottom review in over a year, so internally might not be so consistent.

PR #903, opened 2020-07-31
README: Emphasize that GSConnect replaces KDEConnect

doronbehar:
Closes https://github.com/andyholmes/gnome-shell-extension-gsconnect/issues/901 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
-[GSConnect][ego] is a complete implementation **replacing** desktop [KDE
-Connect][kdeconnect]. It includes GNOME Shell, Nautilus, [Chrome][chrome] and
-[Firefox][firefox] integrations.
+[GSConnect][ego] is a complete implementation of [KDE Connect][kdeconnect]
+especially for GNOME Shell with Nautilus, [Chrome][chrome] and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,6 @@
-[GSConnect][ego] is a complete implementation of [KDE Connect][kdeconnect] for
-GNOME Shell with Nautilus, [Chrome][chrome] and [Firefox][firefox] integration.
+[GSConnect][ego] is a complete implementation **replacing** desktop [KDE
+Connect][kdeconnect]. It includes GNOME Shell, Nautilus, [Chrome][chrome] and
+[Firefox][firefox] integrations.
```
andyholmes, line 3: I'd prefer if this were the same as the extensions.gnome.org text:

```
GSConnect is a complete implementation of KDE Connect especially for
GNOME Shell with Nautilus, Chrome and Firefox integration. It does
not rely on the KDE Connect desktop application and will not work
with it installed.
```

ferdnyc:
If only those cowards at the W3C hadn't deprecated the `<blink>` tag...

README.md
doronbehar: OK I guess `will not work with it installed` is good enough.

PR #913, opened 2020-08-13
Enable word-wrap in the Remote Keyboard TextView widget

isantop:
Currently this widget does not have wrap enabled, which means that with very long text entry, the widget grows in width indefinitely until the user pressed enter (which also inserts a new line on the remote device). This can be annoying, as the window may grow to a length which is wider than the screen.

Instead, this allows long lines of text to wrap independently of newlines and makes entering long paragraphs of text much simpler and more pleasant. Specifically, it enabled the `WORD_CHAR` wrap mode, meaning that text will first wrap to word boundaries, falling back to characters if the word is longer than the textview width.

The window can still be resized for better fit text in the TextView, and for easier entry of longer text.

Before this change:
Screenshot from 2020-08-13 13-27-08

After:
Screenshot from 2020-08-13 13-33-23

### DIFF
--- a/src/service/plugins/mousepad.js
+++ b/src/service/plugins/mousepad.js
@@ -474,7 +474,8 @@ var KeyboardInputDialog = GObject.registerClass({
             border_width: 6,
             hexpand: true,
             vexpand: true,
-            visible: true
+            visible: true,
+            wrap_mode: Gtk.WrapMode.WORD_CHAR
         });
         scroll.add(this.text);
 


ferdnyc:
Heh, I clearly don't use the remote keyboard enough, because I never noticed that and **wow** is it annoying! This fixes it handily, thanks.

Merging, and pinging @andyholmes just so this doesn't sneak past his notice.

ferdnyc:
(Who the heck came up with `Gtk.WordWrap.WRAP_NONE` as the default, anyway!?!) :stuck_out_tongue_closed_eyes:

andyholmes:
Who's been contributing fixes behind my back - I'll maintainerize 'em! :wink:

PR #914, opened 2020-08-14
eslintrc: add most rules from upstream GJS

andyholmes:
Add most rules from GJS's eslintrc, except the ones that are most
annoying to fix (at the moment).


andyholmes:
This should be all good, I just wanted it as a PR in case I needed to revert.

PR #942, opened 2020-09-10
Service UI: improve the error dialog UX

andyholmes:
Improve the error dialog shown for unexpected errors, using similar
patterns to the dialog used by `gnome-extensions`.


andyholmes:
Example:

Screenshot from 2020-09-10 16-12-34

PR #965, opened 2020-09-20
Explicitly place our menu below the network menu

Pobega:
On systems where an extension has placed itself at a higher position
than the network menu (for example using an extension that adds more
controls to the volume slider) the AggregateMenu index of '4' would not
be where you expect it.

This change makes it so that we figure out exactly where the network
menu index is and place ourselves right below it.

### DIFF
--- a/src/extension.js
+++ b/src/extension.js
@@ -93,7 +93,11 @@ const ServiceIndicator = GObject.registerClass({
         this._item.label.clutter_text.x_expand = true;
         this.menu.addMenuItem(this._item);
 
-        AggregateMenu.menu.addMenuItem(this.menu, 4);
+        // Find current index of network menu
+        let menuItems = AggregateMenu.menu._getMenuItems();
+        let networkMenuIndex = menuItems.indexOf(AggregateMenu._network.menu) || 3;
+        // Place our menu below the network menu
+        AggregateMenu.menu.addMenuItem(this.menu, networkMenuIndex + 1);
 
         // Service Menu -> Devices Section
         this.deviceSection = new PopupMenu.PopupMenuSection();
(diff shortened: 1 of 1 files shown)


Pobega:
Screenshot from 2020-09-20 08-01-01
This is what happens on my system without it. This is the same fix we discussed a few weeks ago in the Gnome extensions Matrix chat, currently used in my extension without issues; https://github.com/Pobega/gnome-shell-extension-mullvad-indicator/blob/master/extension.js#L95

(you can also see my extension in the screenshot)

Pobega:
Forgot to lint :')

src/extension.js
```diff
@@ -93,7 +93,11 @@ const ServiceIndicator = GObject.registerClass({
         this._item.label.clutter_text.x_expand = true;
         this.menu.addMenuItem(this._item);
 
-        AggregateMenu.menu.addMenuItem(this.menu, 4);
+        // Find current index of network menu
+        let menuItems = AggregateMenu.menu._getMenuItems();
+        let networkMenuIndex = menuItems.indexOf(AggregateMenu._network.menu) || 3;
```
andyholmes, line 98: It's probably a rare case, but if the item is not in the menu `indexOf()` will return `-1` which is actually a truthy value so the expression will never evaluate to `3`.

Maybe remove the `||` and add a second line:

```js
let menuIndex = networkMenuIndex > -1 ? networkMenuIndex : 3;
```

andyholmes:
Thanks! Just that one note and LGTM.

src/extension.js
Pobega: Oof, good catch, thanks!

Pobega: Should be resolved by e71a54e

andyholmes: LGTM, thanks!

PR #983, opened 2020-09-28
enable 'prefer-const' in eslint

Pobega:
Gjs style guidelines state that `const` should be preferred but it was missing from `eslint`.

https://gjs.guide/guides/gjs/style-guide.html#const-vs-let

I realized in retrospect my previous PR broke this rule so figured I'd push this your way in case you want it.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -192,6 +192,7 @@
         "object-curly-spacing": "error",
         "operator-assignment": "error",
         "operator-linebreak": "error",
+        "prefer-const": "error",
         "prefer-numeric-literals": "error",
         "prefer-promise-reject-errors": "error",
         "prefer-rest-params": "error",
--- a/installed-tests/fixtures/backend.js
+++ b/installed-tests/fixtures/backend.js
@@ -271,6 +271,7 @@ var ChannelService = GObject.registerClass({
         try {
             // Try to parse strings as <host>:<port>
             if (typeof address === 'string') {
+                // eslint-disable-next-line prefer-const
                 let [host, port] = address.split(':');
                 port = parseInt(port) || this.port;
                 address = Gio.InetSocketAddress.new_from_string(host, port);
--- a/installed-tests/fixtures/backend.js
+++ b/installed-tests/fixtures/backend.js
@@ -86,10 +86,10 @@ var ChannelService = GObject.registerClass({
 
     async _onIncomingChannel(listener, connection) {
         try {
-            let host = connection.get_remote_address().address.to_string();
+            const host = connection.get_remote_address().address.to_string();
 
             // Create a channel
-            let channel = new Channel({
+            const channel = new Channel({
                 backend: this,
                 host: host,
                 port: this.port,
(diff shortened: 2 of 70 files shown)


Pobega:
Also might consider `no-var` as well; https://eslint.org/docs/rules/no-var

Is there an upstream eslintrc yours is forked from that I should submit this to instead? I tried upstream's and it was different than the one in this repo (I actually switched from upstream's to yours since I found it works better)

Pobega:
There are some hairy lints in:

```
/var/home/pobega/Documents/gnome-shell-extension-gsconnect/installed-tests/fixtures/backend.js
  274:22  error  'host' is never reassigned. Use 'const' instead  prefer-const

/var/home/pobega/Documents/gnome-shell-extension-gsconnect/src/service/backends/lan.js
  402:22  error  'host' is never reassigned. Use 'const' instead  prefer-const

/var/home/pobega/Documents/gnome-shell-extension-gsconnect/src/service/ui/contacts.js
  99:5  error  'pixbuf' is never reassigned. Use 'const' instead  prefer-const

/var/home/pobega/Documents/gnome-shell-extension-gsconnect/src/service/utils/uri.js
  85:9  error  'res' is never reassigned. Use 'const' instead  prefer-const

/var/home/pobega/Documents/gnome-shell-extension-gsconnect/webextension/gettext.js
  69:11  error  'iter' is never reassigned. Use 'const' instead  prefer-const
```

Let me know if this change is something you even want here and I can go through them and fix them.

andyholmes:
Thanks for looking into this! `prefer-const` is a good idea, there are just a few rules in the upstream eslintrc I had skipped since they were big changes I hadn't gotten around to testing. I believe that was before I had written any unit tests, so it should be easier now.

For reference [this](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/5295341/merged-eslint.txt) what the eslint would look like merged with upstream's. Some of those I don't care for, but we can discuss each if you want to. `no-var` is probably fine, but it's not in upstream's eslint and I think the only places we use them are to take advantage of hoisting on purpose, for example:

```js
try {
    var FooImport = import.gi.Foo;
} catch (e) {
    FooImport = null;
}
```
So I think if we add that rule, we'll just have to add exceptions anywhere we use it any ways, but :shrug:. Here's what I'd recommend for the current lint errors:

1. `// eslint-disable-next-line prefer-const`
2. `// eslint-disable-next-line prefer-const`
3. move the `pixbuf` assignment to line 99 (eg. `const pixbuf = loader.get_pixbuf();`
4. break the two variable assignments onto separate lines
5. break the two variable assignments onto separate lines

Pobega:
I pushed those as 0655dfb.

I see no reason to include `no-var` now that I'm thinking about it, since there are legitimate reasons to use it and it's a pretty well known rule that you shouldn't use it if you don't have to (as in it would get caught in manual code review anyway.) And as for the upstream rules I'm not particularly opinionated, I just thought `prefer-const` would be a nice-to-have :)

andyholmes:
```js
const [host, portstr] = address.split(':');
const port = parseInt(portstr) || this.port;
```

This looks good to me :) I think I'll squash the whole thing into something like "Use `prefer-const`".

Pobega:
Great, pushed as 4f84cef. Appreciate the feedback

cbix:
Sorry if this is too off-topic, but I'm new to gjs (old to js, tho) and would love to see your eslintrc made to an up-to-date eslint plugin! Only found [eslint-plugin-gjs](https://www.npmjs.com/package/eslint-plugin-gjs) but that's apparently been abandoned and doesn't work at all anymore.

Great project still, been using gsconnect for years ❤

andyholmes:
Probably best just to use the upstream [eslintrc](https://gitlab.gnome.org/GNOME/gjs/-/blob/master/.eslintrc.yml).

PR #985, opened 2020-10-01
More verbose logging when `ssh-keygen -R` fails

mavit:
This can happen if, e.g., `~/.ssh/known_hosts` contains any line that is not recognised as valid by the current version of OpenSSH.

### DIFF
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -325,7 +325,7 @@ var Plugin = GObject.registerClass({
                     });
                 });
             } catch (e) {
-                debug(e, this.device.name);
+                logError(e, this.device.name + ': `' + argv.join(' ') + '`');
             }
         }
     }
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -308,12 +308,14 @@ var Plugin = GObject.registerClass({
      */
     async _removeHostKey(host) {
         for (let port = 1739; port <= 1764; port++) {
+            let argv;
             try {
-                const ssh_keygen = this._launcher.spawnv([
+                argv = [
                     Config.SSHKEYGEN_PATH,
                     '-R',
                     `[${host}]:${port}`,
-                ]);
+                ];
+                const ssh_keygen = this._launcher.spawnv(argv);
 
                 await new Promise((resolve, reject) => {
                     ssh_keygen.wait_check_async(null, (proc, res) => {
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -308,26 +308,34 @@ var Plugin = GObject.registerClass({
      */
     async _removeHostKey(host) {
         for (let port = 1739; port <= 1764; port++) {
-            let argv;
             try {
-                argv = [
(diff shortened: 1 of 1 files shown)


mavit:
> The eslint error is "328:56 error 'argv' is not defined no-undef"

Oops, thanks.  Fixed.

> even though the failure isn't fatal or even problematic really.

In circumstances that trigger it, it’s fatal to SFTPing, and requires manual user intervention to fix.

> But the way you're building the message doesn't totally work in JavaScript.  In particular, `` ` `` characters signify JS template strings — analogous to Python `f" "` strings, as one example — so they shouldn't appear in the output.)

I don't think I follow this argument.  The difference to me seems mainly to be one of style.

andyholmes:
> I don't think I follow this argument. The difference to me seems mainly to be one of style.

I think it just looks confusing, but it does work. Template strings in JavaScript can have inlined code and other quotations, but I think you can still put backticks *inside* regular quotations, you just can't put backticks in backticks  without escaping them

```js
let foo = `\`"Equation" 1 + 2 = ${Math.floor(1 + 2)}\``;
```

In any case, I think we might as well go all out here and return the `stderr` output when there's an error:

```js
try {
    const ssh_keygen = this._launcher.spawnv([
        Config.SSHKEYGEN_PATH,
        '-R',
        `[${host}]:${port}`,
    ]);

    await new Promise((resolve, reject) => {
        ssh_keygen.communicate_utf8_async(null, null, (proc, res) => {
            try {
                const stdout = proc.communicate_utf8_finish(res)[1];
                const status = proc.get_exit_status();

                if (status !== 0) {
                    throw new Gio.IOErrorEnum({
                        code: Gio.io_error_from_errno(status),
                        message: `${GLib.strerror(status)}\n${stdout}`.trim(),
                    });
                }

                resolve();
            } catch (e) {
                reject(e);
            }
        });
    });
} catch (e) {
    logError(e, this.device.name);
}
```

Which gives output like:

```sh
Gjs-WARNING **: 15:39:42.899: JS ERROR: Device Name: Gio.IOErrorEnum: Operation not permitted
/home/andrew/.ssh/known_hosts:199: invalid line
/home/andrew/.ssh/known_hosts is not a valid known_hosts file.
Not replacing existing known_hosts file because of errors
```

I think that's probably more helpful, thoughts? I'm assuming if there's a problem with the `argv`, that `ssh-keygen` will report that on `stderr` which save us from doing it manually.

mavit:
Yes, this looks like a better outcome, to me.

I must admit I don't understand how `ssh-keygen`'s stderr can be read here as stdout, though, although I am not familiar with Gio.

andyholmes:
> I must admit I don't understand how ssh-keygen's stderr can be read here as stdout, though, although I am not familiar with Gio.

It's not normally, but the existing flags are set to merge `stderr` into `stdout`:
https://github.com/andyholmes/gnome-shell-extension-gsconnect/blob/a00c4453f6c53e80c9f8b682879509ad3a73aa61/src/service/plugins/sftp.js#L57-L60

andyholmes (approved):
LGTM

PR #1000, opened 2020-10-23
Fix contacts not appearing in Messaging window

sigmaSd:
I'm not used at all to javascript, but testing the latest release of gsconnect I had to apply this patch for the contacts to show up. (it errored with missing getter for read-only property avatar)

### DIFF
--- a/src/service/ui/contacts.js
+++ b/src/service/ui/contacts.js
@@ -300,15 +300,6 @@ const AddressRow = GObject.registerClass({
     GTypeName: 'GSConnectContactsAddressRow',
     Template: 'resource:///org/gnome/Shell/Extensions/GSConnect/ui/contacts-address-row.ui',
     Children: ['avatar', 'name-label', 'address-label', 'type-label'],
-    Properties: {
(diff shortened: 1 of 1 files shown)


andyholmes:
Does the error disappear if you instead just remove the GObject property map?

https://github.com/andyholmes/gnome-shell-extension-gsconnect/blob/109586a7803a083d80fa7519edff810f575a18cb/src/service/ui/contacts.js#L303-L311

In the change you made, you've overriden the avatar widget in the template file:

https://github.com/andyholmes/gnome-shell-extension-gsconnect/blob/109586a7803a083d80fa7519edff810f575a18cb/data/ui/contacts-address-row.ui#L21-L24

sigmaSd:
Yes removing the GObject property map also fixes the issue.

andyholmes:
Okay, if you can make that the change instead I will happily merge it.

sigmaSd:
Done

andyholmes:
Thanks!

sigmaSd:
Thanks as well!

PR #1029, opened 2020-12-12
replace On/Off labels in notification preferences with gtk switch

premek:
To get to know the project a little I decided to implement a change in UI where there was a little inconsistency of using On Off labels in *Notification Preferences* while Gtk Switch is used everywhere else. I hope it's ok.

I did it the same as it is in Plugins in the Advanced menu

Screenshot from 2020-12-12 01-28-04

 _('On')  and  _('Off')  are probably not used anywhere else, is there anything needed to do with translations?


andyholmes:
Thanks for the contribution. I won't be reviewing this myself since I'm close to archiving the project, but if someone takes over maintainership they can review it then.

daniellandau:
Seems to work, thanks!

Issue #872, opened 2020-06-08
Custom battery level alert when charging


noccy80:
**Is your feature request related to a problem? Please describe.**
I try to avoid charging my phone to 100% to preserve battery health, but at times I forget to check the phone and end up getting the notification from GSC that my battery is fully charged.

**Describe the solution you'd like**
A toggle and an input to specify a percentage at which you want to receive a notification, f.ex. 80%. The notification should only show up if the device is charging.

**Describe alternatives you've considered**
I've tried apps to do it on the phone, but they always come with extra crap that might affect performance and require permissions.

**System Details (please complete the following information):**
 - GSConnect version: 38
 - Installed from: extensions.gnome.org

**Additional context**

andyholmes:
I probably won't add this this feature myself, but I would accept a patch for it.

andyholmes:
Sorry for the late reply.

Neither are very difficult, but I don't have any plans to add UI features myself as I don't have as much time to dedicate to GSConnect. If someone contributes a patch, I'd probably accept it as I assume they'll also have time/incentive to maintain it.

daniellandau:
Fix has been merged.

PR #1038, opened 2021-01-03
Add custom charge level notifications #872

amija004:
#872 
I added two toggles for 60% and 80% battery level notifications while device is charging. Testing was done on my laptop, ASUS FA506IV running Pop!_OS 20.04, connecting to my Pixel 3A running GrapheneOS. Virtual Pixel 3A using genymotion was also functional.

I attempted to add a slider or spinbutton to add a custom charge level, but could not figure out how to make it work. Glade kept breaking the UI files just by opening them, and I couldn't find a slider or spinbutton in the repo that I could copy.

If someone could give me some pointers on how to implement a custom level, I'd greatly appreciate it, as that would be much more useful. I could not get the schema to accept a key with an integer type, nor could I get the spinbutton or the slider to update the key value. I had been editing the UI files by hand, but I assume that there is a better method, as all the documentation I could find either uses Glade or compiles them.

### DIFF
--- a/data/ui/preferences-device-panel.ui
+++ b/data/ui/preferences-device-panel.ui
@@ -732,7 +732,6 @@
                                         <property name="vexpand">True</property>
                                         <property name="label" translatable="yes">Set Custom Charge Level</property>
                                         <accessibility>
-                                          <relation type="label-for" target="battery-custom-notification-spin-row"/>
                                           <relation type="label-for" target="battery-custom-notification-spin"/>
                                         </accessibility>
                                       </object>
@@ -748,10 +747,6 @@
                                         <property name="valign">right</property>
                                         <property name="adjustment">custom-battery-adjustment</property>
                                         <signal name="value-changed" handler="_newCustomChargeLevel">
-                                        <accessibility>
-                                          <relation type="controlled-by" target="battery-custom-notification-spin-row"/>
-                                          <relation type="labelled-by" target="battery-custom-notification-spin-label"/>
-                                        </accessibility>
                                       </object>
                                       <packing>
                                         <property name="left_attach">1</property>
--- a/data/ui/preferences-device-panel.ui
+++ b/data/ui/preferences-device-panel.ui
@@ -603,7 +603,7 @@
                     <child>
                       <object class="GtkFrame" id="battery-device">
                         <property name="visible">True</property>
-                        <property name="can_focus">False</property>
+                        <property name="can_focus">True</property>
                         <property name="margin_bottom">32</property>
                         <property name="label_xalign">0</property>
                         <child>
@@ -677,12 +677,12 @@
                                 <property name="height_request">56</property>
                                 <property name="visible">True</property>
                                 <property name="can_focus">True</property>
-                                <property name="action_name">settings.custom-battery-notification</property>
-                                <property name="selectable">False</property>
+                                <property name="activatable">True</property>
+                                <property name="action_name">settings.custom-battery-notification</property>                                
                                 <child>
                                   <object class="GtkGrid">
                                     <property name="visible">True</property>
-                                    <property name="can_focus">False</property>
+                                    <property name="can_focus">True</property>
                                     <property name="margin_start">12</property>
                                     <property name="margin_end">12</property>
                                     <property name="hexpand">True</property>
@@ -691,7 +691,7 @@
                                     <child>
                                       <object class="GtkLabel" id="battery-custom-notification-label">
                                         <property name="visible">True</property>
-                                        <property name="can_focus">False</property>
+                                        <property name="can_focus">True</property>
                                         <property name="halign">start</property>
                                         <property name="valign">center</property>
                                         <property name="hexpand">True</property>
@@ -728,7 +728,7 @@
                                       <object class="GtkSpinButton" id="battery-custom-notification-value">
                                         <property name="visible">True</property>
                                         <property name="value">80</property>
-                                        <property name="can_focus">True</property>
+                                        <property name="can_focus">False</property>
                                         <property name="max_length">2</property>
                                         <property name="activates_default">True</property>
                                         <property name="width_chars">2</property>
--- a/data/ui/preferences-device-panel.ui
+++ b/data/ui/preferences-device-panel.ui
@@ -564,6 +564,12 @@
                 <property name="can_focus">False</property>
                 <property name="shadow_type">none</property>
                 <child>
+                  <object class="GtkAdjustment" id="custom-battery-adjustment">
+                    <property name="upper">99</property>
+                    <property name="lower">01</property>
+                    <property name="step_increment">1</property>
+                    <property name="page_increment">5</property>
+                  </object>
                   <object class="GtkBox">
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
@@ -666,11 +672,11 @@
                               </object>
                             </child>
                             <child>
-                              <object class="GtkListBoxRow" id="battery-sixty-notification-row">
+                              <object class="GtkListBoxRow" id="battery-custom-notification-row">
                                 <property name="height_request">56</property>
                                 <property name="visible">True</property>
                                 <property name="can_focus">True</property>
-                                <property name="action_name">settings.sixty-battery-notification</property>
+                                <property name="action_name">settings.custom-battery-notification</property>
                                 <property name="selectable">False</property>
                                 <child>
                                   <object class="GtkGrid">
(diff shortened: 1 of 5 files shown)


daniellandau:
Thanks for the contribution!

From what I've understood currently Glade is considered harmful and editing UI files by hand is the way. I understand you're looking into an example of a SpinButton in a UI file, and I'm sorry I don't have one right now to give, but maybe this code example helps: https://github.com/daniellandau/switcher/blob/master/prefs.js#L185-L195

Corresponding example of an integer valued setting key in a schema: https://github.com/daniellandau/switcher/blob/master/schemas/org.gnome.shell.extensions.switcher.gschema.xml#L40-L43

I'll try to review the code in detail soon when I have time. In the meantime could you check the failing CI test?

amija004:
Thank you for the examples, I'll check them out once I figure out the failed test.

src/service/plugins/battery.js
```diff
@@ -257,6 +257,64 @@ var Plugin = GObject.registerClass({
...
+        if (this.device.get_action_enabled('ring')) {
+            buttons = [{
+                label: _('Ring'),
+                action: 'ring',
+                parameter: null,
+            }];
+        }
+
+        this.device.showNotification({
+            id: 'battery|sixty',
```
andyholmes, line 279: I think it would be a good idea to drop the `|level` ID suffix on these notifications, or change them all to `battery-level`. There should really only be one of these showing anyways and having multiple IDs complicates the logic of showing/hiding the proper one.

src/service/plugins/battery.js
```diff
@@ -257,6 +257,64 @@ var Plugin = GObject.registerClass({
         });
     }
 
+    /**
+     * Notify the user the remote battery is at 60% charge.
+     */
+    _sixtyBatteryNotification() {
+        if (!this.settings.get_boolean('sixty-battery-notification'))
```
andyholmes, line 264: Your tests are failing because this setting is `false` by default. You can fix the tests by toggling the setting in the unit test before sending the packet.

src/service/plugins/battery.js
amija004: I am trying to implement a spinbutton to replace both of these switches, which will drop this down to a single notification.

src/service/plugins/battery.js
amija004: Thank you. I have never worked with CI tests before, or with a public git repo for that matter. I will try to enable them in my fork to keep the upstream repo as clean as possible. I would greatly appreciate any corrections or tips on how to keep the clutter down while testing.

I am working on creating a switch to enable a custom level alert, and then a spinbutton to determine what that level is. I was going to leave this off by default, as I imagine most people fully charge their phones all the time.

andyholmes: You should be able to run the tests locally with `meson test -C _build` (see [meson docs](https://mesonbuild.com/Unit-tests.html#testing-tool)). Probably we should add directions for running tests in the `CONTRIBUTING.md` page...

data/ui/preferences-device-panel.ui
```diff
@@ -564,6 +564,12 @@
                 <property name="can_focus">False</property>
                 <property name="shadow_type">none</property>
                 <child>
+                  <object class="GtkAdjustment" id="custom-battery-adjustment">
+                    <property name="upper">99</property>
+                    <property name="lower">01</property>
+                    <property name="step_increment">1</property>
+                    <property name="page_increment">5</property>
+                  </object>
```
andyholmes, line 572: I think you can just set this directly actually, if that's easier:

```xml
<property name="adjustment">
  <object class="GtkAdjustment">
    <property name="upper">99</property>
    <property name="lower">01</property>
    <property name="step_increment">1</property>
    <property name="page_increment">5</property>
  </object>
</property>
```

I'm not sure if GtkBuilder allows you to place objects arbitrarily (that `<child>` tag is for the `GtkViewport`).

amija004: I copied that placement from another repo I found at some point, they had the adjustment separate so I did the same. Your way is much simpler and less confusing.  I have implemented it locally and will push it once I have working code.

src/service/plugins/battery.js
amija004: I could not get local testing to fully work, ESLint kept telling me that the path needed to be a relative path instead of an absolute one.

The other tests ran without a problem though, which was a great help!

data/ui/preferences-device-panel.ui
amija004: I did not change the adjustment location as it kept giving me errors when I tried to relocate it, telling me that the adjustment wasn't allowed inside the GtkGrid. I didn't understand the error or how to fix it, so I left it as is. Its current location outside the template does not throw any errors.

amija004:
I was able to implement the spin button successfully.

I was encountering an issue with the menu where the only way to deselect the spinbutton was by selecting another button or leaving the menu. I am still unsure how to fix that, but I will look into it.

data/ui/preferences-device-panel.ui
andyholmes: Maybe I wasn't clear with my note above, but the `<property name="adjustment">` was to be set on the spin button, such that:

```xml
<object class="GtkSpinButton">
  <property name="adjustment">
    <object class="GtkAdjustment">
      <property name="upper">99</property>
      <property name="lower">01</property>
      <property name="step_increment">1</property>
      <property name="page_increment">5</property>
    </object>
  </property>
<!-- other stuff -->
</object>
```
If you think of GtkBuilder as just objects with properties (whose value can be objects), and child objects (in `<child>` tags) it starts to become obvious how it all works :)

Outside in the template is fine also, though. I frequently put GtkDialogs in the top-level `<interface>` as well :)

andyholmes:
It may be that the GtkListBox row is stealing the focus. Most of those rows are set up so that if you can click the row to toggle the switch. Since you have another widget in there, you may want to just set `<property name="activatable">False</property>` on the GtkListBoxRow and see if that helps.

amija004:
I tried setting activatable to false, but it made the button not act like the others, where if you click the label it toggles the effect. I decided to remove the focus from the spinbutton entirely, and rely solely on the mouse to change the value. I think its less confusing that way, as the continued focus generally leads me to believe the changes haven't been locked in.

I noticed that most other pull requests have far fewer commits than mine. Should I change anything for my pull requests moving forward, or is it ok as is?

andyholmes:
> I noticed that most other pull requests have far fewer commits than mine. Should I change anything for my pull requests moving forward, or is it ok as is?

It usually doesn't matter to me since I typically just squash & merge, but @daniellandau will have final say on MRs now.

amija004:
I found a bug while using my changes that I'm not sure how to fix. You can set the charge level, and it will function as intended, until you close the settings menu and reenter the battery settings menu, at which point the charge level will reset to default value. I believe it has to do with the UI files s setting a value when called. I do not know how to have the UI file pull the current value from the schema key. I tried removing the default from the schema key and UI, but the value just reset to 0 instead of 80 upon reentry. I also tried to reference the schema value by calling "settings.custom-battery-notification" but this just gave me an error.

https://github.com/amija004/gnome-shell-extension-gsconnect/blob/master/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml#L88-L91

    <key name="custom-battery-notification-value" type="d">
      <range min="1" max="99"></range>
      <default>80</default>
    </key>

https://github.com/amija004/gnome-shell-extension-gsconnect/blob/master/data/ui/preferences-device-panel.ui#L727-L746

                                    <child>
                                      <object class="GtkSpinButton" id="battery-custom-notification-value">
                                        <property name="visible">True</property>
                                        <property name="value">80</property>
                                        <property name="can_focus">False</property>
                                        <property name="max_length">2</property>
                                        <property name="activates_default">True</property>
                                        <property name="width_chars">2</property>
                                        <property name="adjustment">battery-custom-adjustment</property>
                                        <signal name="output" handler="_setCustomChargeLevel" swapped="no"></signal>
                                        <accessibility>
                                          <relation type="controlled-by" target="battery-custom-notification-row"/>
                                          <relation type="labelled-by" target="battery-custom-notification-label"/>
                                        </accessibility>
                                      </object>
                                      <packing>
                                        <property name="left_attach">1</property>
                                        <property name="top_attach">0</property>
                                      </packing>
                                    </child>

src/preferences/device.js
```diff
@@ -611,6 +613,15 @@ var Panel = GObject.registerClass({
         }
     }
 
+    _setCustomChargeLevel(spin) {
+        const settings = this.pluginSettings('battery');
+        const oldLevel = settings.get_value('custom-battery-notification-value').unpack();
+        const newLevel = GLib.Variant.new('d', spin.get_value());
+
+        if (newLevel !== oldLevel)
```
andyholmes, line 621: Here you're trying to compare two `GVariant` objects as though they're literals. If you want to check the equality of variants you should do use [`GLib.Variant.equal()`](https://gjs-docs.gnome.org/glib20/glib.variant#method-equal) like `if (!newLevel.equal(oldLevel)`.

There's not much point in creating intermediate GVariant objects in JavaScript though. You should probably just do:

```js
let oldLevel = settings.get_double('custom-battery-notification-value');
let newLevel = spin.get_value();

if (newLevel !== oldLevel)
    settings.set_double('custom-battery-notification-value', newValue);
```

You should also have a second method like `_getCustomeChargeLevel()` that is called when the widget is constructed. I think the simpler way is just to use a property binding, but that's up to you.

daniellandau:
Thanks, I'll merge this now and open another pull request with fixes for the remaining issues.

PR #1039, opened 2021-01-03
Show transfered file in folder as default action of notification (single click)

### LINKED ISSUE #1009: Click on "Transfer successful" notification should open file or it's location with file selected

HaMF:
This adresses issue #1009, but probably needs some work. See issue discussion for details

closes #1009

### DIFF
--- a/src/service/device.js
+++ b/src/service/device.js
@@ -743,7 +743,6 @@ var Device = GObject.registerClass({
     openPathSelectFile(action, parameter) {
         const path = parameter.unpack();
         const uri = path.includes('://') ? path : `file://${path}`;
-        debug(uri, null)
 
         Gio.DBus.session.call(
             'org.freedesktop.FileManager1',
--- a/src/service/device.js
+++ b/src/service/device.js
@@ -483,6 +483,13 @@ var Device = GObject.registerClass({
         openPath.connect('activate', this.openPath);
         this.add_action(openPath);
 
+        const openPathSelectFile = new Gio.SimpleAction({
+            name: 'openPathSelectFile',
+            parameter_type: new GLib.VariantType('s'),
+        });
+        openPathSelectFile.connect('activate', this.openPathSelectFile);
+        this.add_action(openPathSelectFile);
+
         // Preference helpers
         const clearCache = new Gio.SimpleAction({
             name: 'clearCache',
@@ -731,6 +738,25 @@ var Device = GObject.registerClass({
         // Normalize paths to URIs, assuming local file
         const uri = path.includes('://') ? path : `file://${path}`;
         Gio.AppInfo.launch_default_for_uri_async(uri, null, null, null);
+    }
+
+    openPathSelectFile(action, parameter) {
+        const path = parameter.unpack();
+        const uri = path.includes('://') ? path : `file://${path}`;
+        debug(uri, null)
+
+        Gio.DBus.session.call(
+            'org.freedesktop.FileManager1',
+            '/org/freedesktop/FileManager1',
+            'org.freedesktop.FileManager1',
+            'ShowItems',
+            new GLib.Variant('(ass)', [[uri], 's']),
+            null,
+            Gio.DBusCallFlags.NONE,
+            -1,
+            null,
+            null
+        );
     }
 
     _clearCache(action, parameter) {
--- a/src/service/plugins/share.js
+++ b/src/service/plugins/share.js
@@ -177,8 +177,8 @@ var Plugin = GObject.registerClass({
                     this.device.name
                 );
                 action = {
-                    name: 'openPath',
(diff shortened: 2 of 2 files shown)


src/service/device.js
```diff
@@ -733,6 +740,24 @@ var Device = GObject.registerClass({
...
+            '/org/freedesktop/FileManager1',
+            'org.freedesktop.FileManager1',
+            'ShowItems',
+            new GLib.Variant('(ass)', [[uri], 's']),
+            null,
+            Gio.DBusCallFlags.NONE,
+            -1,
+            null,
+            null
+        );
```
sonnyp, line 758: looks like there should be an error to catch here

https://www.andyholmes.ca/articles/dbus-in-gjs.html#direct-calls

Logging the error is enough.

sonnyp (changes requested):
Thanks! - and sorry for the late review

> if we want to use this behavior when "Open Folder is clicked

Yes. This is better.

For the label, I think GNOME already has a wording for this action but I couldn't find it. Otherwise I'd suggest "Show in folder". 

Given this is a GNOME Shell extension, I don't think we need the previous method as fallback.

> if dbus is really the only way to achieve this (that's what I found, should work for nautilus and dolphin)

Yep, there are no wrapper exposed in GNOME libraries AFAIK

https://github.com/search?q=org.freedesktop.FileManager1&type=code

> how tests can be added

I wouldn't say it's necessary here. But if you like checkout `installed-tests/suites/components/`. I guess you could listen on the bus and wait for what's of interested.

@andyholmes wdyt? any better idea for testing this?

> how I can check the transfer-failed notification works

Perhaps send a large file and disconnect network midway. Why is it relevant?

src/service/device.js
```diff
@@ -733,6 +740,24 @@ var Device = GObject.registerClass({
         Gio.AppInfo.launch_default_for_uri_async(uri, null, null, null);
     }
 
+    openPathSelectFile(action, parameter) {
```
sonnyp, line 743: consider renaming to `showPathInFolder`

andyholmes:
> I wouldn't say it's necessary here. But if you like checkout `installed-tests/suites/components/`. I guess you could listen on the bus and wait for what's of interested.
> 
> @andyholmes wdyt? any better idea for testing this?

I don't think there are many UI tests in the suite yet, those are kind of hard to do in a CI at the moment. Probably this one would just have to be tested locally.

ferdnyc:
Jumping in way late on this, but I may pick it up if @HaMF doesn't plan to continue.

@HaMF 
> > if we want to use this behavior when "Open Folder is clicked
@sonnyp 
> Yes. This is better.
> 
> For the label, I think GNOME already has a wording for this action but I couldn't find it. Otherwise I'd suggest "Show in folder".

The Nautilus wording (seen in both the Recent list and Search results is "Open Item Location". Which I suppose may not technically be a folder, so I have to give them that. (It may be at the root of a removable drive, for example.)

However, when _outside_ the file manager itself, the wording is usually different. Common are:

* "Show in folder" (Chrome & Firefox download lists)
* "Open Containing Folder" (Evince, GNOME Builder)
* No text since there's just a 📁 icon with no tooltip, JFC. (GNOME Web)
* No idea, it has never found a single photo on my system (which contains plenty of photos). (GNOME Photos)

HaMF:
@ferdnyc Thanks for the research regarding the wording. Feel free to take it and finish it, I won't get around to it in the next days. If you haven't done so until next week when I'm back I'll look into it.

@sonnyp thanks for the review and the valuable pointers. Sorry must have missed the notification. I'll check in detail if ferdnyc does not pick it up

src/service/device.js
```diff
@@ -733,6 +740,24 @@ var Device = GObject.registerClass({
...
+        const path = parameter.unpack();
+        const uri = path.includes('://') ? path : `file://${path}`;
```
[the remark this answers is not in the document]
ferdnyc: @sonnyp Looking at this again, I wonder if that's really necessary. By the time it gets here, the uri has _already_ been through `Gio.File`:
https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/4208acc233bb2757b8c84c19e74715413961b06c/src/service/plugins/share.js#L179-L189

The URI coercion here is (as I read it) just a safety valve, _in case_ someone adds a call to `openPath()` (or `openPathSelectFile()`) with a bare pathname. But in normal/expected usage, the parameter is the result of a `file.get_uri()` already. Round-tripping through `Gio.File` a second time may be overkill?

ferdnyc:
I've submitted #1491 containing both @HaMF's commits from this PR (thanks, @HaMF!), as well as my further changes to address the review input here.

This can probably be closed in favor of #1491. I'll wait a bit, tho, so others have time to look over that PR and decide whether it's the right path.

PR #1047, opened 2021-01-30
Added tooltips to plugin menu

mr-kenhoff:
I recently installed GSConnect and was not quite sure what all the plugins do. 
Therefore I added tooltips that briefly describe what they do.

We can talk about the formulation of the description texts ;)

### DIFF
--- a/data/ui/preferences-section-row.ui
+++ b/data/ui/preferences-section-row.ui
@@ -1,57 +1,57 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.38.2 -->
 <interface domain="org.gnome.Shell.Extensions.GSConnect">
   <requires lib="gtk+" version="3.24"/>
   <template class="GSConnectPreferencesSectionRow" parent="GtkListBoxRow">
-    <property name="height_request">56</property>
-    <property name="selectable">False</property>
+    <property name="height-request">56</property>
     <property name="visible">True</property>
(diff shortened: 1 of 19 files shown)


mr-kenhoff:
Example:

Screenshot from 2021-01-30 17-06-05
 (Unfortunately my cursor is hidden during screenshots)

sonnyp:
Thanks!

I agree - plugins should be more descriptive. 

While tooltips is definitely an improvement - I don't think the description should be difficult to discover.
Tooltips also don't work well on touch devices.

Would you mind turning those tooltips into `GtkLabel` with the `.subtitle` class like in GNOME Settings:

Screenshot from 2021-03-28 13-40-55
Screenshot from 2021-03-28 13-39-48

sonnyp:
Let me know if you need assistance - happy to provide some guidance.

sonnyp:
Looking at the code looks like passing subtitle to `new SectionRow` parameters would do the trick.

mr-kenhoff:
Thanks for Your feedback. I will change the descriptions to subtitles.

mr-kenhoff:
Done... Would be happy for a review.

sonnyp:
many thanks - I made some changes - see my commits

sonnyp:
For future reference, I think we should replace this list of plugins with something else.

Even with the descriptions - it is still not clear what each plugin does or why should I enable or disable one.

PR #1077, opened 2021-03-27
Korean Translation

UtsushimiNeneka:
Can you add Korean translation to this extension?

I made ko.po file.


andyholmes:
ok I imported the file in crowdin. You'll have to pick whether to use Crowdin or pull-request though, because Crowdin will overwrite what's in the repository.

Issue #1073, opened 2021-03-23
Extension doesn't work after updating gjs to 1.68.0


Noobsai:
After updating gjs to 1.68.0 the extension doesn't work. Device list is disappeared (stuck in a searching) and cannot connect to phone.
Downgrade back to 1.66.2 is resolved the problem.
Event log:
```
JS ERROR: TypeError: this._instance is undefined
_init@/gsconnect@andyholmes.github.io/service/utils/dbus.js:69:20
_exportDevice@/gsconnect@andyholmes.github.io/service/manager.js:295:26
_loadDevices@/gsconnect@andyholmes.github.io/service/manager.js:266:18
start@/gsconnect@andyholmes.github.io/service/manager.js:453:14
vfunc_startup@/gsconnect@andyholmes.github.io/service/daemon.js:258:22
vfunc_handle_local_options@/gsconnect@andyholmes.github.io/service/daemon.js:664:18
@/gsconnect@andyholmes.github.io/service/daemon.js:727:17
```
Maybe this gjs PR broke something: https://gitlab.gnome.org/GNOME/gjs/-/merge_requests/591

KonstantMan:
I think every one who uses GNOME with up to date distros like Arch has the exact same issue , kde connect works just fine so it's not a firewall problem

daniellandau:
I'll try to get that fixed this weekend.

daniellandau:
Indeed that PR seems to be related. If I remove the CONSTRUCT_ONLY flag on line 46 of dbus.js it starts working again.

@andyholmes did you have a specific reason to add that flag?

Probably anyways the correct fix would be to figure out the naming clash or whatever in detail, but I think the priority should be to get a fix out, so if there's no pressing reason to have that parameter be CONSTRUCT_ONLY I think that's the fix I'll go with.

andyholmes:
> Probably anyways the correct fix would be to figure out the naming clash or whatever in detail, but I think the priority should be to get a fix out, so if there's no pressing reason to have that parameter be CONSTRUCT_ONLY I think that's the fix I'll go with.

Well it can only be set at construction, but we're not exactly publishing a library here so probably doesn't matter :)

It might be because I named the variable `_instance` while the property name is `g-instance`, thus maybe the variable should have been named `_g_instance`?

Issue #1071, opened 2021-03-20
Does not work for Fedora 34: missing ciphers.


andyholmes:
Should be noted that this does not work for Fedora 34 on a fresh installation due to an issue with missing ciphers.

_Originally posted by @xeu100 in https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1069#issuecomment-802800881_

andyholmes:
This is probably an issue with `gnutls >= 3.7`, although it's unclear if it only affects Fedora. Error looks something like:

```
JS ERROR: Gio.TlsError: Error performing TLS handshake: No supported cipher suites have been found.
_handshake/</<@/var/home/andrew/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:595:44
@./daemon.js:728:17
```

jnsh:
> This is probably an issue with `gnutls >= 3.7`, although it's unclear if it only affects Fedora.

FWIW, I'm not experiencing this issue on Arch, with gnutls 3.7.1. Looks like F34 has gnutls 3.7.0 out the box, and 3.7.1 package is in testing ([source](https://src.fedoraproject.org/rpms/gnutls)). May be worth checking if update to 3.7.1 fixes this, or there could be some differences in how the package is build between the two distributions.

andyholmes:
It could just as easily be an error caused by an older Android device, but I really don't know at this point :) GSConnect isn't really a priority for me anymore, I just happened to be working with gnutls when I got the e-mail :)

fareszr:
> > This is probably an issue with `gnutls >= 3.7`, although it's unclear if it only affects Fedora.
> 
> FWIW, I'm not experiencing this issue on Arch, with gnutls 3.7.1. Looks like F34 has gnutls 3.7.0 out the box, and 3.7.1 package is in testing ([source](https://src.fedoraproject.org/rpms/gnutls)). May be worth checking if update to 3.7.1 fixes this, or there could be some differences in how the package is build between the two distributions.

after updating gnutls its still doesn't work
devices don't show up 
running on latest fedora 34 beta.

on a modern android phone

andyholmes:
Seems likely that Fedora changed its system priority file to exclude the ciphers supported by the Android app, since this is only seems to be affecting Fedora 34 users.

fareszr:
[...]

PR #1082, opened 2021-03-28
Fix compatibility with GJS 1.68

daniellandau:
Fixes: https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1073
Fixes: https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1071

Related to this change in GJS: https://gitlab.gnome.org/GNOME/gjs/-/merge_requests/591/

Parameters with CONSTRUCT_ONLY flag set seem to work differently now, or maybe
are just broken :shrug:

### DIFF
--- a/installed-tests/fixtures/backend.js
+++ b/installed-tests/fixtures/backend.js
@@ -27,7 +27,7 @@ var ChannelService = GObject.registerClass({
             'port',
             'Port',
             'The port used by the service',
-            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
+            GObject.ParamFlags.READWRITE,
             0,  GLib.MAXUINT16,
             DEFAULT_PORT
         ),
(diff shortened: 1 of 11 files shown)


sonnyp (commented):
I can confirm this fixes it for Fedora 34 / gjs 1.68 :+1: 

There are more classes with `GObject.ParamFlags.CONSTRUCT_ONLY` in the code base aren't they affected by the issue?

GitGangGuy:
Can we get this merged? I'm on Arch and stuck with a broken GSConnect :)

daniellandau:
Current master doesn't work fully with Gnome 3.38 so some work to figure out what to do with that, and anyways the review queue on extensions.gnome.org isn't instant so it'll be a couple days before the fix arrives through there.

I'd like to first also find out whether CONSTRUCT_ONLY should be removed from all the other places too.

GitGangGuy:
Great! Can't wait for the update to be rolled out.

PR #1084, opened 2021-03-28
Fix error in log when getting a notification with an empty body

daniellandau:
JS ERROR: Error: Expected type string for argument 'body' but got type undefined


sonnyp (commented):
LGTM 

I'd suggest to stop using `Object.assign` in favor of direct assignment here though. It would remove the need for this and make the code simpler.

PR #1088, opened 2021-03-29
mpris: Initialize conditional fields for requestNowPlaying response

luk1337:
This fixes an issue where previous song metadata is not fully
zeroed-out after song changed.

Test: Play song with cover art then switch to one without it,
        observe that on kde-connect-android you still see previous
        song's cover art.


luk1337:
bump @daniellandau @sonnyp

daniellandau:
Thanks!

PR #1100, opened 2021-04-14
Charge level notification fixes

daniellandau:


data/ui/preferences-device-panel.ui
```diff
@@ -696,7 +696,7 @@
                                         <property name="valign">center</property>
                                         <property name="hexpand">True</property>
                                         <property name="vexpand">True</property>
-                                        <property name="label" translatable="yes">Custom Charge Level Notification</property>
+                                        <property name="label" translatable="yes">Charged Up to Custom Level Notification</property>
```
ferdnyc, line 699: You're gonna hate me, but... that's even harder to parse, IMHO.

I'd maybe go with something like:
> Notify when charged to:

and then add a % sign to/after the spinbox.


I'd probably also change the others to "Notify when charging complete", "Notify when battery low". Both for consistency, and to be more imperative. A "Low Battery Notification" is the thing that happens _when_ the battery is low (if enabled, by switching on "Notify when...")

In _faaact_, rather than the separate "full" notification and "custom level" options, why not just have:
> Notify when charged to:  [default 100%]
> Notify when battery below: [default whatever... 20% I think?]

Are there _really_ people who use the custom level notification, and **also** need fully-charged notifications? And why would only the upwards-charge level be customizable?

PR #1134, opened 2021-06-22
Expose LoopStatus and Shuffle control to clients

luk1337:
This change follows KDE Connect change merged here:
https://invent.kde.org/network/kdeconnect-kde/-/merge_requests/401

### DIFF
--- a/src/service/plugins/mpris.js
+++ b/src/service/plugins/mpris.js
@@ -228,6 +228,12 @@ var Plugin = GObject.registerClass({
             }
 
             // Player Properties
+            if (packet.body.hasOwnProperty('setLoopStatus'))
+                player.LoopStatus = packet.body.setLoopStatus;
+
+            if (packet.body.hasOwnProperty('setShuffle'))
+                player.Shuffle = packet.body.setShuffle;
+
             if (packet.body.hasOwnProperty('setVolume'))
                 player.Volume = packet.body.setVolume / 100;
 
@@ -260,6 +266,8 @@ var Plugin = GObject.registerClass({
                     canGoNext: player.CanGoNext,
                     canGoPrevious: player.CanGoPrevious,
                     canSeek: player.CanSeek,
+                    loopStatus: player.LoopStatus,
+                    shuffle: player.Shuffle,
 
                     // default values for members that will be filled conditionally
                     albumArtUrl: '',


src/service/components/mpris.js
```diff
@@ -325,7 +325,7 @@ var Player = GObject.registerClass({
 
     get LoopStatus() {
         if (this._LoopStatus === undefined)
-            this._LoopStatus = 'None';
+            this._LoopStatus = null;
```
andyholmes, line 328: `null` is an invalid type for this property; what's the purpose of doing this?

src/service/components/mpris.js
```diff
@@ -396,7 +396,7 @@ var Player = GObject.registerClass({
 
     get Shuffle() {
         if (this._Shuffle === undefined)
-            this._Shuffle = false;
+            this._Shuffle = null;
```
andyholmes, line 399: Same as above.

src/service/components/mpris.js
```diff
@@ -682,7 +682,7 @@ const PlayerProxy = GObject.registerClass({
     }
 
     get LoopStatus() {
-        return this._get(this._player, 'LoopStatus', 'None');
+        return this._get(this._player, 'LoopStatus');
```
andyholmes, line 685: See above.

src/service/components/mpris.js
```diff
@@ -743,7 +743,7 @@ const PlayerProxy = GObject.registerClass({
     }
 
     get Shuffle() {
-        return this._get(this._player, 'Shuffle', false);
+        return this._get(this._player, 'Shuffle');
```
andyholmes, line 746: Ditto

src/service/plugins/mpris.js
```diff
@@ -228,6 +228,12 @@ var Plugin = GObject.registerClass({
             }
 
             // Player Properties
+            if (packet.body.hasOwnProperty('setLoopStatus'))
+                player.LoopStatus = packet.body.setLoopStatus;
+
+            if (packet.body.hasOwnProperty('setShuffle'))
+                player.Shuffle = packet.body.setShuffle === 'true';
```
andyholmes, line 235: I'm not sure I see the reason for coercion and comparison here. I think kdeconnect-android should just send a real boolean and that should be passed to the proxy.

andyholmes (changes requested):
Thanks for the contribution :)

I don't see a problem here as long as the type issues are fixed. Bonus points if you can add some unit tests.

src/service/components/mpris.js
luk1337: These fields are optional thus I need some way to know whether given player sets them or not. See: https://specifications.freedesktop.org/mpris-spec/2.2/Player_Interface.html

If you wish for me to use different fallback value than `null`, let me know.

src/service/plugins/mpris.js
luk1337: https://invent.kde.org/network/kdeconnect-android/-/merge_requests/255, will force push this away when that gets merged. Thanks!

src/service/components/mpris.js
andyholmes: I think the fallbacks of `LoopStatus`/`None` and `Shuffle`/`false` would best.

I reason that if the MPRIS player doesn't support the properties the `try..catch` will return the fallback silently, and if the remote doesn't then it should just ignore those fields. If one does and the other doesn't, I think a player that doesn't support `LoopStatus` is effectively `None` and same for `Shuffle` being `false`.

Does that seem reasonable?

luk1337: Uh but then how would we know if it's good to add these values to requestNowPlaying response so that KDE Connect app can hide not supported controls?

andyholmes: Ah I see why you did that now. In that case you can re-use the `_get()` function and leave out the fallback as you did below in the `PlayerRemote` class, but the property getters here in the `Player` class can not return `null`.

The problem is the `Player` class is the base class for both local players we consume and remote players we export. If we return `null` as an exported player our code will throw when we try to pack the `null` into a DBus variant, while the other side of DBus will probably get an incorrect error code (or no reply at all).

Unfortunately, when we're exporting players we sort of have to pretend to support everything, because we can't really dynamically export different parts of the MPRIS interface (not easily anyways).

andyholmes: Sidenote: If you do want to go through to work of supporting optional properties on MPRIS proxies we export, you can open another PR and I'll try and help you through that.

luk1337: Ok, since kde-connect-android MR just got merged I fpushed this PR according to your suggestions.

daniellandau:
👍

PR #1135, opened 2021-06-27
Update nautilus-gsconnect.py

stuaxo:
When I was restarting nautilus I noticed this message:

```
GSConnect: No translation file found for domain
Initializing nautilus-dropbox 2019.02.14
Traceback (most recent call last):
  File "/home/stu/.local/share/nautilus-python/extensions/nautilus-gsconnect.py", line 129, in _get_managed_objects
    self.devices[object_path] = (props['Name'],
KeyError: 'Name'
```

This fixes the error as the dict was empty in this case, but not idea if something else breaks.

### DIFF
--- a/nautilus-extension/nautilus-gsconnect.py
+++ b/nautilus-extension/nautilus-gsconnect.py
@@ -125,6 +125,8 @@ class GSConnectShareExtension(GObject.Object, FileManager.MenuProvider):
 
         for object_path, props in objects.items():
             props = props['org.gnome.Shell.Extensions.GSConnect.Device']
+            if not props:
+                continue
 
             self.devices[object_path] = (props['Name'],
                                          Gio.DBusActionGroup.get(


sonnyp (commented):
Any idea in which case this happens or how to reproduce?

stuaxo:
I might have to guess a bit here:

The thing I do: restart nautilus.  `nautilus -q`

If that isn't enough, then it may be that the DLNA device has disappeared, so if restarting nautilus isn't enough, try something like restarting pulseaudio-dlna, or turning off some dlna device that was previously present and then restarting nautilus.


There are a few places with the same code fragment, so it might be better to have a utility function, since I patched the one place the crash was happening for me.

daniellandau:
This seems like a corner case and the fix doesn't seem like it should do any harm so I'm just going to merge without trying to get to the very bottom of this.

Issue #1139, opened 2021-07-11
GSConnect doesn't work with Microsoft Edge Dev (v93)


AnishDe12020:
> ***ATTENTION**: GSConnect only supports the latest, stable version of GNOME. We are no longer accepting bug reports for previous versions.*

**Describe the bug**

GSConnect works fine with Gnome but the browser extension doesn't work. The extensions has got a red indicator and clicking on it shows "Service unavailable". I am using Microsoft Edge Dev and it is not working for a few days (it never worked).

**Expected behavior**

Well, I expect the browser extension to work and I should be able to share webpages to my phone via right clicking on a page.

**Screenshots**

![image](https://user-images.githubusercontent.com/63192115/125186032-9d067900-e245-11eb-9907-40f23c72fb2b.png)

**Support Log**

Please generate a support log ([Instructions](../wiki/Help#generate-support-log)) and paste any messages related to this issue between the two ``` lines below.

This is the background.html for the extension:
![image](https://user-images.githubusercontent.com/63192115/125186094-fb335c00-e245-11eb-80e5-12e77c04a8f8.png)

This is support log generated from the gnome extension:

```

GSConnect: 45 (user)
GJS:       16702
Session:   x11
OS:        Pop!_OS 21.04
--------------------------------------------------------------------------------
-- Journal begins at Tue 2021-07-06 15:05:42 IST, ends at Sun 2021-07-11 12:46:43 IST. --
Jul 11 12:46:03 systemd[23433]: Started Application launched by gsd-media-keys.
Jul 11 12:46:03 systemd[23433]: app-gnome-flameshot-34467.scope: Succeeded.
Jul 11 12:46:03 gnome-shell[24223]: Window manager warning: Invalid WM_TRANSIENT_FOR window 0x680000a specified for 0x680001a.
Jul 11 12:46:03 org.flameshot.Flameshot[34184]: QPainter::begin: Paint device returned engine == 0, type: 2
Jul 11 12:46:03 org.flameshot.Flameshot[34184]: QPainter::setRenderHint: Painter must be active to set rendering hints
Jul 11 12:46:03 org.flameshot.Flameshot[34184]: QPainter::setCompositionMode: Painter not active
Jul 11 12:46:03 org.flameshot.Flameshot[34184]: QPainter::translate: Painter not active
Jul 11 12:46:03 org.flameshot.Flameshot[34184]: QPainter::setPen: Painter not active
[...]

PR #1141, opened 2021-07-13
Added support for the Browser Extension to work with Microsoft Edge

AnishDe12020:
I added support for the Browser Extension to work with the Microsoft Edge Beta and Microsoft Edge Dev (the two versions available for Linux as of now). This was simply done by adding the path to the Native Messaging Host directories to the [`utils.js`](https://github.com/AnishDe12020/gnome-shell-extension-gsconnect/blob/f2eb98992090316797901477c7895ec3c6ba482e/src/shell/utils.js) [`manifests`](https://github.com/AnishDe12020/gnome-shell-extension-gsconnect/blob/f2eb98992090316797901477c7895ec3c6ba482e/src/shell/utils.js#L160) array. The extension now works as it should.

This should fix the issue about which i had opened an issue #1139

### DIFF
--- a/src/shell/utils.js
+++ b/src/shell/utils.js
@@ -165,7 +165,7 @@ function installService() {
         [`${confDir}/BraveSoftware/Brave-Browser/NativeMessagingHosts/`, google],
         [`${homeDir}/.mozilla/native-messaging-hosts/`, mozilla],
         [`${homeDir}/.config/microsoft-edge-dev/NativeMessagingHosts`, google],
-        [`${homeDir}/.config/microsoft-edge-beta/NativeMessagingHosts`, google]
+        [`${homeDir}/.config/microsoft-edge-beta/NativeMessagingHosts`, google],
     ];
 
     // If running as a user extension, ensure the DBus service, desktop entry,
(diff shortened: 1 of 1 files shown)


daniellandau:
Great, thanks! Could you fix the linter issues?

AnishDe12020:
Yep, fixed them, sorry about committing twice. Can you run the checks?

daniellandau:
That's ok, we do squash merges. Thank you again for thr contribution!

sonnyp:
Should we submit the extension to https://microsoftedge.microsoft.com/addons/Microsoft-Edge-Extensions-Home ?

Weird time we're living in :)

AnishDe12020:
> Should we submit the extension to https://microsoftedge.microsoft.com/addons/Microsoft-Edge-Extensions-Home ?
> 
> Strange times :)

Well given that Microsoft Edge is supported now, why not considering edge is becoming a popular browser :eyes: . There are some fees associated and an application process though.

sonnyp:
> There are some fees associated and an application process though.

That appears to be incorrect.

https://docs.microsoft.com/en-us/microsoft-edge/extensions-chromium/publish/create-dev-account

> There is no registration fee involved in submitting extensions to the Microsoft Edge program.

AnishDe12020:
Oh sorry about that, there are fees associated for the chrome web store. Got confused

PR #1144, opened 2021-07-22
Update Spanish translation

oscfdezdz:


po/es.po
```diff
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: gsconnect\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-09-10 22:31-0700\n"
-"PO-Revision-Date: 2020-09-16 18:44\n"
+"POT-Creation-Date: 2021-07-22 08:07+0200\n"
```
sonnyp, line 5:
```suggestion
"POT-Creation-Date: 2020-09-10 22:31-0700\n"
```

This field shouldn't be updated according to https://www.gnu.org/software/trans-coord/manual/gnun/html_node/PO-Header.html

sonnyp (approved):
¡Muchas gracias!

PR #1146, opened 2021-07-25
handle remote keyboard inputs of multiple keys

ForrestHilton:
This pull request handles keyboard inputs of multiple keys, which can be generated by voice to text input. On my devices, text to speech still is not working correctly, but this edit makes it behave exactly the same as on KDE. The glitches need to be fixed on the phone side.

### DIFF
--- a/src/service/components/input.js
+++ b/src/service/components/input.js
@@ -609,13 +609,8 @@ class Controller {
         try {
             this._ensureAdapter();
 
-            // detecting if input was sent by text to speech
-            if (typeof input === 'string' && input.length > 1 && modifiers === 0) {
-                for (let i = 0; i < input.length; i++)
-                    this._session.pressKey(input[i], 0);
-                return;
-            }
-            this._session.pressKey(input, modifiers);
+            for (let i = 0; i < input.length; i++)
+                this._session.pressKey(input[i], modifiers);
         } catch (e) {
             return debug(e);
         }
--- a/src/service/components/input.js
+++ b/src/service/components/input.js
@@ -606,6 +606,14 @@ class Controller {
      * High-level keyboard input
      */
     pressKey(input, modifiers) {
+        // detecting if input was sent by text to speech
+        if ( typeof input == "string" && input.length > 1 && modifiers == 0 ) {
+            debug(input);
+            for (let i = 0; i < input.length; i++) {
+                this.pressKey(input[i], 0);
+            }
+        }
+
         try {
             this._ensureAdapter();
             this._session.pressKey(input, modifiers);
--- a/src/service/components/input.js
+++ b/src/service/components/input.js
@@ -606,19 +606,24 @@ class Controller {
      * High-level keyboard input
      */
     pressKey(input, modifiers) {
-        // detecting if input was sent by text to speech
-        if (typeof input === 'string' && input.length > 1 && modifiers === 0) {
-            debug(input);
-            for (let i = 0; i < input.length; i++) 
-                this.pressKey(input[i], 0);
-        }
-
         try {
             this._ensureAdapter();
-            this._session.pressKey(input, modifiers);
         } catch (e) {
-            debug(e);
+            return debug(e);
         }
+        
+        debug(input);
+        debug(modifiers);
+        
+        // detecting if input was sent by text to speech
+        if (typeof input === 'string' && input.length > 1 && modifiers === 0) {
+            for (let i = 0; i < input.length; i++) {
+                this._session.pressKey(input[i], 0);
+            }
+            return
+        }
+
+        this._session.pressKey(input, modifiers);
     }
 
     destroy() {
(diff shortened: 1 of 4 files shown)


ForrestHilton:
Hey, I see a couple issues with this and I will prove it later.

sonnyp (commented):
Looks interesting, thanks!

Could you share a bit more details? Which typeof of device are you using? Which text to speech solution?

So that the reviewer can try and reproduce this feature.

This wouldn't be the final solution but perhaps this would behave as expected, could you give it a try?

```js
    pressKey(input, modifiers) {
        try {
            this._ensureAdapter();
        } catch (e) {
            return debug(e);
        }
        
        debug(input);
        debug(modifiers);
        
        // detecting if input was sent by text to speech
        if (typeof input === 'string' && input.length > 1 && modifiers === 0) {
            for (let i = 0; i < input.length; i++) {
                this._session.pressKey(input[i], 0);
             }
             return
        }

        this._session.pressKey(input, modifiers);
    }
```

ForrestHilton:
I think I used Gboard, and it is only on SM-T330NU that I have been able to get anything to send with voice to text.

ForrestHilton:
Your version behaves perfectly. Unlike my pull request. Thank you for your time. 

I never actually tested what would happen if you sent modifier keys with a multi caricature packet. 
As for preventing the phone from sending duplicate packets, and making it work for all models, that needs to be handled on the phone's end.

sonnyp:
> Your version behaves perfectly, Unlike my pull request. Thank you for your time.

Team effort!

If you would like this to be released in GSConnect, some refactoring is gonna be necessary.

I would suggest to start with creating a `pressKeys` method next to `pressKey`. Then find the place were `pressKey` is called and called `pressKeys` instead `pressKey` if it's a list of characters.

It might as well be enough, let's see.

ForrestHilton:
I read your deleted statements, and I think I will start there.

ForrestHilton:
Thanks for giving me a change to look at this. I think I found all the usage of that function, but I don't know if a straight rename is exactly what you want, and if it isn't I don't know enough to adjust it.

sonnyp:
Apologize, I removed my previous comment because I misread something in the code and thought it didn't make sense.

src/service/components/input.js
```diff
@@ -605,13 +605,27 @@ class Controller {
...
+        if (typeof input === 'string' && input.length > 1 && modifiers === 0) {
+            for (let i = 0; i < input.length; i++)
+                this._session.pressKey(input[i], 0);
+            return;
         }
+        this._session.pressKey(input, modifiers);
+    }
+
+    pressKey(input, modifiers) {
+        debug(new Error().stack);
```
sonnyp, line 628: not used ?

sonnyp (commented):
Please wrap the all of the code in the try/catch

not sure why it's like this, but I'm sure there is a good reason

sonnyp:
BTW I found this on the topic https://bash-prompt.net/guides/android-dictation/

src/service/components/input.js
```diff
@@ -605,13 +605,27 @@ class Controller {
...
         } catch (e) {
-            debug(e);
+            return debug(e);
+        }
+        
+        debug(input);
+        debug(modifiers);
+        
+        // detecting if input was sent by text to speech
+        if (typeof input === 'string' && input.length > 1 && modifiers === 0) {
```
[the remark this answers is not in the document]
andyholmes: > Not sure why we shouldn't accept multiple keys with a modifier and pass the modifier along with each key?

As far as I know kdeconnect-android will never send modifiers with strings of text, but it probably wouldn't hurt either. Might be worth leaving a code comment either way.

[relevant kdeconnect-android snippet](https://github.com/KDE/kdeconnect-android/blob/master/src/org/kde/kdeconnect/Plugins/MousePadPlugin/KeyListenerView.java#L85-L89).

ForrestHilton:
I think this is ready to merge. Thank you for your time, and patience.

As for that tutorial, I don't know why it didn't work on my phone, but I can see that the phone never sent anything at all when I use the voice input despite everything else working.

src/service/components/input.js
```diff
@@ -605,12 +605,14 @@ class Controller {
...
+    pressKeys(input, modifiers) {
         try {
             this._ensureAdapter();
-            this._session.pressKey(input, modifiers);
+
+            for (let i = 0; i < input.length; i++)
+                this._session.pressKey(input[i], modifiers);
         } catch (e) {
-            debug(e);
+            return debug(e);
```
sonnyp, line 615:
```suggestion
            debug(e);
```

sonnyp (approved):

sonnyp:
Thanks! before merging I will fix tests and add one for this

daniellandau:
I fixed the existing tests, @sonnyp did you want to look into adding a new test for new functionality?

PR #1147, opened 2021-07-27
Show SHA256 fingerprints

### LINKED ISSUE #1091: Fingerprint verification is no longer accurate.
Since a few months ago, KDE Connect stopped using a pair of SHA1 fingerprint hashes to confirm pairing, and now uses a SHA256 hash of the combination of both certificates.

This is unfortunate since it means any older version of KDE Connect or GSConnect connecting to an Android device will appear to be fraudulent. The affects all versions of GSConnect up to v46.

See:

* https://invent.kde.org/network/kdeconnect-kde/commit/e7518493df7398f27f7dffbfc3f79750bc1fda50
* https://invent.kde.org/network/kdeconnect-android/commit/d09ad45e112bf5ce5e50ab4bb0ed651dccdc8775

strugee:
I changed the monkeypatched function name to improve code clarity (being unfamiliar with the codebase, it was unclear to me at the call sites what kind of fingerprint the function returned) and also to avoid clashing with future GLib APIs.

Ref #1091

### DIFF
--- a/src/service/__init__.js
+++ b/src/service/__init__.js
@@ -354,22 +354,22 @@ Gio.TlsCertificate.new_for_paths = function (certPath, keyPath, commonName = nul
 
 Object.defineProperties(Gio.TlsCertificate.prototype, {
     /**
-     * Compute a SHA1 fingerprint of the certificate.
+     * Compute a SHA256 fingerprint of the certificate.
      * See: https://gitlab.gnome.org/GNOME/glib/issues/1290
      *
-     * @return {string} A SHA1 fingerprint of the certificate.
+     * @return {string} A SHA256 fingerprint of the certificate.
      */
-    'fingerprint': {
+    'sha256': {
         value: function () {
             if (!this.__fingerprint) {
                 const proc = new Gio.Subprocess({
-                    argv: [Config.OPENSSL_PATH, 'x509', '-noout', '-fingerprint', '-sha1', '-inform', 'pem'],
+                    argv: [Config.OPENSSL_PATH, 'x509', '-noout', '-fingerprint', '-sha256', '-inform', 'pem'],
                     flags: Gio.SubprocessFlags.STDIN_PIPE | Gio.SubprocessFlags.STDOUT_PIPE,
                 });
                 proc.init(null);
 
                 const stdout = proc.communicate_utf8(this.certificate_pem, null)[1];
-                this.__fingerprint = /[a-zA-Z0-9:]{59}/.exec(stdout)[0];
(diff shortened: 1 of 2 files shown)


andyholmes:
Thanks for the patch!

Unfortunately this doesn't quite solve the problem. It probably wasn't clear in the issue I opened, because the method used here is a bit odd. What KDE Connect does now is to:

1. Take the DER format of each certificate
2. Compare them byte-by-byte to find the largest
3. Concatenate them into a single bytearray, largest first
4. generate a SHA256 hash of the combined bytearray

The Qt/C++ code is probably clearest here (comments are mine):

```cpp
QByteArray Device::verificationKey() const
{
    // Get the DER binary format of the local and remote TLS certificates
    auto a = KdeConnectConfig::instance().certificate().publicKey().toDer();
    auto b = certificate().publicKey().toDer();

    // Figure out which is  "logically" larger and ensure it is first (most significant)
    if (a < b) {
        std::swap(a, b);
    }

    // Add them sequentially and produces the SHA256 hash
    QCryptographicHash hash(QCryptographicHash::Sha256);
    hash.addData(a);
    hash.addData(b);
    return hash.result().toHex();
}
```

The reason it does this is so that both devices show precisely the same verification key. I'm not sure there's an elegant way to do this in GJS, given the current restrictions we work with.

Probably it will require getting the PEM of each certificate, using `openssl` to dump the DER format somewhere, opening them with GFile, then using [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/uint8array) to compare the size and combine them, before finally passing them to [`GLib.Checksum`](https://gjs-docs.gnome.org/glib20/glib.checksum). At least that's what occurs to me off the top of my head.

strugee:
Ahhh, funky. Ok, makes sense.

What _doesn't_ make sense though is my KDE Connect Android app still showing two distinct fingerprints. Maybe F-Droid is out of date? 1.17.0 is what I'm running. I'll look into this more in the morning :+1:

strugee:
I have terrible time management and looked into this instead of going to sleep because it didn't take very long. The SHA256 fingerprints I was seeing in the Android app were in the "Encryption Info" screen - I completely ignored the verification code which is what's computed using that algorithm :sweat_smile:

IIUC this change still needs to be made anyway, so I amended the commit message so it doesn't close the linked issue. The verification code will have to be implemented separately.

andyholmes:
Ah right, I had forgotten about that :stuck_out_tongue_closed_eyes:. Okay I will merge this one, thanks for the PR!

PR #1148, opened 2021-07-29
Use KDE Connect's verification codes

### LINKED ISSUE #1091: Fingerprint verification is no longer accurate.

strugee:
Fixes #1091

Still TODO:

* Diagnose segfaults in the daemon
* Figure out if the UTF-8 errors are a problem
* Display verification codes in the notification (right now you need to have the preferences window open)
* Clean up the style
* Clean up the .ui file diff that Glade made a mess of
* Fix the verification code computation so it actually works

Some notes on the segfaults: they seem to happen during the pairing process. I think there's a race condition or something like that during the verification code computation. Most of the time I just get `Segmentation fault (core dumped)`, but occasionally I get a longer assertion error:

```
GLib:ERROR:../glib/gchecksum.c:1618:g_checksum_update: code should not be reached
Bail out! GLib:ERROR:../glib/gchecksum.c:1618:g_checksum_update: code should not be reached
Aborted (core dumped)
```

The UTF-8 errors come from the preferences pane and all look like this:

```
(gsconnect-preferences:335958): Gjs-CRITICAL **: 01:53:39.208: JS ERROR: TypeError: malformed UTF-8 character sequence at offset 0
@/var/home/alex/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/gsconnect-preferences:114:21
```


strugee:
Also, I used Unicode 🔑 instead of an icon because I couldn't find a suitable icon in what Glade said was available. If there's a better way, let me know.

One more thing - this is basically my first time writing anything approaching GTK/GObject/etc. code, so please review with extra scrutiny. I might have broken everything!

andyholmes:
> Diagnose segfaults in the daemon

You can run the daemon with GDB if necessary. Let me know if that's something you want help with and after I can add it to the contributing document.

> Figure out if the UTF-8 errors are a problem

These are fine to leave and not related to your PR. It's just a corner-case in GJS and GDBus related to virtual functions.

> Display verification codes in the notification (right now you need to have the preferences window open)

I think that the combined verification code is what needs to be shown in the notification, at least I think that's what Android does. I think the older keys are effectively being deprecated.

src/service/device.js
```diff
@@ -186,20 +198,66 @@ var Device = GObject.registerClass({
...
+    get verify_code() {
+        let verifyCode = _('Not available');
+        const [remoteCert, localCert] = this.getCerts();
+
+        // XXX MAKE THIS WORK
+        // let a = ByteArray.toGBytes(localCert.get_certificate());
+        // let b = ByteArray.toGBytes(remoteCert.certificate);
+        let a = ByteArray.toGBytes(ByteArray.fromString('a'));
+        let b = ByteArray.toGBytes(ByteArray.fromString('b'));
+        // Lexographic a < b comparison; this is what the < operator on QByteArray does
```
ferdnyc, line 247: Nit: "lexicographic"

src/service/device.js
```diff
@@ -186,20 +198,66 @@ var Device = GObject.registerClass({
...
+            [a, b] = [b, a]; // swap
+
+
+        const checksum = new GLib.Checksum(GLib.ChecksumType.SHA256);
+        checksum.update(ByteArray.fromGBytes(a));
+        checksum.update(ByteArray.fromGBytes(b));
+        verifyCode = checksum.get_string();
+        checksum.free();
+
+        return verifyCode;
```
ferdnyc, line 258: Your comment mentioned segfaults... according to [the Checksum docs](https://gjs-docs.gnome.org/glib20~2.66.1/glib.checksum#method-get_string)...

> `get_string()`
> Returns:
> (`String`) — the hexadecimal representation of the checksum. The
> returned string is owned by the checksum and should not be modified
> or freed.

I'm gonna go out on a limb and say that means it's probably _NOT_ OK to access the Checksum-owned string _after_ freeing the Checksum. Perhaps, to be safe...?
```js
verifyCode = GLib.strdup(checksum.get_string());
```

src/preferences/device.js
```diff
@@ -306,15 +309,30 @@ var Panel = GObject.registerClass({
...
+
+        this.device.bind_property(
+            'verify-code',
+            this.verifycode,
+            'label',
+            GObject.BindingFlags.SYNC_CREATE
+        );
 
+        // Action setup
         this._setupActions();
```
ferdnyc, line 336: Phew, there's that mystery solved! :smiling_imp:

src/service/device.js
ferdnyc: (Note that you'd then have to free the newly-allocated `verifyCode` String at some later point, once you're finished with it.)

ferdnyc (commented):
Hey, I know this has been here a good long while already, but I jotted down a few things as I was perusing it just now.

src/service/device.js
andyholmes: Well, that is bizarre so I've asked about it in the #javascript channel. SpiderMonkey should be able to handle this without the `GLib.strdup()`, but I'm surprise that function is even in our bindings and that it seems to work! `GLib.free()` is definitely a no-go in GJS, so I'm not sure if this will leak memory or if SM can free it.

ferdnyc: @andyholmes Heh, always fun when the soft underbelly of a native C API is exposed to a higher-level lanugage. Just like Python's lazy garbage collector frees developers from worrying much about allocation lifetimes... _in theory_. 

Until you find yourself working with SWIG-generated C/C++ library bindings in Python. Then, unless both the native library and its bindings are _really_ meticulous, you **definitely** start to care about your object lifetimes. In C++, not properly cleaning up your allocations "merely" leaks memory. But if Python's gc goes stomping around, cleaning up _**for**_ you, invalidating pointers it hasn't been told are still very much in use over on the native side, You're Gonna Have A Bad Time™.

ferdnyc:
Closing this as completed, since it was brought home by @daniellandau in #1493 and is now merged. Thanks @strugee !

ferdnyc:
Whoops. Premature close, as there's more to this PR than what went in with #1493. Reopening, and attempting to reconcile with the current `main` branch HEAD.

ferdnyc:
> attempting to reconcile with the current `main` branch HEAD.

That actually wasn't hard at all, in terms of actual change conflicts. Whether the code on this branch is still compatible with the `main` source is another matter, which I guess the CI will at least help us determine...

strugee:
Hey, sorry for disappearing - life got in the way :-)

Glad to hear my work was useful for #1493.

src/preferences/device.js
strugee: Lol, I recall feeling the same way writing it. But it looked out of place otherwise 😂

andyholmes:
Thanks for the work on this! I'll close this up now, since we've integrated the changes.

PR #1150, opened 2021-08-07
Update German translation

daPhipz:
I am a member of the German l10n team for the GNOME Project.
In this PR, I translated some untranslated strings and edited previous translations in order to comply with the rules of our translation team.

If you need me to update not only `po/de.po`, but also the `pot` file, I'll do that, of course. I followed [these](https://github.com/GSConnect/gnome-shell-extension-gsconnect/wiki/Translating#option-2-git) translation instructions.
Have a nice weekend!


po/de.po
```diff
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: gsconnect\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-12-18 19:41+0100\n"
-"PO-Revision-Date: 2020-12-18 20:11+0100\n"
+"POT-Creation-Date: 2021-08-07 07:51+0200\n"
```
sonnyp, line 5:
```suggestion
"POT-Creation-Date: 2020-12-18 19:41+0100\n"
```

This field shouldn't be updated according to https://www.gnu.org/software/trans-coord/manual/gnun/html_node/PO-Header.html

sonnyp (approved):
Vielen Dank!

po/de.po
daPhipz: Yeah, when following the wiki instructions for translating GSConnect, one step is to update the `pot` file. When doing so, this line changes in every `po` file - which is intended.
But since I didn't commit the new `pot` file, you are right! Thanks for noticing!

Issue #1116, opened 2021-05-18
share-text option in cli does not work


CharlieQiu2017:
> ***ATTENTION**: GSConnect only supports the latest, stable version of GNOME. We are no longer accepting bug reports for previous versions.*

**Describe the bug**

The GSConnect CLI provides an option "share-text", but using this option results in error "options.lookup_value is null"
In file `src/service/daemon.js`, change line 717 to
`this._cliShareText(id, options);`
resolves this issue.

**Steps To Reproduce:**

1. Open terminal
2. Execute `cd .local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/`
3. Execute `./daemon.js -d <device_id> --share-text="Hello!"`
4. The program crashes. The phone does not receive any text.

**Expected behavior**

The text is successfully sent to my phone

**System Details (please complete the following information):**

 - **GSConnect version:** 43
   - **Installed from:** GNOME Extensions Website
 - **GNOME/Shell version:** 3.36
 - **Distro/Release:** Pop_OS 20.04 LTS

**GSConnect environment (if applicable):**

 - **Paired Device(s):** Samsung Galaxy S10e
 - **KDE Connect app version:** 1.17.0
 - **Plugin(s):** N/A

andyholmes:
Thanks for reporting, you can open a PR if you like.

PR #1162, opened 2021-08-26
fixed --share-text throwing error

adarshhegde:
Fixed bug #1116 where adding share-text would throw an error


andyholmes:
Looks good, thanks!

PR #1168, opened 2021-09-09
Update URLs in AppStream XML

ferdnyc:
Just a quick fix to switch a bunch of `github.com/andyholmes/...` URLs over to `github.com/GSConnect/...`, where appropriate. (Meaning: all current values, and historical items after December 2020.)


ferdnyc:
I was going to submit a similar PR to the Fedora packaging repo, and suddenly noticed we weren't exactly setting the best example on that front. :wink:

PR #1169, opened 2021-09-12
MousePad Plugin: add abilities to receive a string as a set of key input

knard:
from the code implemented in KDEConnect Android (KeyListenerView.java) or
in KDEConnect (windowsremoteinput.cpp or x11remoteinput.cpp) the value
associated the "key" property of a network packet can contains multiple
charaters.


sonnyp:
Thank you. This is already being done in https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1146/files

knard:
Yw, I forgot to check pull request.

PR #1178, opened 2021-10-01
Add Finnish translations

jonnelafin:
Remove the compiled translation

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -7,6 +7,7 @@ de
 es
 et
 fr
+fi
 gl
 hu
 it
--- a/po/fi.po
+++ b/po/fi.po
@@ -1,0 +1,1128 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the org.gnome.Shell.Extensions.GSConnect package.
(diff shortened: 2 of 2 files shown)


po/LINGUAS
```diff
@@ -7,6 +7,7 @@ de
 es
 et
 fr
+fi
```
daniellandau, line 10: Pitäisi olla fi_FI

jonnelafin: Or should it? I submitted a translation for guake recently and they noticed it shoud be the opposite according to the ISO standard (at least according to [wikipedia](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)) I found two other projects where contributors were complaining on the use of fi_FI: [nextcloud](https://github.com/nextcloud/appstore/issues/479) and [woocommerence](https://github.com/woocommerce/woocommerce/issues/6302).

My commit however contradicts with itself, using fi once and fi_FI on the actual po file. I can change one of these to match the other. I myself have no preference. Do you want for this project to use fi or fi_FI?

P.S: Olisin voinu kirjottaa kans suomeks mut ajattelin et parempi kirjottaa englanniks jos joku muu törmää tähän tulevaisuudessa.

daniellandau: Either way is fine for me.

jonnelafin: I amended a commit changing both to just "fi"

PR #1233, opened 2022-01-06
RunCommand Plugin: fix menu item replacement

### LINKED ISSUE #1028: gsconnect causes gnome-shell to lag and possibly memory leak
**Describe the bug**

When GSConnect extension is enabled the whole desktop stops responding for a second or two every minute or a few minutes. Only mouse moves. Video in chrome stops, sound keeps playing. I also noticed slowly increasing MEM% value in htop. (under 1 when gsconnect disabled, ~1.2 when enabeld, over 4 after some time of running). It probably started after a gnome update.

**Steps To Reproduce:**

1. just enable the extension and try to use desktop normally (google-chrome, and other apps -- it is not chrome related, it happens also with chrome not running) 

**Expected behavior**

no lags, no memleak

**Screenshots**

**Support Log**

[gsconnect.log](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/5683585/gsconnect.log)

 - **GSConnect version:** 44
   - **Installed from:** GNOME Extensions Website (not sure, sorry)
 - **GNOME/Shell version:** 3.38.2
 - **Distro/Release:** debian

 - **Paired Device(s):** xiaomi android phone
 - **KDE Connect app version:** 
 - **Plugin(s):** all enabled, didn't tested to disable some

**Additional Notes:**

andyholmes:
When the menu item was being removed, the action name wasn't prefixed
with the scope (eg. `device.commands`), resulting in an infinite amount
of invisible menu items being exported over D-Bus.

fixes #1028

### DIFF
--- a/src/service/plugins/runcommand.js
+++ b/src/service/plugins/runcommand.js
@@ -197,7 +197,7 @@ var Plugin = GObject.registerClass({
         const index = menuActions.indexOf('commands');
 
         if (index > -1) {
-            this.device.removeMenuAction('commands');
+            this.device.removeMenuAction('device.commands');
             this.device.addMenuItem(item, index);
         }
     }


daniellandau:
I'm not totally convinced that there couldn't be other issues causing lag and stuttering (for example this doesn't solve the system log filling issue), but at least this solves this one!

Issue #1227, opened 2021-12-28
[Question] What's the purpose of TOOLTIP_BROWSE_ID in tooltip.js#L18?


nlpsuge:
Hi,

I'm adopting tooltip.js into my project, but I don't understand the purpose of TOOLTIP_BROWSE_ID in tooltip.js#L18:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/8e85f0fc0490dbf92d16a31ee5d73b46ae43e653/src/shell/tooltip.js#L18

It's never assigned to a meaningful value (which is a non-zero value in the case). Looks like it and some other code using it can be removed?

andyholmes:
Ah, that's a mistake. It should probably be set as the return value of a call to `GLib.timeout_add()`, so that it's disabled after a period. Probably 500ms, from glancing at the source of `gtktooltip.c`.

nlpsuge:
> It should probably be set as the return value of a call to `GLib.timeout_add()`

Assign this line's return value to TOOLTIP_BROWSE_ID, right? looks very likely, any way :)
https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/8e85f0fc0490dbf92d16a31ee5d73b46ae43e653/src/shell/tooltip.js#L246

andyholmes:
That seems right :)

PR #1236, opened 2022-01-10
Remove tooltip_browse_id in destory() to meet the review guidelines

nlpsuge:
This PR moves two global variables into Tooltip and removes tooltip_browse_id in destory() to meet the review guidelines. (https://extensions.gnome.org/review/28736)

I adopted tooltip.js in [Another Window Session Manager](https://github.com/nlpsuge/gnome-shell-extension-another-window-session-manager), it's useful when the users want to know what does an icon do.

Tested on my machine, works.

Fix: https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1227

### DIFF
--- a/src/shell/tooltip.js
+++ b/src/shell/tooltip.js
@@ -15,6 +15,9 @@ const Main = imports.ui.main;
  * Adapted from: https://github.com/RaphaelRochet/applications-overview-tooltip
  * See also: https://github.com/GNOME/gtk/blob/master/gtk/gtktooltip.c
  */
+var TOOLTIP_BROWSE_ID = 0;
+var TOOLTIP_BROWSE_MODE = false;
+
 var Tooltip = class Tooltip {
 
     constructor(params) {
@@ -23,9 +26,6 @@ var Tooltip = class Tooltip {
         this._bin = null;
         this._hoverTimeoutId = 0;
         this._showing = false;
-
-        this._tooltip_browse_id = 0;
-        this._tooltip_browse_mode = false;
 
         this._destroyId = this.parent.connect(
             'destroy',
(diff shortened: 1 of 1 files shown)


andyholmes (changes requested):
The tooltip browse mode needs a "global" of some sort to work properly. If that's not possible, browse mode should just be removed.

The way this works is to use a single variable for all widgets with tooltips. It tracks when the mouse moves between widgets quickly, so that the user doesn't have to wait the full 500ms when they're "browsing" the tooltips for information. Then when the mouse is not on any of the widgets long enough for the hover timeout to expire, the browse mode is disabled.

So the fix here is to either move the timeout ID to an ancestor widget that contains all the widgets with tooltips, or just remove it entirely.

nlpsuge:
> The way this works is to use a single variable for all widgets with tooltips. It tracks when the mouse moves between widgets quickly, so that the user doesn't have to wait the full 500ms when they're "browsing" the tooltips for information.

Now, I understand what the 'browse mode' is.

I tested with the gsconnect released at the https://extensions.gnome.org/extension/1319/gsconnect/, seems it doesn't have the 'browse mode' actually, the code is not fully working. I hover the cursor on an icon and wait until the tip appears and then move to another icon beside it, the tip still waits 500ms to show up, if I move the cursor quickly, the tips of the icons don't show up.

I'll modify the code in my extension, and push the changes here.

I should keep the TOOLTIP_BROWSE_ID and TOOLTIP_BROWSE_MODE global variables and remove TOOLTIP_BROWSE_ID in the destroy().

andyholmes:
[Screencast from 2022-01-09 05:57:59 PM.zip](https://github.com/GSConnect/gnome-shell-extension-gsconnect/files/7848882/Screencast.from.2022-01-09.05.57.59.PM.zip)

Here's a screencast of it working for me, is this not how it works for you? I made this the other day, but realized I'd have to ZIP it so didn't bother adding it :P

ferdnyc:
> I made this the other day, but realized I'd have to ZIP it so didn't bother adding it :P

https://ezgif.com/, for all your video-to-GIF needs. (Animated gifs embed fine.)

nlpsuge:
> Here's a screencast of it working for me, is this not how it works for you?

No, I tested again just now. It works well. 

BTW, I think two reasons make the previous test not right. 
1. The gap between the icons is a little bit large and the icon is a little bit small too, it's a little bit easy to move the cursor out of the icon... 
2. And another extension `Net speed Simplified` always change the network speed, and that will change the position of the panel menu, and the tool tip could disappear due to this.

nlpsuge:
I've pushed changes into this PR. I've tested on my machine, works.

andyholmes (approved):

andyholmes:
Thanks!

Issue #1244, opened 2022-02-06
CJK text copied from Firefox on my Arch with GSConnect, then paste on my android phone become escaped UTF-8 uncoide characters


everyx:
**Describe the bug**

When I copy any CJK text on my Firefox (version: 96.0.3) and sync clipboard to my android phone, it's become escaped UTF-8 uncoide characters, but on telegram, It's worked great.

For example: `嗯` -> `\u55ef`

**Steps To Reproduce:**

1. Go to 'Firefox web browser'
2. Find any CJK characters, such as '嗯' here
3. Copy it
4. Paste on a android phone which connected before.

**Expected behavior**

We will see `嗯`

**Support Log**

```

GSConnect: 47 (user)
GJS:       17000
Session:   wayland
OS:        Arch Linux
--------------------------------------------------------------------------------
2月 06 20:47:02 org.gnome.Shell.Extensions.GSConnect[1513]: [/service/device.js:sendPacket:446]: XiaoMi 10s: {
                                                               "id": 1644151622992,
                                                               "type": "kdeconnect.clipboard",
                                                               "body": {
                                                                 "content": "\\u55ef"
                                                               }
                                                             }
```

**System Details (please complete the following information):**

 - **GSConnect version:** [e.g. 35, 36, ... (displayed in the GSConnect "About" window)]
   - **Installed from:** [e.g. GNOME Extensions Website, GitHub, Package Manager, ...]
 - **GNOME/Shell version:** [e.g. 3.36, ...]
 - **Distro/Release:** [e.g. Ubuntu 18.04, Fedora 29, Arch, ...]

**GSConnect environment (if applicable):**

 - **Paired Device(s):** [e.g. Galaxy Note 8, Pixel 2, Honor 9, ...]
 - **KDE Connect app version:** [e.g. 1.12.7]
 - **Plugin(s):** [if the issue only occurs when using certain plugin(s)]

**Additional Notes:**

```shell
                   -`                    everyx@arch 
                  .o+`                   ----------- 
                 `ooo/                   OS: Arch Linux x86_64 
                `+oooo:                  Host: TM1604 XMAKB3M0P1A12 
               `+oooooo:                 Kernel: 5.16.5-arch1-1 
               -+oooooo+:                Uptime: 5 hours, 13 mins 
             `/:-:++oooo+:               Packages: 1435 (pacman), 30 (flatpak) 
            `/++++/+++++++:              Shell: fish 3.3.1 
[...]

PR #1245, opened 2022-02-06
fix: prefer text/plain;charset=utf-8

everyx:
Close: #1244


andyholmes:
Good catch and a nice fix, thanks!

PR #1299, opened 2022-03-16
Protect against empty SMS address values

ferdnyc:
As the branch name indicates, this started out as a quick update to the docstring for `MessageBox` in the sms plugin, since it failed to document any of the current values.

But then I discovered that I was having a problem with contact matching on my system, because one of my contacts has an _empty_ phone number listed. That number was therefore matching basically _**every**_ address that came over the wire, since the test for matching just used `foo.endsWith(bar)` as the condition, and everything ends with `""`.

So, the real meat of this PR is the second commit, which first sanity-checks that neither string being matched is empty before testing for address matches. WorksForMe™, in repairing my Messaging window's ability to identify the participants in SMS conversations.

### DIFF
--- a/src/service/__init__.js
+++ b/src/service/__init__.js
@@ -167,7 +167,7 @@ String.prototype.equalsPhoneNumber = function (number) {
     const a = this.toPhoneNumber();
     const b = number.toPhoneNumber();
 
-    return (a.endsWith(b) || b.endsWith(a));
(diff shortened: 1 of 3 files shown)


andyholmes:
I made you an admin, then changed the branch protection rules to require linear history and the `test` job to pass.

Either it was not running because your branch was in fork (maybe some weird git tracking issue?) or perhaps there was a non-linear history issue?

PR #1315, opened 2022-03-26
Presenter Plugin: fix handling of specialKey in input

### LINKED ISSUE #1341: Presentation Remote does not work in Gnome 42 Pop OS 22.04
> ***ATTENTION**: GSConnect only supports the latest, stable version of GNOME. We are no longer accepting bug reports for previous versions.*

**Describe the bug**

Presentation remote does not work on Gnome 42 (Pop OS 22.04)

**Steps To Reproduce:**

1. Go to KDE Connect Android App
2. Click on Presentation Remote
3. Click next, previous, full screen, or exit presentation
4. Nothing happen on host computer (tested in Evince and Okular)

**Expected behavior**

The presentation remote should work in Evince and/or Okular

**Screenshots**

image

**Support Log**

Please generate a support log ([Instructions](../wiki/Help#generate-support-log)) and paste any messages related to this issue between the two ``` lines below.

```
```

 - **GSConnect version:** 50
   - **Installed from:** GNOME Extensions Website
 - **GNOME/Shell version:** 42
 - **Distro/Release:** Pop OS 22.04

 - **Paired Device(s):** Xiaomi 11 Lite 5G
 - **KDE Connect app version:** 1.19.1
 - **Plugin(s):** [if the issue only occurs when using certain plugin(s)]

**Additional Notes:**

Add any additional information about the problem or your system.

not-chicken:
Fix passing key symbols (ie. `Number`) to a function expecting a character or string of characters.

Fixes #1341

### DIFF
--- a/src/service/components/input.js
+++ b/src/service/components/input.js
@@ -609,8 +609,12 @@ class Controller {
         try {
             this._ensureAdapter();
 
-            for (let i = 0; i < input.length; i++)
-                this._session.pressKey(input[i], modifiers);
+            if (typeof input === 'string') {
+                for (let i = 0; i < input.length; i++)
+                    this._session.pressKey(input[i], modifiers);
+            } else {
+                this._session.pressKey(input, modifiers);
+            }
         } catch (e) {
             debug(e);
         }


src/service/components/input.js
```diff
@@ -609,8 +609,12 @@ class Controller {
...
             this._ensureAdapter();
 
-            for (let i = 0; i < input.length; i++)
-                this._session.pressKey(input[i], modifiers);
+            if (typeof input === 'string') {
+                for (let i = 0; i < input.length; i++)
+                    this._session.pressKey(input[i], modifiers);
+            } else {
+                this._session.pressKey(input, modifiers);
+            }
```
ferdnyc, line 617: I'm a little wary of passing `input` to `pressKey()` without being more sure what type it _does_ have, if it's not a string. What other types would it be getting? Keysyms, I guess, which should be Numbers.

How do you feel about a slightly more explicit check?
```suggestion
            const in_type = typeof input;
            switch(in_type) {
                case 'number':
                    this._session.pressKey(input, modifiers);
                    break;
                case 'string':
                    for (let i = 0; i < input.length; i++)
                        this._session.pressKey(input[i], modifiers);
                    break;
                default:
                    throw new Error(`Unrecognized input type '${in_type}'`);
            }
```

not-chicken: It shouldn't get any other type as input since `pressKeys` gets either `input.key` which should always be a string ([kde-connect android code](https://github.com/KDE/kdeconnect-android/blob/master/src/org/kde/kdeconnect/Plugins/MousePadPlugin/KeyListenerView.java#L85-L89)), or `input.specialKey` which is already checked to contain a valid keysym [here](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/master/src/service/plugins/mousepad.js#L193).

andyholmes:
Thanks!

PR #1321, opened 2022-04-01
Support wayland enviroment without gnome

JingMatrix:
I am using `sway`, this commit gives support to `sway` for the clipboard function without breaking original codes.
I used `wl-clipboard`.


JingMatrix:
Forget to run eslint, I will do another pull-request

ferdnyc:
...I'm confused. How do you run a GNOME Shell extension _without_ GNOME!??!

JingMatrix:
Yes, actually you can, just need `gjs` installed.
Please discuss under #1322 , the code in this pull request is not that good.

PR #1322, opened 2022-04-02
Cli support for general wayland environment

JingMatrix:
This commit use `wl-clipboard` and `ydotool` to support general wayland environment, such as sway.


JingMatrix:
Please tell me if you have intention to extend this plugin to non-gnome environment, @ferdnyc @andyholmes .
If so, I can add a page to the project wiki to show how it actually works.

andyholmes:
I don't have much of an opinion on this. GSConnect was intended to be specifically for GNOME Shell, but if another maintainer wants to give it a :+1: I won't stand in the way :slightly_smiling_face:

daniellandau:
I have limited time for maintenance, so if making it more generic comes with associated maintenance effort on an ongoing basis I'm all for it. I'll read through the code hopefully soon.

src/service/components/clipboard.js
```diff
@@ -105,6 +117,19 @@ var Clipboard = GObject.registerClass({
             );
 
             this._onOwnerChange();
+            if (!HAVE_GNOME) {
+                // Directly subscrible signal
+                this.signalHanlder = Gio.DBus.session.signal_subscribe(
```
andyholmes, line 122:
```suggestion
                this.signalHandler = Gio.DBus.session.signal_subscribe(
```

src/service/components/clipboard.js
```diff
@@ -276,6 +301,10 @@ var Clipboard = GObject.registerClass({
             Gio.bus_unwatch_name(this._nameWatcherId);
             this._nameWatcherId = 0;
         }
+
+        if (!HAVE_GNOME && this.signalHanlder)
+            Gio.DBus.session.signal_unsubscribe(this.signalHanlder);
```
andyholmes, line 306:
```suggestion
        if (!HAVE_GNOME && this.signalHandler)
            Gio.DBus.session.signal_unsubscribe(this.signalHandler);
```

andyholmes (changes requested):
Sorry this sat around for so long. I don't see any reason not to merge at this point, so a few fixups and I will accept this.

andyholmes:
Oh, just one more thing: add the SPDX comments to `ydotool.js` and `wl_clipboard.js` like so:

```js
// SPDX-FileCopyrightText: GSConnect Developers https://github.com/GSConnect
//
// SPDX-License-Identifier: GPL-2.0-or-later
```

JingMatrix:
Thanks for reviewing.
Unfornately, I have forgotten most of the logic of gsconnect, though I am using this pull request for the whole year.
It seems after merging the main branch, remote mousepad isn't working. I need some time to debug it.
Also, I should add a wiki page to explain the usage.

andyholmes:
Okay, no problem. I don't expect many folks will start using this for a little while, at least.

If you can get it working for you, we can merge and you can be "codeowner" for that module, if you like.

JingMatrix:
I test it on my machine with both sway and gnome-shell sessions,  and there is no problem. (It turned out that I forgot to start ydotoold in sway.)
I would be happy to maintain this module.

### Notes for dev
1. The enviroment variable `XDG_SESSION_TYPE`  is not reliable as user can change it or it is not set at all, but gnome doesn't change it back to 'gnome'.
2. Not sure how to pass the test that gives error:
```
(gjs:475): Gjs-WARNING **: 18:11:09.216: JS ERROR: Mock Backend: ImportError: No JS module 'ydotool' found in search path
mockComponents/Components.acquire@/usr/libexec/installed-tests/gsconnect/fixtures/utils.js:230:9
```

### Notes for users
When `gsconnect` is not running as a Gnome extension, for example, running under `sway`,
1. You might still  need to install the `gnome-shell` package to provide the `Gvc-1.0.typelib` file of the gnome module `libgnome-volume-control`.
4. Install `ydotool` and start the service: systemctl status --user ydotool.service 
5. Install `wl-clipboard` and `wl-type`

JingMatrix:
I found that the env GNOME_SETUP_DISPLAY is specially set by gnome, which can thus be used to detect gnome.

Now this pull request should pass all tests without problems, you can merge it.

P.S. it might be interesting to write a test for non-gnome envirment, maybe in the future.

andyholmes:
> P.S. it might be interesting to write a test for non-gnome envirment, maybe in the future.

Definitely, I saw somewhere I think Google Chrome or Chromium had a Wayland version of xvfb built with python or something. I never had time to really get into it though.

JingMatrix:
As a note, I add the wiki page of its usage: [CLI usage without Gnome](/GSConnect/gnome-shell-extension-gsconnect/wiki/CLI-usage-without-Gnome-environment).

It will be helpful to put some instructions described there into the meson build process.

andyholmes:
That's great, thanks for writing all that up!

JingMatrix:
Given what said by @pobrn , it is better not to use environment variable for detecting the 'Gnome'.
I was for using `imports` to detect `Gnome`, @andyholmes changed it out of the reason of simplicity (I guess).
Hence, it is favorable to revert the changes regarding the environment variable.
I will do it soon to fix #1706.

andyholmes:
FWIW, [this was my suggestion](https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1322#discussion_r1231377133):

> This won't work in the service, because it's runs outside of the `gnome-shell` process (meaning it will always throw an error). However, I think this would work:
> 
> ```js
> const HAVE_GNOME = GLib.getenv('XDG_SESSION_TYPE') === 'gnome'
> ```
> 
> You can put this in `src/service/__init__.js` as `globalThis.HAVE_GNOME` if you want, or leave it here.

JingMatrix:
@andyholmes Thanks for mentioning it, I totally forgot it. So I guess we may use the existence of the directory `/usr/share/cinnamon/js/ui/*` as a detector.

Or, alternatively, is there a way to communicate bewteen the `extention.js` and `daemon.js`?
If so, we can set `HAVE_GNOME` in `extension.js`.

JingMatrix:
I propose to use `imports.gi.Shell` as a heuristic for setting `HAVE_GNOME`.
See my pull-request for details: https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1715#issuecomment-1841296211

PR #1337, opened 2022-04-21
WIP: Initial Bluetooth handoff impl

Wazzaps:
This is an initial implementation of my new KDE connect feature:

image

It lets you move bluetooth devices between the PC and the phone. Mostly intended for wireless headphones.

TODO list:
- Add spinner while operation in progress
- Refine UX
- Try many more devices
  - Bluetooth LE devices probably won't work at all, though they might not need to (need to check if they have multipath by default)
- Make sure everything is translatable
- Un-uglify code (I don't know enough about GObjects to use it correctly)
  - There are numerous object lifetime problems (Why is this a problem in a GC'd language? Whatever), which cause soft-crashes of the extension.
    Example: `Attempting to run a JS callback during garbage collection. This is most likely caused by destroying a Clutter actor or GTK widget with ::destroy signal connected, or using the destroy(), dispose(), or remove() vfuncs. Because it would crash the application, it has been blocked.`
- BSD support (Decouple from bluez + find hardware with working bluetooth and bsd at the same time)
- Tests


rohmishra:
For BT LE, Multi-Stream and Broadcast is for source. AFAIK receiving from multiple sources is still vendor dependent and not required.

PR #1358, opened 2022-05-20
sms: Fix 'cache is not iterable'

ferdnyc:
This PR fixes the following (harmless) error messages in the journal output, when there's no previous cache for a particular SMS thread:

```text
GSConnect[749788]: [/service/plugins/sms.js:_handleDigest:229]: SM-A536U1: cache is not iterable
                   _handleDigest@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/sms.js:229:35
                   _handleMessages@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/sms.js:331:22
                   handlePacket@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/sms.js:199:22
                   handlePacket@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:421:25
                   _readLoop@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:345:22
                   async*setChannel@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:323:18
                   _onChannel@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/manager.js:225:20
                   channel@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/core.js:448:19
                   _onIncomingChannel@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:216:18
                   Async*@/home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:727:17
```

### DIFF
--- a/src/service/plugins/sms.js
+++ b/src/service/plugins/sms.js
@@ -225,7 +225,7 @@ var Plugin = GObject.registerClass({
             }
 
             // If this message is marked read, mark the rest as read
-            if (message.read === MessageStatus.READ) {
+            if (message.read === MessageStatus.READ && cache.length > 0) {
                 for (const msg of cache)
                     msg.read = MessageStatus.READ;
             }


daniellandau:
What is `cache` when it isn't iterable? Is it some object that doesn't have length? If it's undefined then wouldn't this just replace the old error with a new error?

ferdnyc:
> What is `cache` when it isn't iterable? Is it some object that doesn't have length? If it's undefined then wouldn't this just replace the old error with a new error?

It can't be `undefined`, we already checked for that and `continue`d to the next loop iteration a few lines previous. So it has to be some sort of non-iterable value (an empty string? an empty object?); either of those will fail a test of `cache.length > 0` without error.

ferdnyc:
I'm open to a better method of testing for an iterable `cache`, my JS is hardly expert-level. It doesn't appear that `typeof` will work, though, because an iterable array will still return `'object'` for `typeof cache`.

daniellandau:
I don't like having something that ends up as `undefined > 0` which does evaluate to false, but is a very confusing way to go about it.

ferdnyc:
> I don't like having something that ends up as `undefined > 0` which does evaluate to false, but is a very confusing way to go about it.

Fair enough. On further examination, I believe this may have occurred for me only because of some local changes that altered the structure of the cache contents — IOW, it shouldn't be something that anyone encounters normally. I'll close it as probably unnecessary.

PR #1360, opened 2022-05-22
Add v49 & v50 to metainfo

ferdnyc:
Just some housekeeping: Add the two most recent releases to the metainfo file.


daniellandau:
Thanks!

PR #1370, opened 2022-05-31
meson: remove deprecated functions

andyholmes:
Remove usage of `meson.source_root()` and `meson.build_root()`, then
bump the required version to `>= 0.56.0`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('gsconnect',
   'c',
   version: '53',
-  meson_version: '>= 0.46.0'
+  meson_version: '>= 0.59.0',
 )
 
 gnome = import('gnome')
@@ -19,13 +19,7 @@ libexecdir = join_paths(prefix, get_option('libexecdir'))
 localedir = join_paths(prefix, get_option('localedir'))
 sysconfdir = get_option('sysconfdir')
 
-
-# GSettings schema dir
-if get_option('gsettings_schemadir') != ''
-  gschemadir = get_option('gsettings_schemadir')
-else
-  gschemadir = join_paths(datadir, 'glib-2.0', 'schemas')
-endif
+gschemadir = join_paths(datadir, 'glib-2.0', 'schemas')
 
 
 # GNOME Shell LIBDIR
@@ -66,7 +60,7 @@ run_target(
     'DATADIR=' + datadir,
     'LOCALEDIR=' + localedir,
     'GSCHEMADIR=' + gschemadir,
-    join_paths(meson.source_root(), 'build-aux', 'ego', 'mkzip.sh')
+    join_paths(meson.project_source_root(), 'build-aux', 'ego', 'mkzip.sh')
   ]
 )
 
@@ -79,19 +73,17 @@ run_target(
     'LOCALEDIR=' + localedir,
     'GSCHEMADIR=' + gschemadir,
     'INSTALL=true',
-    join_paths(meson.source_root(), 'build-aux', 'ego', 'mkzip.sh')
+    join_paths(meson.project_source_root(), 'build-aux', 'ego', 'mkzip.sh')
   ]
 )
 
 
 # Post-Install script for distributions without the hooks
-if get_option('post_install')
-    meson.add_install_script(
-      env_util.path(),
-      'GSCHEMADIR=' + gschemadir,
-      join_paths(meson.source_root(), 'build-aux', 'meson', 'post-install.sh')
-    )
-endif
+gnome.post_install(
+  glib_compile_schemas: true,
+  gtk_update_icon_cache: true,
+  update_desktop_database: true,
+)
 
 
 # Extension Source
@@ -105,15 +97,15 @@ eslint = find_program('eslint', required: false)
 
 if eslint.found()
   test('ESLint (Source)', eslint,
-    args: join_paths(meson.source_root(), 'src'),
+    args: join_paths(meson.project_source_root(), 'src'),
     suite: 'lint',
   )
   test('ESLint (Installed Tests)', eslint,
-    args: join_paths(meson.source_root(), 'installed-tests'),
+    args: join_paths(meson.project_source_root(), 'installed-tests'),
     suite: 'lint',
   )
   test('ESLint (WebExtension)', eslint,
-    args: join_paths(meson.source_root(), 'webextension'),
+    args: join_paths(meson.project_source_root(), 'webextension'),
     suite: 'lint',
   )
 endif
--- a/.github/workflows/gnome-3-36.yml
+++ b/.github/workflows/gnome-3-36.yml
@@ -18,7 +18,6 @@ jobs:
           meson --prefix=/usr \
                 --libdir=lib/ \
                 -Dgnome_shell_libdir=/usr/lib64/ \
-                -Dpost_install=true \
                 _build
 
       - name: Uninstalled Tests
--- a/.github/workflows/gnome-3-38.yml
+++ b/.github/workflows/gnome-3-38.yml
@@ -18,7 +18,6 @@ jobs:
           meson --prefix=/usr \
                 --libdir=lib/ \
                 -Dgnome_shell_libdir=/usr/lib64/ \
-                -Dpost_install=true \
                 _build
 
       - name: Uninstalled Tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     container:
-      image: gsconnect/gsconnect-ci:latest
+      image: ghcr.io/gsconnect/gsconnect-ci:main
     steps:
       - uses: actions/checkout@v2
       - name: Prepare
@@ -21,7 +21,6 @@ jobs:
           meson --prefix=/usr \
                 --libdir=lib/ \
                 -Dgnome_shell_libdir=/usr/lib64/ \
-                -Dpost_install=true \
                 _build
 
       - name: Uninstalled Tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,7 +107,6 @@ $ meson --prefix=/usr \
         --libdir=lib/ \
         -Dgnome_shell_libdir=/usr/lib64/ \
         -Dfirewalld=true \
-        -Dpost_install=true \
         _build
 $ ninja -C _build install
 ```
--- a/build-aux/meson/post-install.sh
+++ b/build-aux/meson/post-install.sh
@@ -1,5 +1,0 @@
-#!/bin/sh
-
-GSCHEMA_DIR="${DESTDIR}/${GSCHEMADIR}"
-
-glib-compile-schemas "${GSCHEMA_DIR}"
--- a/data/meson.build
+++ b/data/meson.build
@@ -63,7 +63,11 @@ dbus = dependency('dbus-1', required: false)
 if get_option('session_bus_services_dir') != ''
   dbus_dir = get_option('session_bus_services_dir')
 elif dbus.found()
-  dbus_dir = dbus.get_pkgconfig_variable('session_bus_services_dir')
+  dbus_dir = dbus.get_variable(
+    pkgconfig: 'session_bus_services_dir',
+    internal: 'session_bus_services_dir',
+    default_value: join_paths(datadir, 'dbus-1', 'services'),
+  )
 else
   dbus_dir = join_paths(datadir, 'dbus-1', 'services')
 endif
--- a/installed-tests/meson.build
+++ b/installed-tests/meson.build
@@ -27,8 +27,8 @@ installed_tests_prepared = custom_target(
   build_by_default: true,
            command: [
              env_util,
-             'MESON_BUILD_ROOT=' + meson.build_root(),
-             'MESON_SOURCE_ROOT=' + meson.source_root(),
+             'MESON_BUILD_ROOT=' + meson.project_build_root(),
+             'MESON_SOURCE_ROOT=' + meson.project_source_root(),
              'G_TEST_BUILDDIR=' + installed_tests_builddir,
              'G_TEST_SRCDIR=' + installed_tests_srcdir,
              join_paths(installed_tests_srcdir, 'prepare-tests.sh')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,27 +1,11 @@
 # See https://github.com/GSConnect/gnome-shell-extension-gsconnect/wiki/Packaging
 
-# Run meson/post-install.sh (glib-compile-schemas)
-option(
-  'post_install',
-  type: 'boolean',
-  value: false,
-  description: 'Run meson/post-install.sh (eg. glib-compile-schemas)'
-)
-
 # GNOME Shell LIBDIR
 option(
   'gnome_shell_libdir',
   type: 'string',
   value: '',
   description: 'LIBDIR for GNOME Shell (eg. $LIBDIR/gnome-shell/Gvc-1.0.typelib)'
-)
-
-# GSettings schema directory
-option(
-  'gsettings_schemadir',
-  type: 'string',
-  value: '',
-  description: 'GSettings Schema directory'
 )
 
 # DBus service file
--- a/src/service/core.js
+++ b/src/service/core.js
@@ -83,14 +83,14 @@ var Packet = class Packet {
     /**
      * Update the packet from a dictionary or string of JSON
      *
-     * @param {Object|string} source - Source data
+     * @param {Object|string} data - Source data
      */
-    update(source) {
+    update(data) {
         try {
             if (typeof data === 'string')
-                Object.assign(this, JSON.parse(source));
+                Object.assign(this, JSON.parse(data));
             else
-                Object.assign(this, source);
+                Object.assign(this, data);
         } catch (e) {
             throw Error(`Malformed data: ${e.message}`);
         }


meson.build
```diff
@@ -89,7 +89,7 @@ if get_option('post_install')
     meson.add_install_script(
       env_util.path(),
       'GSCHEMADIR=' + gschemadir,
-      join_paths(meson.source_root(), 'build-aux', 'meson', 'post-install.sh')
+      join_paths(meson.project_source_root(), 'build-aux', 'meson', 'post-install.sh')
```
eli-schwartz, line 92: If you bump just a bit further to 0.57.0 you can drop this entirely, and use https://mesonbuild.com/Gnome-module.html#gnomepost_install

andyholmes: Ah that's cool, I didn't know about that function!

I think we need the option to set the GSchema directory [for NixOS](https://github.com/NixOS/nixpkgs/blob/18a07645e5428709a48e617fd52dc8aedcc85e4e/pkgs/desktops/gnome/extensions/gsconnect/default.nix#L63), unless there's a different way to do that in NixOS.

eli-schwartz: I don't really know how Nix works, but what are they doing for the entire rest of the gome ecosystem, which do in fact install to the *only* gnome-approved location (overridable via XDG_DATA_DIRS, not by specifying random paths)?

eli-schwartz: The Nix manual shows this, which says it sets things up via XDG_DATA_DIRS for you: https://nixos.org/manual/nixpkgs/stable/#ssec-gnome-hooks

ferdnyc: The Nix manual _also_ shows [this](https://nixos.org/manual/nixpkgs/stable/#ssec-gnome-common-issues), though:

> #### 17.12.4.4. I am packaging a project that cannot be wrapped, like a library or GNOME Shell extension.
> You can rely on applications depending on the library setting the necessary environment variables but that is often easy to miss. Instead we recommend to patch the paths in the source code whenever possible. Here are some examples:
> 
> * [Replacing a GI_TYPELIB_PATH in GNOME Shell extension](https://github.com/NixOS/nixpkgs/blob/7bb8f05f12ca3cff9da72b56caa2f7472d5732bc/pkgs/desktops/gnome-3/core/gnome-shell-extensions/default.nix#L21-L24) – we are using substituteAll to include the path to a typelib into a patch.
> 
> * The following examples are hardcoding GSettings schema paths. To get the schema paths we use the functions
> 
>     * glib.getSchemaPath Takes a nix package attribute as an argument.
> 
>     * glib.makeSchemaPath Takes a package output like $out and a derivation name. You should use this if the schemas you need to hardcode are in the same derivation.
> 
>     [Hard-coding GSettings schema path in Vala plug-in (dynamically loaded library)](https://github.com/NixOS/nixpkgs/blob/7bb8f05f12ca3cff9da72b56caa2f7472d5732bc/pkgs/desktops/pantheon/apps/elementary-files/default.nix#L78-L86) – here, substituteAll cannot be used since the schema comes from the same package preventing us from pass its path to the function, probably due to a [Nix bug](https://github.com/NixOS/nix/issues/1846).
> 
>     [Hard-coding GSettings schema path in C library](https://github.com/NixOS/nixpkgs/blob/29c120c065d03b000224872251bed93932d42412/pkgs/development/libraries/glib-networking/default.nix#L31-L34) – nothing special other than using [Coccinelle patch](https://github.com/NixOS/nixpkgs/pull/67957#issuecomment-527717467) to generate the patch itself.

ferdnyc: (Configuring `XDG_DATA_DIRS` per-package only works for installed applications, it can't be set per-Extension because they all get launched in the GNOME Shell process environment.)

jtojnar: Yeah, unfortunately the Meson’s `gnome.post_install` will not work for us. Ideally it would accept `gsettings_schema_dirs` option, added that to my TODO list.

But I recognize that relying on projects to compile the schemas is a bit of a hack – the only reason we have not gotten to doing it ourselves is that relying on the project’s build system works pretty well in practice – so I would not mind if GSconnect switched to `gnome.post_install`. We can always compile the schemas in the custom location ourselves for the packages that cannot do that and it is not a huge deal. The important thing is there is a single place where we can replace the path where GSconnect looks for the schemas so that it can find it after we move them – be it an option, or a constant definition in the source code we can patch.

eli-schwartz: Quite frankly I don't understand why absolutely any of this is necessary. Doesn't Nix already have dedicated per-package datadirs for exactly this reason? If building XDG_DATA_DIRS based on that isn't good enough, then considering this extension already hardcodes a gschema from_file (??) why isn't configuring it based on the datadir sufficient?

Why does there need to be another option in addition to the datadir?

andyholmes: I *think* the only two places this matters is [`config.js.in`](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/main/data/config.js.in) and [`metadata.json.in`](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/main/data/metadata.json.in) where we replace the `@GSETTINGS_SCHEMA_DIR@` placeholder, which should only come from the variable `gschemadir` in `/meson.build`.

IIRC `-Dgsettings_schemadir` was only for NixOS, so if we remove that we could just have a line like:

```meson
gschemadir = join_paths(datadir, 'glib-2.0', 'schemas')
```

I guess that would be the easiest to patch?

eli-schwartz: > Nix has a mode of operation where it creates so called “environment” by combining the trees of multiple packages into a single tree using symlinks. If we used the standard GSettings schema path, `gschemas.compiled` files from different packages would conflict.

I don't understand how this is related.

You have multiple prefixes, multiple datadirs, and multiple gschema directories. The trees of multiple packages will automatically conflict on gschemas.compiled, whether that directory is called:
- `{hash-of-gsconnect}/share/glib-2.0/schemas/` or
- `{hash-of-gsconnect}/share/gsettings-schemas/gsconnect/glib-2.0/schemas`

Elsewhere, you build a list of XDG_DATA_DIRS to collect various directories as inputs to a program, and create a wrapper program that sets the variable before running the program. Presumably, this would also be how you handle
> so called "environment[s]"

But failing that, assuming that it *is* hardcoded... why is it not sufficient to hardcode `{hash-of-gsconnect}/share/glib-2.0/schemas/` inside the program, such that the schema is loaded directly from there without looking anywhere other than inside the prefix in question?

eli-schwartz: But yes, if you have a mode where you allow overlaying directory trees from multiple applications using a shared cache file, you should be taking that into account. What do you do for mime packages? *.desktop files?

andyholmes: > I _think_ the only two places this matters is [`config.js.in`](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/main/data/config.js.in) and [`metadata.json.in`](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/main/data/metadata.json.in) where we replace the `@GSETTINGS_SCHEMA_DIR@` placeholder, which should only come from the variable `gschemadir` in `/meson.build`.
> 
> IIRC `-Dgsettings_schemadir` was only for NixOS, so if we remove that we could just have a line like:
> 
> ```meson
> gschemadir = join_paths(datadir, 'glib-2.0', 'schemas')
> ```
> 
> I guess that would be the easiest to patch?

Okay, this has been pushed but is waiting on GSConnect/gsconnect-ci#9

jtojnar: > You have multiple prefixes, multiple datadirs, and multiple gschema directories. The trees of multiple packages will automatically conflict on gschemas.compiled, whether that directory is called:

Not necessarily. The path is different for each package – so unless one tries to have multiple variants of the same package in the environment, there won’t be conflicts:

```
/nix/store/x6qf6457yxbcjbmgriyzy06clc39dh6a-geary-40.0/share/gsettings-schemas/geary-40.0/glib-2.0/schemas/gschemas.compiled
/nix/store/m2d0yjh7xmv2ccbadqhv79gv4ckr5isy-evince-42.2/share/gsettings-schemas/evince-42.2/glib-2.0/schemas/gschemas.compiled
```

> But yes, if you have a mode where you allow overlaying directory trees from multiple applications using a shared cache file, you should be taking that into account. What do you do for mime packages? *.desktop files?

Caches are generally not necessary for programs to work so packages do not include them and we have hooks that will remove them if they are generated. Instead, NixOS is responsible for generating them in the system environment – Nix’s and Nixpkgs’ `buildEnv` functions are too low level for that.

eli-schwartz: And this is still completely ignoring the fact that I suggested, you do not need to set up data dirs at all. If the software ***hardcodes*** `/nix/store/x6qf6457yxbcjbmgriyzy06clc39dh6a-geary-40.0/share/glib-2.0/schemas/` then it will just point there regardless of whether you set up an environment merger and regardless of the value of XDG_DATA_DIRS. So I really don't understand the problem that necessitates adding new options to all software.

jtojnar: > That's also why Meson doesn't run them if DESTDIR is set :) because no other distro runs them in the build environment either.

Right, I am aware. We do not usually set `DESTDIR` because we just have no need for it – the package build can only write to the Nix store paths it owns. The side effect of projects compiling their schemas was a just nice bonus, allowing us to not deal with the issue for the time being.

> But then you're not overlaying them together! You have to manually set XDG_DATA_DIRS to use these totally third-party (as far as glib is concerned) schema directories.

The individual packages usually already know the path to their datadir (either it is set by a wrapper or hardcoded using a patch). But the overlaying is **not** intended for the individual packages (for the most part, though sometimes, the mechanism is used for packages that use plug-ins, then there might need to be extra wrappers) – the purpose is to have single entry point we can symlink to a global path (e.g. `/run/current-system/sw/` on NixOS) whose subdirectories can be added to system-wide variables like `PATH` or `XDG_DATA_DIRS`. The symlink then can be swapped without having change the environment variables and restart everything.

> You cannot have it both ways. Either the goal is to merge the tree and fake out glib by telling glib that "yes, /glib-2.0/schemas/ is right here, it has all the files", and multiple packages conflict on the gschemas, or, you rely on XDG_DATA_DIRS to tell glib where to find schemas, in which case you do not need to merge the tree at all, just set the environment variable pointing at the store versions

Since the merged trees is not used to find the schemas, it is irrelevant for this.

> and exclude that directory from the merger (or map it specially).

That would be preferred but, as I mentioned, the component responsible for merging the trees is very low-level and does not care about tree semantics. Ideally we would introduce content-aware merge function but since we do not have it yet, we are stuck with this ancient hack. I have it on my TODO list but the separation of concerns between Nix, Nixpkgs and NixOS makes it somewhat difficult to decide what project should take care of that.

> You are creating extra work for yourself, and extra work for downstream developers, by designing an environment merger with brittle install layout requirements, that fails BOTH at being what Nix tries to be, AND at being developer-friendly.

Yes, it is extra work for us but until we have a proper fix there is not much else we can do. Yes, it is potentially brittle and we have to deal with issues occasionally but it somehow works for the most part. There should not really be any extra work for upstreams as long as they follow the best practice of [not repeating](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) installation path constants. I should not have added the option but I opened https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/197 long time ago, when I have not yet had a clear concept of roles of upstream and downstream.

andyholmes: > I should not have added the option but I opened https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/197 long time ago, when I have not yet had a clear concept of roles of upstream and downstream.

To be fair, neither did I :) Anyways, having a hacky script kicking around bothers me more than a meson switch, so I'm glad to get rid of it!

ferdnyc:
Meanwhile, and while we're primping and pruning, I happened to notice in the Nix packaging that they're applying this patch to the `meson.build` for `installed-tests`: 

[installed-tests-path.patch](https://github.com/NixOS/nixpkgs/blob/staging-next/pkgs/desktops/gnome/extensions/gsconnect/installed-tests-path.patch)
```diff
diff --git a/installed-tests/meson.build b/installed-tests/meson.buildJan Tojnar, 2 years ago: • gnomeExtensions.gsconnect: 41 → 43
index c7eff2fb..ef4f6052 100644
--- a/installed-tests/meson.build
+++ b/installed-tests/meson.build
@@ -1,5 +1,5 @@
-installed_tests_execdir = join_paths(libexecdir, 'installed-tests', meson.project_name())
-installed_tests_metadir = join_paths(datadir, 'installed-tests', meson.project_name())
+installed_tests_execdir = join_paths(get_option('installed_test_prefix'), 'libexec', 'installed-tests', meson.project_name())
+installed_tests_metadir = join_paths(get_option('installed_test_prefix'), 'share', 'installed-tests', meson.project_name())
 
 installed_tests_srcdir = meson.current_source_dir()
 installed_tests_builddir = meson.current_build_dir()
diff --git a/meson_options.txt b/meson_options.txt
index 8912e052..ca6ee5eb 100644
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -116,6 +116,13 @@ option(
   description: 'Native Messaging Host directory for Mozilla'
 )
 
+option(
+  'installed_test_prefix',
+  type: 'string',
+  value: '',
+  description: 'Prefix for installed tests'
+)
+
 option(
   'installed_tests',
   type: 'boolean',
```

...Will the changes here obviate the need for that patch? (I don't _think_ so, but I didn't look too deeply.)

If not, it certainly seems like the sort of thing that could be directly incorporated here up the stream, as it were. The `installed_test_prefix` option could easily be wired up to have no effect unless it's set. That way it won't affect anyone else, and Nix will have one less patch to carry.

Unless there's something I'm overlooking?

meson.build
eli-schwartz: Alright, so based on discussion it seems that the only real issue Nix has is that the un-optimized environment builder doesn't know how to exclude specific files, leading to a conflict when multiple store prefixes try to overlay the same file. The actual package itself does not care, and the software works if you don't create merged environments out of it.

So that's something which Nix already intends to fix, apparently, and when it's fixed then gnome-shell extensions on Nix can perfectly happily use `datadir/glib-2.0/schemas` too.

Whose responsibility it is to work around that can be debated, naturally.

For my part, I conclude that Meson probably does not need new functionality for `gnome.post_install()`.

jtojnar:
> Meanwhile, and while we're primping and pruning, I happened to notice in the Nix packaging that they're applying this patch to the `meson.build` for `installed-tests`:

Yeah we need to do that because we want to move the installed tests into a separate package to reduce the package size and since unlike regular distros each package has different prefix, we cannot just move them without updating path references.

Unfortunately, I do not think this is suitable for upstreaming since the patch hardcodes `share` and `libexec` directories. It might be more feasible with two separate options but I am not sure it would be worth it given that it will be likely only useful to Nixpkgs (I do not think Guix or Gobo are doing installed tests), and it is quite small patch to maintain downstream.

ferdnyc:
@jtojnar 

> Unfortunately, I do not think this is suitable for upstreaming, 

Not as-is, no. I was thinking something like this:

```meson
if get_option('installed_tests_prefix')
  installed_tests_execdir = join_paths(
    get_option('installed_tests_prefix'), 'libexec', 'installed-tests', meson.project_name())
  installed_tests_metadir = join_paths(
    get_option('installed_tests_prefix'), 'share', 'installed-tests', meson.project_name())
else
  installed_tests_execdir = join_paths(
    libexecdir, 'installed-tests', meson.project_name())
  installed_tests_metadir = join_paths(
    datadir, 'installed-tests', meson.project_name())
endif
```

> since the patch hardcodes `share` and `libexec` directories. 

Yeah, which unfortunately it probably HAS to do. While it's likely that `libexecdir` and `datadir` are relative (to `prefix`), and it would _probably_ work to just add something like this:
```meson
installed_tests_execdir = join_paths(
    get_option('installed_tests_prefix'), libexecdir, 'installed_tests', meson.project_name())
installed_tests_metadir = join_paths(
    get_option('installed_tests_prefix'), datadir, 'installed_tests', meson.project_name())
```
...I ultimately decided against that because of one of Meson's (IMHO) [more questionable features](https://mesonbuild.com/Reference-manual_functions.html#join_paths):

> If any one of the individual segments is an absolute path, all segments before it are dropped.

Which means that, while `libexecdir` is _probably_ going to be relative most of the time, if it's ever NOT then it would screw us and end up installing to the wrong directory.

(In fact, the docs even warn about not relying on `libexecdir` et al being relative in all cases, and always explicitly using `get_option('prefix')` when constructing paths that aren't being passed to an `install_dir:`.)

@jtojnar 
> It might be more feasible with two separate options but I am not sure it would be worth it given that it will be likely only useful to Nixpkgs (I do not think Guix or Gobo are doing installed tests), and it is quite small patch to maintain downstream.

\*shrug\* Oh, sure, but it's just one more thing to have to keep in sync with our upstream changes, or for us to have to be careful not to break when we _make_ those changes... I mean, there's a reason Fedora packaging policy [strongly advises pushing all local packages upstream, whenever possible](https://docs.fedoraproject.org/en-US/packaging-guidelines/PatchUpstreamStatus/#_all_patches_should_have_an_upstream_bug_link_or_comment). (Though they do make allowances for Fedora-specific patches that will forever stay local, and I can certainly understand the position that this is a Nix-specific patch in the same vein.)

ferdnyc:
> (In fact, the docs even warn about not relying on `libexecdir` et al being relative in all cases, and always explicitly using `get_option('prefix')` when constructing paths that aren't being passed to an `install_dir:`.)

LOL! I just noticed that Nix is <s>blamed</s> [cited as one of the reasons](https://mesonbuild.com/Release-notes-for-0-62-0.html#all-directory-options-now-support-paths-outside-of-prefix) why `get_option()` is now allowed to return absolute paths in addition to relative ones, for install destinations. (The other reason is Autotools / CMake parity, but that's boring.)

jtojnar:
> Which means that, while `libexecdir` is _probably_ going to be relative most of the time, if it's ever NOT then it would screw us and end up installing to the wrong directory.

Yeah, we explicitly use absolute paths for all path options and I am ~~at fault~~ responsible for https://github.com/mesonbuild/meson/pull/9903. Though `libexecdir` and `datadir` will most often be under `prefix` so Meson will relativize them. But I would still not rely on them being relative since it could lead to silent weirdness. Well, in this case, it would just make `installed_tests_prefix` a no-op but that would still be confusing or frustrating.

> Oh, sure, but it's just one more thing to have to keep in sync with our upstream changes, or for us to have to be careful not to break when we _make_ those changes...

I mean, yeah, in the end it is upstream’s choice if they want to have downstream-specific patches in-tree but doing so does not reduce the need for caution, the only way to do that would be adding Nix into CI. And that would introduce a dependency on Nix people. Sometimes it can be worth it since it can reveal incorrect assumptions but unlike C, Gjs/JavaScript is pretty self-contained so IMO it likely would not reveal much.

ferdnyc:
I mean, personally I wouldn't object to running Nix-based CI, alongside our current monolithically-Fedora pool. Some OS diversity is rarely a _bad_ thing, at **worst** it's slightly redundant and fails to introduce any meaningful variation into the environment. (But we already know that's not the case, with Nix.)

I mean, we have [a whole separate repo](https://github.com/gsconnect/gsconnect-ci) **just** to maintain the GSConnect CI "platform". Limiting it to a single OS almost seems like a waste. (In fact, it probably makes sense to continue this further over there, at gsconnect/gsconnect-ci#11.)

eli-schwartz:
> ...I ultimately decided against that because of one of Meson's (IMHO) [more questionable features](https://mesonbuild.com/Reference-manual_functions.html#join_paths):
> 
> > If any one of the individual segments is an absolute path, all segments before it are dropped.

If you really want to avoid dropping that, you "could" use string concat, rather than path-based joining (which BTW is modeled after python's os.path.join).

```
installed_tests_execdir = get_option('installed_tests_prefix') + '/' + libexecdir + '/installed_tests/' + meson.project_name()
```

Of course you may then end up with:

```
/nix/store/fdgdfgdfgdfgdfgd-foo-installed-tests-1.0/nix/store/mlohg98kkf-foo-1.0/libexec
```

where I think maybe you actually wanted the word "libexec"... unfortunately, the problem here isn't actually that join_paths will drop segments before an absolute path, the problem here is that absolute paths are allowed.

I suppose you can always add the options:

- installed_test_exedir
- installed_test_metadir

and if they are defined to `''`then use the existing calculations. Nix can just pass two options instead of one.

jtojnar:
My dream solution would be making subprojects so easy to use that projects would naturally split themselves into subprojects and express dependencies using `dependency` (disregarding the source similarly to CMake’s `find_package`). But that would require pretty massive changes – dropping the limitation on only having subprojects in `subprojects/` directory and other things to make it more ergonomic (e.g. sharing version number). Then we could just build and install each project separately, passing it custom installation directories.

Though I have not really thought it out in details how it would affect e.g. passing options.

Issue #1406, opened 2022-07-04
Slideshow / Presentation Remote not working


bevsxyz:
> ***ATTENTION**: GSConnect only supports the latest, stable version of GNOME. We are no longer accepting bug reports for previous versions.*

**Describe the bug**

I have been trying to control remotely control slides from the KDEConnect android app. The forward and backward buttons register but do not register any kind of action with Google slides in firefox or Impress for that matter. The pointer button doesn't seem to even respond at all.

**Steps To Reproduce:**

1. Pair with the kdeconnect app
2. Open any slideshow
3. From the slideshow control panel
4. Try to change slides

**Expected behavior**

Easily change the slides

**Screenshots**

If applicable, add screenshots to help explain your problem. You can drag-and-drop or cut-and-paste images directly into this edit window, to include them in your report.

**Support Log**

Please generate a support log ([Instructions](../wiki/Help#generate-support-log)) and paste any messages related to this issue between the two ``` lines below.

```

GSConnect: 50 (user)
GJS:       17200
Session:   wayland
OS:        Fedora Linux 36 (Workstation Edition)
--------------------------------------------------------------------------------
Jul 04 13:01:16 systemd[9800]: dbus-:1.3-org.gnome.TextEditor@10.service: Consumed 4.408s CPU time.
Jul 04 13:01:26 gnome-shell[12614]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
Jul 04 13:01:32 gsd-media-keys[16114]: Unsupported modifier, resource creation failed.
Jul 04 13:01:32 gsd-media-keys[16114]: XXX: resource creation failed
Jul 04 13:01:39 gnome-shell[12614]: g_signal_handler_disconnect: assertion 'handler_id > 0' failed
Jul 04 13:01:39 gnome-shell[12614]: g_signal_handler_disconnect: assertion 'handler_id > 0' failed
Jul 04 13:01:39 gnome-shell[12614]: g_signal_handler_disconnect: assertion 'handler_id > 0' failed
Jul 04 13:01:39 gnome-shell[12614]: g_signal_handler_disconnect: assertion 'handler_id > 0' failed
Jul 04 13:01:39 gnome-shell[12614]: g_signal_handler_disconnect: assertion 'handler_id > 0' failed
Jul 04 13:01:39 gnome-shell[12614]: g_signal_handler_disconnect: assertion 'handler_id > 0' failed
[...]

PR #1409, opened 2022-07-06
Fixes issue #1406

adam-husain:
### Brief Description
Fixes issue #1406. The Slideshow Remote feature has a bug where Previous Slide and Next Slide does not work. The pull request fixes this issue.

### Explanation
The feature did not work because the buttons would send a `kdeconnect.mousepad.request` with the `specialKey = 8` (for Previous Slide) and a `specialKey = 9` (for Next Slide). However, these `specialKey` values are key-mapped to `Page Up` and `Page Down` input which has no effect on a presentation slideshow.
The fix converts the `specialKey = 8` and `9` values to `4` and `6` respectively, which corresponds to Left and Right arrow key inputs.

### DIFF
--- a/src/service/plugins/mousepad.js
+++ b/src/service/plugins/mousepad.js
@@ -191,7 +191,14 @@ var Plugin = GObject.registerClass({
 
                 // Special key (eg. non-printable ASCII)
                 } else if (input.specialKey && KeyMap.has(input.specialKey)) {
-                    keysym = KeyMap.get(input.specialKey);
+                    // Convert PageUp and PageDown to Left and Right respectively
+                    // This is done to fix Slideshow Remote
+                    if (input.specialKey === 8)
+                        keysym = KeyMap.get(4)
+                    else if (input.specialKey === 9)
+                        keysym = KeyMap.get(6)
+                    else
+                        keysym = KeyMap.get(input.specialKey);
                     this._input.pressKeys(keysym, modifiers);
                     this._sendEcho(input);
                 }
--- a/src/service/plugins/mousepad.js
+++ b/src/service/plugins/mousepad.js
@@ -194,9 +194,9 @@ var Plugin = GObject.registerClass({
                     // Convert PageUp and PageDown to Left and Right respectively
                     // This is done to fix Slideshow Remote
                     if (input.specialKey === 8)
-                        keysym = KeyMap.get(4)
+                        keysym = KeyMap.get(4);
                     else if (input.specialKey === 9)
-                        keysym = KeyMap.get(6)
+                        keysym = KeyMap.get(6);
                     else
                         keysym = KeyMap.get(input.specialKey);
                     this._input.pressKeys(keysym, modifiers);
--- a/src/service/plugins/mousepad.js
+++ b/src/service/plugins/mousepad.js
@@ -191,14 +191,7 @@ var Plugin = GObject.registerClass({
 
                 // Special key (eg. non-printable ASCII)
                 } else if (input.specialKey && KeyMap.has(input.specialKey)) {
-                    // Convert PageUp and PageDown to Left and Right respectively
-                    // This is done to fix Slideshow Remote
-                    if (input.specialKey === 8)
-                        keysym = KeyMap.get(4);
-                    else if (input.specialKey === 9)
-                        keysym = KeyMap.get(6);
-                    else
-                        keysym = KeyMap.get(input.specialKey);
+                    keysym = KeyMap.get(parseInt(input.specialKey));
                     this._input.pressKeys(keysym, modifiers);
                     this._sendEcho(input);
                 }


adam-husain:
I found the actual problem. `Page Up` and `Page Down` SHOULD work for apps like Google Slides, LibreOffice and OnlyOffice. However, the current version does not handle the input correctly.

In technical terms: The input number `8` and `9` for Previous and Next Slide respectively are mapped as a string value. This causes a `null` to be returned as the `KeyMap` only accepts integers as key. By using the `parseInt()` function, it is converted to an int key.

I have tested and committed the fixed version.

bevsxyz:
I think this is the better way to handle it. Could possibly resolve unrelated issues in the future. I will test it out later. Thanks @adam-husain for the quick response ❤️

src/service/plugins/mousepad.js
```diff
@@ -191,7 +191,7 @@ var Plugin = GObject.registerClass({
 
                 // Special key (eg. non-printable ASCII)
                 } else if (input.specialKey && KeyMap.has(input.specialKey)) {
-                    keysym = KeyMap.get(input.specialKey);
+                    keysym = KeyMap.get(parseInt(input.specialKey));
                     this._input.pressKeys(keysym, modifiers);
```
andyholmes, line 195: I think the error is calling `this._input.pressKeys()` instead of `this._input.pressreleaseKeysym()`. If `specialKey` were a string, it would never pass the call to `KeyMap.has(input.specialKey)`.

adam-husain: This is weird. Installing the extension from [Gnome Extensions](https://extensions.gnome.org/extension/1319/gsconnect), the Slideshow does not work. However, after cloning the repo and building it works.
I assumed my fix worked because after building, it just works for some reason.

Is this an issue just on my computer or is this common?

bevsxyz: Then the issue has been resolved in a commit after the last version release?

Definitely not the mousepad.js file under this PR. The last commit for that was in 2021.

bevsxyz: Oh shit! The gnome extension website version is still at v50, whereas the Github release page shows v53. It needs updating. I'll try to use the Github release and comment.

bevsxyz: The v53 release does not work. But as @adam-husain said building from the repo works. The full screen and exit presentation buttons work as well.

andyholmes: This might have been inadvertently fixed by f2798d7, but the above error should still be fixed I think.

bevsxyz: Okay I think from what i understand the current function simulates key press but not a release, and you're suggesting a function which would simulate a key press followed by a release. Is that how it works?

andyholmes: Well right now it calls `this._input.pressKeys()` which takes a list of characters, a string. It should be using `this,_input.pressreleaseKeysym()` which takes a single key symbol (a code for the symbol of a character) and presses/releases it. Since `KeyMap` is a map from KDE Connect codes to keysyms, I think that's what it should have always been.

ferdnyc: That sounds more correct to me, as well. The idea that you'd need to use `parseInt()` on the `KeyMap.get()` **after** passing a `KeyMap.has()` check _without_ it just doesn't make sense, in terms of program logic.

(If the same `parseInt()` were added to the `KeyMap.has()` check, then I could potentially get behind that change. But without it? Something's clearly off.)

bevsxyz: Looking at the definition, https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/281a9c9c4cf156374bf651f0956ecfc04ba08771/src/service/components/input.js#L214

I agree.

```js
keysym = KeyMap.get(input.specialKey);
this._input.pressreleaseKeysym(keysym);
```
That would be the change required then.

ferdnyc: Hrm. Which we can't do as a suggestion. But assuming @adam-husain has enabled upstream committer pushes to the PR branch, we can do it on the command line. Since it sounds like everyone is in agreement (yes?), I'll attempt to push a commit with that change now.

ferdnyc: Didn't work, so I opened #1418 from my fork. I included @adam-husain's original commits in the branch, so attribution is preserved.

ferdnyc: So, that was the wrong definition of `.pressreleaseKeysym()`. The actual one for the _input_ component is here:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/281a9c9c4cf156374bf651f0956ecfc04ba08771/src/service/components/input.js#L214-L222

...But it suffers from the same problem as the other: It doesn't take a `modifiers` argument, so the modifier keys are potentially getting lost if we use `pressreleaseKeysym()`. (See #1418.)

ferdnyc: (Er, sorry, duh. You had the right definition all along, @bevsxyz . Sourcegraph led me down a wrong path.)

andyholmes:
Did we decide this issue was a false positive? I see there was a [comment](https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1418#discussion_r922162320) stating this could be closed after v54 + testing.

ferdnyc:
> Did we decide this issue was a false positive?

It's been like 3000 years, but looking at #1418, seems like it. Not exactly a false positive, but as you said (there), this appears to have been a response to a real bug... but one that was already fixed. It just hasn't made it to a release, yet.

I stand by my statement that adding `parseInt()` to keymap.get()`, but NOT `keymap.has()`, just doesn't make any sense. So I don't think this should be merged. And I don't _believe_ any change is necessary, because the bug @adam-husain was addressing has already been fixed elsewhere (on the `main` branch).

andyholmes:
Closing then, let me know if this was in error and I'll re-open.

PR #1418, opened 2022-07-14
Fix mousepad special key presses (replaces #1409)

ferdnyc:
An update to PR #1409 by @adam-husain, with the changes discussed in that PR's review comments.

Fixes: #1406

### DIFF
--- a/src/service/plugins/mousepad.js
+++ b/src/service/plugins/mousepad.js
@@ -191,7 +191,14 @@ var Plugin = GObject.registerClass({
 
                 // Special key (eg. non-printable ASCII)
                 } else if (input.specialKey && KeyMap.has(input.specialKey)) {
-                    keysym = KeyMap.get(input.specialKey);
+                    // Convert PageUp and PageDown to Left and Right respectively
+                    // This is done to fix Slideshow Remote
+                    if (input.specialKey === 8)
+                        keysym = KeyMap.get(4)
+                    else if (input.specialKey === 9)
+                        keysym = KeyMap.get(6)
+                    else
+                        keysym = KeyMap.get(input.specialKey);
                     this._input.pressKeys(keysym, modifiers);
                     this._sendEcho(input);
                 }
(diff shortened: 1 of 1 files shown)


ferdnyc:
Hmm. The change breaks the mousepad unit tests, but that may be a case of the test needing to be updated. I'll take a look.

ferdnyc:
Yeah, the mousepad test is expecting `pressKeys()` to be called, specifically. Not a great test, really, but it can be updated to expect `pressreleaseKeysym()` instead.

src/service/plugins/mousepad.js
```diff
@@ -192,7 +192,7 @@ var Plugin = GObject.registerClass({
                 // Special key (eg. non-printable ASCII)
                 } else if (input.specialKey && KeyMap.has(input.specialKey)) {
                     keysym = KeyMap.get(input.specialKey);
-                    this._input.pressKeys(keysym, modifiers);
+                    this._input.pressreleaseKeysym(keysym);
                     this._sendEcho(input);
```
ferdnyc, line 196: @andyholmes @bevsxyz

Ooh, wait, what about the `modifiers`? They're not part of the `keysym` value (or, I don't see how they would be), so  it seems like they're getting lost here.

ferdnyc: The input's `._pressKeysym()` (as opposed to `.pressReleaseKeysym()` takes a `modifiers` argument, and that's what `.pressKey()` (not `.pressKeys()`) calls internally when passed an integer first argument (assumed to be a `keysym`):

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/281a9c9c4cf156374bf651f0956ecfc04ba08771/src/service/components/atspi.js#L290-L294

andyholmes: Ah, I was wrong after all. It looks like that's what `Input.pressKeys()` already does:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/281a9c9c4cf156374bf651f0956ecfc04ba08771/src/service/components/input.js#L608-L621

I guess this is really just a bug that was already fixed, but not included in a release :shrug:

ferdnyc: @andyholmes 
Sounds like it. On that assumption, I'm going to close this one. Sounds like we can probably close #1409 as well, tho I suppose there's no harm in waiting until v54 is out there and we can confirm that this is already fixed already.

PR #1443, opened 2022-08-25
Change shebang for compatibility/portability

fabianhjr:
This was causing a very cryptic issue on my side and this way should be more portable:

https://stackoverflow.com/q/21612980/539209


andyholmes:
Thanks!

PR #1450, opened 2022-08-31
readme: fix 'people' link

matttbe:
The previous link was not working for me:

  https://github.com/GSConnect/gnome-shell-extension-gsconnect/people

I guess the correct one is:

  https://github.com/orgs/GSConnect/people


daniellandau:
Hmm, when logged in the original works for me but on a private window it doesn't.

PR #1462, opened 2022-09-14
Change the iPhone icon to a more generic icon

Talkys:
As most users will be using an Android phone, makes more sense to not put an iPhone icon on the symbolic svg. Removing the notch is a little detail that makes the icon more generic.


andyholmes:
Thanks for the contribution! This icon is actually from the `phonelink-*` icon set in the [GNOME Icon Development Kit](https://gitlab.gnome.org/Teams/Design/icon-development-kit), so I'm not terribly inclined to change it, especially after so many years.

I'll mark as needing review and let other maintainers voice their opinion though.

daniellandau:
I didn't know where it was from, but it looks exactly like my Android phone so I don't see a need to change it.

PR #1463, opened 2022-09-16
Remote Keyboard: Fixed referencing text area from plugin.

WSID:
Previously, plugin side (`plugins/mousepad.js`) referenced text area as `text`, while it is `entry` in ui side (`ui/mousepad.js`).

Changed its name in plugin side, to `entry`.


andyholmes:
LGTM, thanks!

PR #1469, opened 2022-09-24
Added Remote Input Touchpad for mouse.

WSID:
Touch pad that can be controlled by mouse, to remote control mouse on paired device.

Touchpad 2022-09-24 07-55-12

-   Touchpad
    -   Drag area to move mouse.
    -   Long press to down left mouse button while moving mouse. (Performs drag and drop)
    -   Short click to perform click of left, right, or middle.
-   Buttons
    -   Button press to perform click.
-   Dialog
    -   Scroll to perform scroll. Scroll events does not reach touchpad. So it need to be taken from dialog.

## Remarks

1.  Icons are added (f8cdcc2)

    A icon `mouse-with-smaller-scrollwheel-symbolic` is added from [Gnome Icon Library](https://apps.gnome.org/app/org.gnome.design.IconLibrary/).

    Two icons are modified from this.

    -   `mouse-left-button-symbolic`
    -   `mouse-right-button-symbolic`
    
2.  Translatable string changed (6751de3)

    Action name and dialog title is changed, which are marked as translatable. And one translatable string is added. Though `.pot` is not updated accordingly.

## Further works that need.

-   Handling touch inputs on touch pad.
    I have not tested it with touch input. But I don't think this would be hard.

-   Touch pad preference.
    Some adjustable options...
    -   Speed and Acceleration
    -   Long press delay
    -   2 finger, 3 finger click actions (if touch handling is added)

-   Locking pointer device, to send all input to paired device.
    Grab a pointer device, and send all of event from the device.
    -   On Wayland session, moving mouse pointer from `Gdk` not work. Might need workaround if this is needed.


data/ui/mousepad-input-dialog.ui
```diff
@@ -73,9 +73,81 @@
...
+            <property name="orientation">vertical</property>
+            <child>
+              <object class="GtkEventBox" id="touchpad-eventbox">
+                <child>
+                  <object class="GtkLabel">
+                    <property name="can_focus">False</property>
+                    <property name="vexpand">True</property>
+                    <property name="halign">fill</property>
+                    <property name="justify">center</property>
+                    <property name="label" translatable="True">Touchpad.
```
andyholmes, line 90:
```suggestion
                    <property name="label" translatable="yes">Touchpad.
```

andyholmes:
This looks really good, thanks! I'll be happy to merge, just two things:

1.  apply the suggested changed
2. run `ninja -C _build gnome-shell-extension-gsconnect-pot` and push the relevant changes

data/ui/mousepad-input-dialog.ui
WSID: It is addressed via 9de72c5.

andyholmes:
LGTM, thanks!

PR #1475, opened 2022-10-01
Port to GNOME 43

MacTavishAO:


MacTavishAO:
I think there's still room for an improvement here. I'll push my last changes tomorrow.

MacTavishAO:
@daniellandau I think it's ready now. It can be merged.

daniellandau:
Cool, I'll try to read through it and test ASAP, probably within thr next few days (other maintainers feel free to do it too if you have time before me).

MacTavishAO:
Should I fix CI errors and push a new commit?

steve-todorov:
Hey, 

I've just tried this PR and can confirm it works fine for me so far.
My setup is Ubuntu 22.10 (daily) with Gnome 43.0.
Could not find any issues so far :)

MacTavishAO:
> This is a very cosmetic issue - if you re-open the menu it shows everything properly.
Just reproduced this on my system and can confirm this bug exist. Deleting lines below should fix this issue. https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/9c4e7345e909a644d61d4216be0b75195d6dd43e/src/extension.js#L170-L230

Deleting these lines should be fine as we don't have a submenu anymore. Just to be safe could you also test this and report back before I push a fix?

steve-todorov:
Thanks for the update. I replaced the `extension.js` with the [36c59cd](https://github.com/GSConnect/gnome-shell-extension-gsconnect/commit/36c59cd9a8235ba2f7d070afa7615c27ec768336) revision, restarted gnome, but the issue still persists.
image

MacTavishAO:
> Thanks for the update. I replaced the `extension.js` with the [36c59cd](https://github.com/GSConnect/gnome-shell-extension-gsconnect/commit/36c59cd9a8235ba2f7d070afa7615c27ec768336) revision, restarted gnome, but the issue still persists. image

It's working for me as can be seen in video below.

https://user-images.githubusercontent.com/25158549/193570782-ca9ccc50-773d-47de-a68c-c045e1b4a98b.mp4

Could you try installing this [zip](https://github.com/MacTavishAO/gnome-shell-extension-gsconnect/releases/tag/v56.3) and try again? Maybe it's a one time thing.

steve-todorov:
I've completely uninstalled the extension and installed the one from your zip, but the result is still the same. Might be some corner-case issue that I'm hitting. Anyway - it's not a big deal to me. :)

MacTavishAO:
> I've completely uninstalled the extension and installed the one from your zip, but the result is still the same. Might be some corner-case issue that I'm hitting. Anyway - it's not a big deal to me. :)

You need to restart gnome shell after changing any extension files. More info [here](https://github.com/GSConnect/gnome-shell-extension-gsconnect/wiki/Help#update-error).

nerd:
during testing of this PR on my system, GSConnect is disabled after wake from suspend, requiring me to re-enable it via the quick toggle. this is different behaviour from previous versions of the extension. i am using Fedora with GNOME 43.0

MacTavishAO:
> during testing of this PR on my system, GSConnect is disabled after wake from suspend, requiring me to re-enable it via the quick toggle. this is different behaviour from previous versions of the extension. i am using Fedora with GNOME 43.0

Try removing the line below;
https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/36c59cd9a8235ba2f7d070afa7615c27ec768336/src/extension.js#L359

For some weird reason the service wouldn't kill itself when the extension is disabled via extension manager. I don't know if it's an intended behaviour (can someone test this on gnome 42?) but I'm open for any suggestions on how to fix it. I'm not experienced at all about these stuff so any help is appreciated.

andyholmes:
> For some weird reason, the service wouldn't kill itself when the extension is disabled via extension manager. I don't know if it's an intended behaviour (can someone test this on gnome 42?) but I'm open for any suggestions on how to fix it.

This is expected behaviour for GSConnect; `disable()` gets called when the extension is uninstalled, disabled manually or the screen turns off. So the service will keep running regardless of the Shell extension, while the the `enabled` GSetting just tells the Shell extension whether it should start/stop the service.

src/extension.js
```diff
@@ -31,10 +32,18 @@ Extension.getIcon = Utils.getIcon;
...
-        super._init();
+        super._init({
+            label: 'GSConnect',
+            iconName: 'org.gnome.Shell.Extensions.GSConnect-symbolic',
+            toggleMode: true,
+        });
+
+        // Set QuickMenuToggle header.
+        this.menu.setHeader('org.gnome.Shell.Extensions.GSConnect-symbolic', 'GSConnect',
+            'Sync between your devices');
```
andyholmes, line 46:
```suggestion
            _('Sync between your devices'));
```

This string should be marked translatable.

meson.build
```diff
@@ -1,6 +1,6 @@
 project('gsconnect',
   'c',
-  version: '53',
+  version: '56',
```
andyholmes, line 3: This should be removed before merge.

src/extension.js
```diff
@@ -31,10 +32,18 @@ Extension.getIcon = Utils.getIcon;
  */
 const ServiceIndicator = GObject.registerClass({
     GTypeName: 'GSConnectServiceIndicator',
-}, class ServiceIndicator extends PanelMenu.SystemIndicator {
+}, class ServiceIndicator extends QuickSettings.QuickMenuToggle {
```
andyholmes, line 35:
```suggestion
}, class ServiceToggle extends QuickSettings.QuickMenuToggle {
```

Renaming for consistency (ServiceIndicator, ServiceToggle).

src/extension.js
```diff
@@ -168,7 +153,7 @@ const ServiceIndicator = GObject.registerClass({
         const panelMode = this.settings.get_boolean('show-indicators');
 
         // Hide status indicator if in Panel mode or no devices are available
-        this._indicator.visible = (!panelMode && available.length);
+        featureIndicator._indicator.visible = (!panelMode && available.length);
```
andyholmes, line 156:
```suggestion
        serviceIndicator._indicator.visible = (!panelMode && available.length);
```

Renaming for consistency (ServiceIndicator, ServiceToggle).

src/extension.js
```diff
@@ -410,17 +319,55 @@ const ServiceIndicator = GObject.registerClass({
         this.settings.run_dispose();
 
         // Destroy the PanelMenu.SystemIndicator actors
-        this._item.destroy();
         this.menu.destroy();
 
-        delete AggregateMenu._gsconnect;
+        delete QuickSettingsMenu._gsconnect;
```
andyholmes, line 324: This line can probably just be removed, it was a weird old compatibility thing with other extensions.

src/extension.js
```diff
@@ -410,17 +319,55 @@ const ServiceIndicator = GObject.registerClass({
...
         super.destroy();
     }
 });
 
+const FeatureIndicator = GObject.registerClass(
+class FeatureIndicator extends QuickSettings.SystemIndicator {
```
andyholmes, line 330:
```suggestion
const ServiceIndicator = GObject.registerClass(
class ServiceIndicator extends QuickSettings.SystemIndicator {
```

Renaming for consistency (ServiceIndicator, ServiceToggle).

src/extension.js
```diff
@@ -410,17 +319,55 @@ const ServiceIndicator = GObject.registerClass({
...
+        });
+
+        // Create the icon for the indicator
+        this._indicator = this._addIndicator();
+        this._indicator.icon_name = 'org.gnome.Shell.Extensions.GSConnect-symbolic';
+        // Hide the indicator by default
+        this._indicator.visible = false;
+
+        // Create the toggle menu and associate it with the indicator
+        this.quickSettingsItems.push(new ServiceIndicator());
```
andyholmes, line 350:
```suggestion
        this.quickSettingsItems.push(new ServiceToggle());
```

Renaming for consistency (see other suggestions).

src/extension.js
```diff
@@ -410,17 +319,55 @@ const ServiceIndicator = GObject.registerClass({
...
+        // Destroy indicator
+        // Workaround: Disabling and enabling extension multiple times increases
+        // padding between new indicator and other indicators. Hide indicator before
+        // destroying to workaround this bug.
+        this._indicator.visible = false;
+        this._indicator.destroy();
+    }
+});
 
+var featureIndicator = null;
```
andyholmes, line 370:
```suggestion
var serviceIndicator = null;
```

Renaming for consistency (ServiceIndicator, ServiceToggle).

src/extension.js
```diff
@@ -443,13 +390,13 @@ function init() {
 
 
 function enable() {
-    serviceIndicator = new ServiceIndicator();
+    featureIndicator = new FeatureIndicator();
```
andyholmes, line 393:
```suggestion
    serviceIndicator = new ServiceIndicator();
```

Renaming for consistency (ServiceIndicator, ServiceToggle).

src/extension.js
```diff
@@ -443,13 +390,13 @@ function init() {
...
     Notification.patchGtkNotificationSources();
 }
 
 
 function disable() {
-    serviceIndicator.destroy();
-    serviceIndicator = null;
+    featureIndicator.destroy();
+    featureIndicator = null;
```
andyholmes, line 400:
```suggestion
    serviceIndicator.destroy();
    serviceIndicator = null;
```

Renaming for consistency (ServiceIndicator, ServiceToggle).

src/extension.js
```diff
@@ -372,17 +289,9 @@ const ServiceIndicator = GObject.registerClass({
 
     async _onServiceChanged(service, pspec) {
```
andyholmes, line 290: This function should probably be used to set the [`St.Button:checked` property](https://gjs-docs.gnome.org/st10/st.button#property-checked) of the toggle, e.g. something as simple as:

```js
this.checked = service.active;
```

Then add an override to `ServiceToggle` for [`St.Button.clicked()`](https://gjs-docs.gnome.org/st10/st.button#vfunc-clicked):

```js
vfunc_clicked(clickedButton) {
    if (this.service.active)
        this.service.stop().catch(logError);
    else
        this.service.start().catch(logError);
}
```

Thus when the button is clicked it just does the right thing, and the button state changes with the actual service state.

src/extension.js
```diff
@@ -410,17 +319,55 @@ const ServiceIndicator = GObject.registerClass({
...
-var serviceIndicator = null;
+        // GSettings
+        this.settings = new Gio.Settings({
+            settings_schema: Config.GSCHEMA.lookup(
+                'org.gnome.Shell.Extensions.GSConnect',
+                null
+            ),
+            path: '/org/gnome/shell/extensions/gsconnect/',
+        });
+
```
andyholmes, line 342:
```suggestion
```

Can remove this.

src/extension.js
```diff
@@ -410,17 +319,55 @@ const ServiceIndicator = GObject.registerClass({
...
+        this.quickSettingsItems.push(new ServiceIndicator());
+
+        // Add the indicator to the panel and the toggle to the menu
+        QuickSettingsMenu._indicators.insert_child_at_index(this, 0);
+        QuickSettingsMenu._addItems(this.quickSettingsItems);
+    }
+
+    destroy() {
+        // Set enabled state to false to kill the service on destroy
+        this.settings.set_boolean('enabled', false);
```
andyholmes, line 359:
```suggestion
```

Can remove this, too.

src/extension.js
```diff
@@ -410,17 +319,55 @@ const ServiceIndicator = GObject.registerClass({
...
+        this.quickSettingsItems.forEach(item => item.destroy());
+        // Destroy indicator
+        // Workaround: Disabling and enabling extension multiple times increases
+        // padding between new indicator and other indicators. Hide indicator before
+        // destroying to workaround this bug.
+        this._indicator.visible = false;
+        this._indicator.destroy();
```
andyholmes, line 366: This should be properly fixed before merge.

MacTavishAO:
> This is expected behaviour for GSConnect; `disable()` gets called when the extension is uninstalled, disabled manually or the screen turns off. So the service will keep running regardless of the Shell extension, while the the `enabled` GSetting just tells the Shell extension whether it should start/stop the service.

Shouldn't the service at least disconnect from the phone? Because even after disabling the extension, I can still control my computer using the app on my phone.

andyholmes (changes requested):
Looks really good, here is a first-pass review.

src/extension.js
MacTavishAO: Inserting "super.destroy();" after destroying the indicator itself fixes it.

src/extension.js
MacTavishAO: Going through with this will cause the extension to not follow the last state on every boot.
Since we won't be updating the enabled key, value registered in gsettings will be considered the default one. For example; if enabled is set to true, it will start the service and enable the toggle regardless of the fact that it may have been toggled off by the user in previous boot. I purpose skipping this change for convenience.

MacTavishAO: So, @andyholmes what should I do?

MacTavishAO: > This function should probably be used to set the [`St.Button:checked` property](https://gjs-docs.gnome.org/st10/st.button#property-checked) of the toggle, e.g. something as simple as:
> 
> ```js
> this.checked = service.active;
> ```
> 
> Then add an override to `ServiceToggle` for [`St.Button.clicked()`](https://gjs-docs.gnome.org/st10/st.button#vfunc-clicked):
> 
> ```js
> vfunc_clicked(clickedButton) {
>     if (this.service.active)
>         this.service.stop().catch(logError);
>     else
>         this.service.start().catch(logError);
> }
> ```
> 
> Thus when the button is clicked it just does the right thing, and the button state changes with the actual service state.

We already have _onEnabledChanged doing vfunc_clicked's job for us. I purpose partially implementing this.

andyholmes:
I guess you can change that behaviour if you want, but be aware this means whenever the screen is locked or off, GSConnect will stop running.

andyholmes:
LGTM, if @daniellandau gives the thumbs up we can merge.

andyholmes (approved):

daniellandau:
I won't have time to focus on this before next weekend, but I trust you so 👍.

andyholmes:
Okay thanks for the work on this!

steve-todorov:
Awesome news guys! Is there any ETA on the release?

daniellandau:
Next weekend unless e.g. @sonnyp has time to do a github release and I can just upload to e.g.o.

ferdnyc:
@MacTavishAO 

> Shouldn't the service at least disconnect from the phone? Because even after disabling the extension, I can still control my computer using the app on my phone and mpris is still exposed.

The problem with this is, AIUI there's absolutely no way in GNOME Shell to determine whether an extension is being disabled transiently (due to something like a screen lock activation), disabled explicitly by the user, or even disabled permanently because it's being uninstalled. All three of those actions result in just the same `disable()` call.

I'm actually coming around to the idea that this represents a design flaw in the extension protocol. There's another PR somewhere in the repo where the submitter tried to add icon and metadata cleanup on uninstall of the extension, and we hit the same problem — even if the user was merely disabling (but *not* uninstalling) the extension, the cleanup would trigger, which is not at all what you want. But there's no way to tell the difference.

The extension API commits a bit _too_ completely, IMHO, to this implicit theory of extensions that seems to go something like, "extensions are transient Shell enhancements that don't install anything, don't have any dependencies, and don't have anything useful to do except when they're activated by the Shell". Maybe that was the _**initial idea**_, but the practical reality is that the extension system has been expanded **far** beyond that narrow definition. As a result, it would be extremely helpful if there were some API support for differentiating the various "disable" state transitions, so that they can be handled in different ways.

cwells:
> I guess you can change that behaviour if you want, but be aware this means whenever the screen is locked or off, GSConnect will stop running.

Is there a way to fix this? I only use gsconnect to unlock my screen, so this change makes it entirely useless to me.

MacTavishAO:
> Is there a way to fix this? I only use gsconnect to unlock my screen, so this change makes it entirely useless to me.

See this: https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1516#issuecomment-1320915291

Issue #1366, opened 2022-05-26
Add iOS App Store link into data/ui/preferences-window.ui


p-bo:
See https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1351

ijiki16:
@andyholmes we can close this too.

PR #1365, opened 2022-05-26
Add iOS into OS list in README

p-bo:
See https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1351

PR #1489, opened 2022-10-18
add iOS App Store link into preferences-window.ui

ijiki16:
image

close issues #1366

### DIFF
--- a/data/ui/preferences-window.ui
+++ b/data/ui/preferences-window.ui
@@ -575,6 +575,44 @@
                               </packing>
                             </child>
                             <child>
+                              <object class="GtkLabel" id="kdeconnect-ios">
+                                <property name="visible">True</property>
+                                <property name="can_focus">False</property>
+                                <property name="halign">end</property>
+                                <property name="label">iOS:</property>
+                                <accessibility>
+                                  <relation type="label-for" target="kdeconnect-appstore"/>
+                                </accessibility>
+                              </object>
+                              <packing>
+                                <property name="left_attach">0</property>
+                                <property name="top_attach">2</property>
+                              </packing>
+                            </child>
+                            <child>
+                              <object class="GtkLinkButton" id="kdeconnect-appstore">
+                                <property name="label">App Store</property>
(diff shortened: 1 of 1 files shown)


data/ui/preferences-window.ui
```diff
@@ -574,6 +574,44 @@
                                 <property name="top_attach">1</property>
                               </packing>
                             </child>
+                            <child>
+                              <object class="GtkLabel" id="kdeconnect-ios">
+                                <property name="visible">True</property>
+                                <property name="can_focus">False</property>
+                                <property name="halign">end</property>
+                                <property name="label">iOS:</property>
```
ferdnyc, line 582: Hmm. I wonder if, just to be super clear (and avoid reports correcting us later), we should label this "iOS/iPadOS" or "iOS & iPadOS"? (I assume it works on iPads as well, no?)

I'm usually perfectly happy to call them both "iOS", but it seems like there's always someone turning up to point out, "iPads don't run iOS!" (Really I'm fine either way, though.)

ferdnyc (approved):

data/ui/preferences-window.ui
ijiki16: I thought about the same thing, but everywhere in other parts of code & docs, it says `iOS`:  https://github.com/GSConnect/gnome-shell-extension-gsconnect/commit/9410f9a6b760dc0b10d515d3df76e5d541b7b036 , https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1365/commits/efae3b6746a628e71a31c3c3f52ce9d5df424bc2. 
 I can easily change the label, but I think these parts must change too.

ferdnyc: @ijiki16 Mmm, fair point. Let's stick with this, then. If we decide to change it, we can change them all at once in a different PR.

ferdnyc:
Are we too late for v54 on this one, @andyholmes ?

andyholmes:
> Are we too late for v54 on this one, @andyholmes ?

Yeah, already released :/

PR #1491, opened 2022-10-19
Show transfered file in folder as default action of notification (replaces #1039)

### LINKED ISSUE #1009: Click on "Transfer successful" notification should open file or it's location with file selected

ferdnyc:
This PR finishes up the changes for PR #1039, with most of the work done by @HaMF over there. I just tacked on a commit with some review suggestions. Differences from the original PR:

1. The function is renamed `showPathInFolder`, as suggested
2. In addition to being the default click action for the notification, it's also made the "Open Folder" button action. (Now renamed to "Show File Location". I decided that "Open File Location" and "Open File" right next to each other was too confusing, so I didn't go with that.)
3. The D-Bus call made to launch the file manager now passes a result callback that finishes the request, or logs an error if encountered.

Closes #1009 
<s>Closes #1039  (←- I don't think this will work, but...)</s> It does not work, probably-sensibly.

### DIFF
--- a/src/service/device.js
+++ b/src/service/device.js
@@ -486,6 +486,13 @@ var Device = GObject.registerClass({
         openPath.connect('activate', this.openPath);
         this.add_action(openPath);
 
+        const showPathInFolder = new Gio.SimpleAction({
+            name: 'showPathInFolder',
+            parameter_type: new GLib.VariantType('s'),
+        });
+        showPathInFolder.connect('activate', this.showPathInFolder);
+        this.add_action(showPathInFolder);
+
         // Preference helpers
         const clearCache = new Gio.SimpleAction({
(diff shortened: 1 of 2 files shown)


ferdnyc:
I tested a version of this code rebased onto `v50`, and it worked with no troubles. I'm not running GNOME 43, so I can't test a build based off `v54`, but the code in question should be unaffected by the version bump.

ferdnyc:
Feedback from the original PR that remains unaddressed:

* Only [this review comment](https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1039#discussion_r654756388) from @sonnyp, for the reasons I outline in my reply there.

andyholmes (approved):
LGTM

PR #1510, opened 2022-11-10
Update nautilus-gsconnect.py

busterbeam:
Just made it shorter and more readable.  but found out that this script hasn't got any access to the notification system to provide a progress bar.  how would one modify `send_files` to allow how complete is the process.  I have an habit to transmit >100MB files between my computer and phone which takes a few minutes it would be nice to check my notification panel and see how far it is to finishing and maybe an ETA that is updated realtime

### DIFF
--- a/nautilus-extension/nautilus-gsconnect.py
+++ b/nautilus-extension/nautilus-gsconnect.py
@@ -2,85 +2,75 @@
 nautilus-gsconnect.py - A Nautilus extension for sending files via GSConnect.
 
 A great deal of credit and appreciation is owed to the indicator-kdeconnect
-developers for the sister Python script 'kdeconnect-send-nautilus.py':
+developers for the sister Python script "kdeconnect-send-nautilus.py":
 
 https://github.com/Bajoja/indicator-kdeconnect/blob/master/data/extensions/kdeconnect-send-nautilus.py
 """
 
-import gettext
-import os.path
-import sys
+from gettext import translation
+from os.path import join, exists
+from sys import argv
 
 import gi
-gi.require_version('Gio', '2.0')
-gi.require_version('GLib', '2.0')
-gi.require_version('GObject', '2.0')
-from gi.repository import Gio, GLib, GObject
+gi.require_version("Gio", "2.0")
+gi.require_version("GLib", "2.0")
+gi.require_version("GObject", "2.0")
+from gi.repository.Gio import (
+    DBusProxy, DBusProxyFlags, BusType, DBusActionGroup, DBusCallFlags)
+from gi.repository.GLib import get_user_data_dir, Variant
+from gi.repository.GObject import Object
 
 
 # Host application detection
 #
-# Nemo seems to reliably identify itself as 'nemo' in argv[0], so we
+# Nemo seems to reliably identify itself as "nemo" in argv[0], so we
 # can test for that. Nautilus detection is less reliable, so don't try.
 # See https://github.com/linuxmint/nemo-extensions/issues/330
-if "nemo" in sys.argv[0].lower():
+if "nemo" in argv[0].lower():
     # Host runtime is nemo-python
-    gi.require_version('Nemo', '3.0')
+    gi.require_version("Nemo", "3.0")
     from gi.repository import Nemo as FileManager
 else:
     # Otherwise, just assume it's nautilus-python
     from gi.repository import Nautilus as FileManager
 
 
-SERVICE_NAME = 'org.gnome.Shell.Extensions.GSConnect'
-SERVICE_PATH = '/org/gnome/Shell/Extensions/GSConnect'
+SERVICE_NAME = "org.gnome.Shell.Extensions.GSConnect"
+SERVICE_PATH = "/org/gnome/Shell/Extensions/GSConnect"
 
 # Init gettext translations
-LOCALE_DIR = os.path.join(GLib.get_user_data_dir(),
-                          'gnome-shell', 'extensions',
-                          'gsconnect@andyholmes.github.io', 'locale')
+LOCALE_DIR = join(
+    get_user_data_dir(), "gnome-shell", "extensions",
+    "gsconnect@andyholmes.github.io", "locale")
 
-if not os.path.exists(LOCALE_DIR):
+if not exists(LOCALE_DIR):
     LOCALE_DIR = None
 
 try:
-    i18n = gettext.translation(SERVICE_NAME,
-                               localedir=LOCALE_DIR)
+    i18n = translation(SERVICE_NAME, localedir=LOCALE_DIR)
     _ = i18n.gettext
-
-except (IOError, OSError) as e:
-    print('GSConnect: {0}'.format(e.strerror))
-    i18n = gettext.translation(SERVICE_NAME,
-                               localedir=LOCALE_DIR,
-                               fallback=True)
+except (IOError, OSError) as error:
+    print(f"GSConnect: {error.strerror}")
+    i18n = translation(SERVICE_NAME, localedir=LOCALE_DIR, fallback=True)
     _ = i18n.gettext
 
 
-class GSConnectShareExtension(GObject.Object, FileManager.MenuProvider):
+class GSConnectShareExtension(Object, FileManager.MenuProvider):
     """A context menu for sending files via GSConnect."""
 
     def __init__(self):
         """Initialize the DBus ObjectManager"""
-
-        GObject.Object.__init__(self)
-
-        self.devices = {}
-
-        Gio.DBusProxy.new_for_bus(Gio.BusType.SESSION,
-                                  Gio.DBusProxyFlags.DO_NOT_AUTO_START,
-                                  None,
-                                  SERVICE_NAME,
-                                  SERVICE_PATH,
-                                  'org.freedesktop.DBus.ObjectManager',
-                                  None,
-                                  self._init_async,
-                                  None)
+        Object.__init__(self)
+        self.devices = dict()
+        DBusProxy.new_for_bus(
+            BusType.SESSION, DBusProxyFlags.DO_NOT_AUTO_START, None,
+            SERVICE_NAME, SERVICE_PATH, "org.freedesktop.DBus.ObjectManager",
+            None, self._init_async, None)
 
     def _init_async(self, proxy, res, user_data):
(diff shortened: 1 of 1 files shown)


nautilus-extension/nautilus-gsconnect.py
```diff
@@ -2,85 +2,75 @@
...
-                                  'org.freedesktop.DBus.ObjectManager',
-                                  None,
-                                  self._init_async,
-                                  None)
+        Object.__init__(self)
+        self.devices = dict()
+        DBusProxy.new_for_bus(
+            BusType.SESSION, DBusProxyFlags.DO_NOT_AUTO_START, None,
+            SERVICE_NAME, SERVICE_PATH, "org.freedesktop.DBus.ObjectManager",
+            None, self._init_async, None)
```
ferdnyc, line 68: No. Hard no from me. For the wrapping alone.

This packed style is not "more readable", in fact it's far easier to lose track of which argument is which. I'd fully support wrapping the args under the call (that aligned-crazy-far-to-the-right style drives me nuts, TBH), but the one-arg-per-line formatting is extremely valuable given how many there are. It also makes diffs that change any parameters much cleaner.

nautilus-extension/nautilus-gsconnect.py
```diff
@@ -2,85 +2,75 @@
...
-SERVICE_NAME = 'org.gnome.Shell.Extensions.GSConnect'
-SERVICE_PATH = '/org/gnome/Shell/Extensions/GSConnect'
+SERVICE_NAME = "org.gnome.Shell.Extensions.GSConnect"
+SERVICE_PATH = "/org/gnome/Shell/Extensions/GSConnect"
 
 # Init gettext translations
-LOCALE_DIR = os.path.join(GLib.get_user_data_dir(),
-                          'gnome-shell', 'extensions',
-                          'gsconnect@andyholmes.github.io', 'locale')
+LOCALE_DIR = join(
```
ferdnyc, line 42: Here again: `os.path.join` is immediately clear to future readers of this code. But reading your version, what's the definition of `join()`? Is it a local function? Is it `str.join()`? `Thread.join()`? Importing generically-named functions from specific packages, when there are at least a half-dozen possible modules it could have come from, hurts code readability far more than the extra characters required for parent module chains.

nautilus-extension/nautilus-gsconnect.py
```diff
@@ -2,85 +2,75 @@
...
-gi.require_version('GLib', '2.0')
-gi.require_version('GObject', '2.0')
-from gi.repository import Gio, GLib, GObject
+gi.require_version("Gio", "2.0")
+gi.require_version("GLib", "2.0")
+gi.require_version("GObject", "2.0")
+from gi.repository.Gio import (
+    DBusProxy, DBusProxyFlags, BusType, DBusActionGroup, DBusCallFlags)
+from gi.repository.GLib import get_user_data_dir, Variant
+from gi.repository.GObject import Object
```
ferdnyc, line 21: Given that Python has an `object` builtin, this strikes me as a colossally terrible idea. "`Object`" alone does nothing to help future code readers understand that the object being discussed is a `GObject.Object`. Ditto `GLib.Variant` on the previous line.

nautilus-extension/nautilus-gsconnect.py
```diff
@@ -148,25 +128,24 @@ def get_file_items(self, *args):
...
         # Only accept regular files
         for uri in files:
-            if uri.get_uri_scheme() != 'file' or uri.is_directory():
-                return ()
+            if uri.get_uri_scheme() != "file" or uri.is_directory():
+                return tuple()
 
         # Enumerate capable devices
-        devices = []
+        devices = list()
```
ferdnyc, line 135: These changes baffle me, especially considering they seem to go in the opposite direction from all the other changes. Find me a person with more than 10 hours' familiarity with Python code who doesn't know that `[]` creates an empty list.

ferdnyc (changes requested):
I have severe reservations about a **lot** of these changes, and code that imports generic names like `Object` and `join` into the local namespace, detached from the parent module, I will oppose without compromise.

The only hope that might exist for _some_ of these to make it into the repo is if they're done as separate commits so we can pick and choose. This change, in addition to being objectionable from my POV, does far too much unrelated stuff to usefully review.

Standard advice is standard: One logical change per commit.

ferdnyc:
@busterbeam 

> found out that this script hasn't got any access to the notification system to provide a progress bar. how would one modify `send_files` to allow how complete is the process.

I'm not sure you would, Nautilus doesn't expose any of that API to scripts. The GSConnect `daemon.js` would probably be the right place to report transfer progress. Theoretically it could use a Shell notification.

Unfortunately, a big part of the reason **Nautilus itself** _doesn't_ use Shell notifications, but instead has its own UI for showing progress on file operations, is that Shell notifications are designed to be pretty static and they don't handle being updated in realtime very well.

But AFAIK the only code that can display Nautilus progress popups is what's compiled into Nautilus.

nautilus-extension/nautilus-gsconnect.py
ferdnyc: (An example of the cleaner-diffs issue: The third arg to `DBusProxy.new_for_bus()` is an optional `GDBusInterfaceInfo`. Let's say we need to start passing one of those in for some reason.

With the previous code, that diff looks like this:
```diff
@@ -4,7 +4,7 @@
 
         Gio.DBusProxy.new_for_bus(Gio.BusType.SESSION,
                                   Gio.DBusProxyFlags.DO_NOT_AUTO_START,
-                                  None,
+                                  dbus_interface_info,
                                   SERVICE_NAME,
                                   SERVICE_PATH,
                                   'org.freedesktop.DBus.ObjectManager',
```

The packed version? This happens:
```diff
@@ -16,10 +16,10 @@
         Object.__init__(self)
         self.devices = dict()
         DBusProxy.new_for_bus(
-            BusType.SESSION, DBusProxyFlags.DO_NOT_AUTO_START, None,
-            SERVICE_NAME, SERVICE_PATH, "org.freedesktop.DBus.ObjectManager",
-            None, self._init_async, None)
+            BusType.SESSION, DBusProxyFlags.DO_NOT_AUTO_START,
+            dbus_interface_info, SERVICE_NAME, SERVICE_PATH,
+            "org.freedesktop.DBus.ObjectManager", None,
+            self._init_async, None)
```

nautilus-extension/nautilus-gsconnect.py
busterbeam: You don't need to know what module a object or function comes from to know what it does.  And to check you can quickly copy the import header then

Help(function_name) to find everything you need to know about it.

Any good IDE will know what your doing as well

nautilus-extension/nautilus-gsconnect.py
busterbeam: Your the one stating all the way down I've made things unclear.  But here you are running for the hills as using names instead of literal empty syntax isn't clear.

busterbeam:
@ferdnyc, apologies that my suggested changes are not helpful.

> The GSConnect `daemon.js` would probably be the right place to report transfer progress.

Thank you for this I will chase this up

nautilus-extension/nautilus-gsconnect.py
ferdnyc: > You don't need to know what module a object or function comes from to know what it does.

Consider that a bit more, and I think you'll find that it's actually _**impossible**_ to know what an object or function does, unless you know its definition -- which requires knowing what module it's from.

Some IDEs can keep track of that for you, sure. Just like the interpreter keeps track of it in the code itself. But none of that changes the fact that **something** has to hold on to that information. If you don't know exactly what class is represented by `Object`, you know _nothing_ about it.

And when reading code in a non-interactive environment (like, say, a GitHub PR diff), there is no IDE hovering over your shoulder to remind you that `Object` is a `GObject.Object`.

nautilus-extension/nautilus-gsconnect.py
ferdnyc: > To be fair some people instead import it as joinpath.

At which point, you're saving a whopping 4 characters over `os.path.join`, and you have to remember this random other name for a standard library function. Where's the _benefit_?

> It doesn't get confused with str.join, because one is clearly a method while the other a function. 

That's a fair point that also covers `Thread.join()`, granted.

> And 9/10 times the variables past to join have the variable name "path" so yeah

True, but then you have to _analyze_ the entire line. The [human] parsing goes from, "Ok, I'm joining a path, and the components are..." (completely linear) to "OK, I'm joining something... these looks like path components... that must've been `os.path.join` back there..." (requires backtracking).

Even if it's not "unclear", it can be "clear[-ish]" while still being "less obvious", and that can still be a negative. If interpreting the less-obvious line might require an extra second or two of thought, once you multiply that by the number of times this code will be read by someone over its lifetime you're talking real numbers.

ferdnyc:
@busterbeam 

> @ferdnyc, apologies that my suggested changes are not helpful.

Well, I think you'll find that parachuting into a project and arbitrarily reformatting the code is almost never considered helpful. If everyone who visited the repo did that, it'd be a pile of 1000-line diffs that did nothing but shuffle things around to suit each person's tastes. How is that productive?

I'm actually glad you brought this up, because it was the push I needed to put together #1512 which will _standardize_ the formatting of the Python code the same way we long ago standardized our JS.

Style debates are always unproductive. Everyone has opinions and everyone's opinion is valid, but since opinions contradict and everyone's opinion can't win, it's usually better if _nobody's_ opinion matters.

Style debates in community code are also complicated by the fact that participants will have different goals and concerns. Project maintainers in open-source projects have to focus on long-term maintainability and code consistency first and foremost, perspectives which don't always seem to make sense to code authors. Norms like the one-argument-per-line format for long function calls arose less from implications for readers of the code, but because the impact on the readability of **_future_ diffs** is immense. And when reviewing project PRs, you end up spending all your time reading diffs of the code, not the code itself.

andyholmes:
Thanks for the contribution, but I'm going to close this one down. It hasn't seen any forward movement since it was opened, and I'm not convinced it fixes anything that needs fixing.

IIRC, the code style was inherited from the original script it was based on, which seems to be PEP-8. If there's some benefit or change in consensus about best-practices, it's best if it is accompanied by a well-know linter so the coding style doesn't just regress later.

PR #1512, opened 2022-11-12
Add Python linting for Nautilus extension

ferdnyc:
This PR adds optional use of both the `black` and `flake8` commands to lint the Python code in `nautilus-extension/nautilus-gsconnect.py`. 

- The test setup is extended as usual: If either command is available, a <s>new test suite 'lint-python' is created.</s> test is added to the 'lint' suite for each available tool.
- Black configuration is in the root '/pyproject.toml' file
- flake8 configuration is in the root '/.flake8' file

A second commit checks in the necessary changes for our current code to pass both of these checks.

`black --check --diff` mode is used for scanning. It will not reformat the code, but will complain if it doesn't like the formatting, and output a diff of what it _would_ change if not prevented from doing so. (That diff will be truncated to the last 100 lines by the test runner, though.)

We could consider, in the future, having a local CI job that would just automatically reformat the code with `black` and commit the changes, to completely relieve humans of the code-style burden. (That's probably easier/better done by a repo Action than from inside the CI container.) Speaking of the container, I'll open a PR in the `gsconnect-ci` repo to add Python, black, and flake8 to the Docker image so the tests can be run there.

Some configuration adjustments are made on each linter, though they're kept to a minimum and for the most part the scans rely on the tools' built-in defaults. Linter configuration or tuning can be a subjective or even arbitrary thing, and both selection of linter tools and details of how they're configured are certainly open for discussion or refinement.

Configs applied by this PR:

1. Black is configured to enforce 80-character lines instead of its arbitrary 88, simply because I am personally baffled why anyone would want to impose a length constraint, but then choose one that still leaves the code unable to fit in an 80-character terminal without wrapping.
2. flake8 is configured to ignore E402, "module level import not at top of file", because it is impossible to pass that check if your code uses `gi.require_version()` before importing modules. (The individual lines could be annotated with overrides for the rule if we wanted to keep it in general, though.

----

This PR was largely inspired by #1510 from @busterbeam, with thanks for the nudge to address this. It's also motivated by the fact that we have very strict ESlint checking of the JavaScript, which makes it seem inconsistent that the Python code is allowed to just do its own thing. While I don't agree with _arbitrarily_ reformatting code based on user preferences, I'm all for **standardizing** code based on clear rules and mutually-adopted guidelines.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+extend-ignore = E402
--- a/meson.build
+++ b/meson.build
@@ -118,7 +118,7 @@ if black.found()
       '--diff',
       join_paths(meson.project_source_root(), 'nautilus-extension')
     ],
-    suite: 'lint-python',
+    suite: 'lint',
   )
 endif
 
@@ -126,7 +126,7 @@ flake8 = find_program('flake8', required: false)
 if flake8.found()
   test('Python Flake8 (Nautilus Extension)', flake8,
     args: join_paths(meson.project_source_root(), 'nautilus-extension'),
-    suite: 'lint-python',
+    suite: 'lint',
   )
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -110,6 +110,26 @@ if eslint.found()
   )
 endif
 
+black = find_program('black', required: false)
+if black.found()
+  test('Python Black (Nautilus Extension)', black,
+    args: [
+      '--check',
+      '--diff',
+      join_paths(meson.project_source_root(), 'nautilus-extension')
+    ],
+    suite: 'lint-python',
+  )
+endif
+
+flake8 = find_program('flake8', required: false)
+if flake8.found()
+  test('Python Flake8 (Nautilus Extension)', flake8,
+    args: join_paths(meson.project_source_root(), 'nautilus-extension'),
+    suite: 'lint-python',
+  )
+endif
+
 
 subdir('data')
 subdir('installed-tests')
--- a/nautilus-extension/nautilus-gsconnect.py
+++ b/nautilus-extension/nautilus-gsconnect.py
@@ -12,9 +12,10 @@ import os.path
 import sys
 
 import gi
-gi.require_version('Gio', '2.0')
-gi.require_version('GLib', '2.0')
-gi.require_version('GObject', '2.0')
+
+gi.require_version("Gio", "2.0")
+gi.require_version("GLib", "2.0")
+gi.require_version("GObject", "2.0")
 from gi.repository import Gio, GLib, GObject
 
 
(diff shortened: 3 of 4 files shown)


ferdnyc:
GSConnect/gsconnect-ci#12 is the PR to update the CI image. It passes, but I'll wait to merge until people have had a chance to discuss any concerns about the choice of linters, their configs, etc. since that could affect the changes necessary there.

andyholmes (approved):

andyholmes:
A thumbs from me on both PR's :+1:

ferdnyc:
I moved the Python checks into the 'lint' suite, since it's already run by name in the CI Uninstalled tests. This way, an additional suite doesn't have to be specified just for the Python linting.

ferdnyc:
A'ighty, going to merge the CI image changes, so we can get a test-run of this PR with the tools enabled.

ferdnyc:
And unlike every piece of cashmere I've ever owned, we're lint-free:

```text
ninja: Entering directory `/__w/gnome-shell-extension-gsconnect/gnome-shell-extension-gsconnect/_build'
[1/1] Generating data/metainfo/org.gnome.Shell.Extensions.GSConnect.metainfo.xml with a custom command
1/8 gsconnect:lint / ESLint (Installed Tests)            OK              4.41s
2/8 gsconnect:lint / ESLint (Source)                     OK              5.39s
3/8 gsconnect:lint / Python Black (Nautilus Extension)   OK              0.15s
4/8 gsconnect:lint / ESLint (WebExtension)               OK              1.14s
5/8 gsconnect:data / Validate metainfo file              OK              0.01s
6/8 gsconnect:data / Validate desktop file               OK              0.01s
[7](https://github.com/GSConnect/gnome-shell-extension-gsconnect/actions/runs/3449585055/jobs/5762487475#step:5:8)/[8](https://github.com/GSConnect/gnome-shell-extension-gsconnect/actions/runs/3449585055/jobs/5762487475#step:5:9) gsconnect:data / Validate desktop file (Preferences) OK              0.01s
8/8 gsconnect:lint / Python Flake8 (Nautilus Extension)  OK              0.17s


Ok:                 8   
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0
```

ferdnyc:
Hmmm. We have an ISSUE_TEMPLATE, but not a PULL_REQUEST_TEMPLATE it looks like. I wonder if it's worth adding one, mentioning (among potential other things) that it's a good idea to run `black nautilus-extension/nautilus-gsconnect.py` from the root directory and commit the results, after changing the Python code? Might stave off a bunch of lint failures. (Heck, maybe we should just write a pre-commit hook.)

ferdnyc:
Setting the template question aside for another PR and merging this, if anyone wants to make adjustments feel free to submit 'em!

PR #1515, opened 2022-11-12
editorconfig: Follow moved files

ferdnyc:
Dumb little PR to update the .editorconfig file, so it looks for the `/data/webextension/` files in the correct place (and with the correct names).


ferdnyc:
Aand, dumb little merge of dumb little PR.

PR #1525, opened 2022-11-22
Replace the issue templates with forms

### LINKED ISSUE #1524: Switch issue template licenses to `.license` files or `.reuse/dep5` configs?
So, turns out the license comment in `.github/ISSUE_TEMPLATE/bug_report.md` or `.../feature-request-or-idea.md` _will_ be included in the prefilled textarea, when GitHub generates a new issue form from the corresponding template file. Not the end of the world, but it also isn't ideal.

Might be worth moving the license metadata out of those `.md` template files using one of the REUSE-supported methods. Just so that we keep things as uncluttered/non-confusing as possible, when users are filling out the template.

ferdnyc:
In addition to Markdown templates, GitHub supports a YAML syntax for defining issue _forms_ in a more structured manner. Issue forms can have multiple input fields; each field can have one of several distinct data types; fields can be required, blocking form submission until the user provides a value; each input field has an `id` attribute which can be targeted for more advanced, targeted issue searches; and forms natively support "decorative/instructional" text that is _part OF_ the form, but will not be submitted _WITH_ the form. (Separately, placeholder text is also supported for free-form text entry fields.)

To preview the submission workflow with the new forms, visit https://github.com/ferdnyc/gnome-shell-extension-gsconnect/issues/new/choose and select one of the two predefined report options provided.

(very indirectly...) Fixes #1524


andyholmes (approved):

ferdnyc:
@andyholmes I hope so... I hope they're actually usable. Part of me is sort of irrationally waiting for some other shoe to drop, to find out I've overlooked some down side or created some sort of obstacle for the users. _Hopefully_ not, but I guess time will tell.

In *theory* there are all sorts of possibilities for how we could up the smarts on these even further — like, generate (or at least update) a list of selections for the GSConnect and/or GNOME Shell version(s) based on the metadata for every new release, that sort of thing. But I figured, get them out there as a first step, and we can overcomplicate them later.

(I did find myself wishing the definition syntax had _some_ sort of facility for defining relationships between the fields — like, it'd be great if there was some way to define a text field that's disabled by default, and only takes input if a checkbox is clicked or a certain value is selected from a dropdown list, that sort of thing.

I know logically that there's no _need_ to do that, and that any field value that indicates "this other field is in use" would technically be redundant with there just _being any data_ in that second field... but it still feels like a potentially handy thing.

(Having the dedicated text field for screenshots only activate in response to an "I want to add screenshots" checkbox was probably the main thing I'd like to actually be able to do. The directions/description can be more involved if only the users who _need_ that field have to bother with it, for one thing.)

Maybe I'll even bounce that off GitHub. (Or more likely, go see what their response has been (if any), the many times other users have no doubt already made the same suggestion/request.)

andyholmes:
These look great to me. I think its a lot easier to jump into a form like this than chopping up blocks of commented text.

PR #1526, opened 2022-11-22
Generate installable extension packages as build artifacts

ferdnyc:
GitHub Actions build artifacts can only take one form: An automatically-created zip file. But the contents of that zip file don't have to be a single file — they can be anything, even the complete contents of an extension package directory.

This PR modifies the build's custom targets to add a new option, `make-pkgdir`. This target runs `build-aux/ego/mkzip.sh` as usual, but signals it to bail out just _before_ zipping up the extension package, so that the final `gsconnect@andyholmes.github.io` package directory is created and populated, but not compressed.

The CI's `upload-artifact` action job is then pointed at that directory as the source of the artifact zip file's contents.

By letting GitHub Actions do the zipping of the extension data, the final artifact will contain the exact contents of the extension package zip file, _rather than_ the packaged zip file itself. This makes the artifact directly installable as an extension distribution.


ferdnyc:
And, bingo-bango: Each CI build artifact can now be downloaded and directly installed just like a packaged extension, because it _is_ the packaged extension. No more fiddly outer, redundant `.zip` layers to unwrap.

https://github.com/GSConnect/gnome-shell-extension-gsconnect/actions/runs/3521330240

andyholmes:
I guess I should have checked at the time, but it appears this is resulting in the executable bit being lost on `daemon.js` and `gsconnect-preferences`. Is there a way around that, or is github doing that for security purposes?

ferdnyc:
@andyholmes 
Yeah, that came up a while ago; I'm sure it's intentional on GitHub's part.

IIRC, I added code to the extension to restore the bit on first launch. It should run right after (before?) the code that installs the Nautilus plugins and etc.

ferdnyc:
Yeah, `ensurePermissions()` in `shell/utils.js` is the code, it's the first thing called from `init()` in `extension.js`.

PR #1528, opened 2022-11-23
chore: add SPDX licensing for snippets and files missing attribution

andyholmes:


src/service/plugins/contacts.js
```diff
@@ -146,12 +146,15 @@ var Plugin = GObject.registerClass({
...
-     *
      * @param {string} input - The QUOTED-PRINTABLE string
      * @return {string} The decoded string
      */
     _decodeQuotedPrintable(input) {
+        /* SPDX-SnippetBegin
+         * SPDX-License-Identifier: MIT
+         * SPDX-SnippetCopyrightText: Mathias Bynens
+         * SPDX-SnippetComment: https://github.com/mathiasbynens/quoted-printable/blob/master/src/quoted-printable.js
+         */
```
ferdnyc, line 157: For instance, this file should probably cause the `reuse lint` check to fail, because the MIT license is not currently installed in `/LICENSES/`. The path to that would seem to be making the header something like,

```js
// SPDX-FileCopyrightText: GSConnect Developers https://github.com/GSConnect
//
// SPDX-License-Identifier: GPL-2.0-or-later
// 
// License for _decodeQuotedPrintable():
// SPDX-License-Identifier: MIT
// SPDX-SnippetCopyrightText: Mathias Bynens
// SPDX-SnippetComment: https://github.com/mathiasbynens/quoted-printable/blob/master/src/quoted-printable.js
```

...I think that could work... (Well, I just tested, and it definitely **does** work to make `reuse lint` fail until the MIT license file is downloaded into the repo.)

Whether we then still duplicate all of that info down here, or limit the local comment to the `SPDX-Snippet...` markers and a reference to the header. Either approach seems pretty reasonable to me.

ferdnyc (commented):
The reuse [FAQ entry](https://reuse.software/faq/#partial-license) on this is,

> ### Can I license only a part of a file as being under a different license?
> The short answer is that yes, you can, but no, there is no standard way for REUSE to recognize this. If you have a small segment of a file that is licensed differently, you should list that license under a separate SPDX-License-Identifier tag in the header.
> 
> You can use your own comments to specify which segment is separately licensed. For instance: “The class Foo is copied from project Bar and licensed under MIT. All changes are licensed under GPL-3.0-or-later.”
> 
> A possible way to circumvent the problem is to extract the segment from the file, and to keep it in its own file.

I think the header suggestion is reasonable -- _in addition_ to marking the sections themselves, any embedded alternate licenses should probably be included in the header as well.

src/service/plugins/contacts.js
andyholmes: Either's fine by me. REUSE won't find things it doesn't know about anyways, so having the snippet outside the header seems relatively moot.

There's also `SDPX-SnippetLineRange: <start>:<finish>` that can be used in the header, although REUSE probably doesn't recognize that field either.

andyholmes:
I'll need to revisit this sometime in the future, unless someone else wants to take it over.

The latest REUSE linter has some issues, so I don't mind putting it off a little longer to see how that plays out.

Issue #1516, opened 2022-11-15
GSConnect doesn't work in GNOME 43 lock screen


shdwchn10:
**Describe the bug**

Right after entering in GNOME 43 lock screen GSConnect stops

**Steps To Reproduce:**

1. Enter lockscreen
2. Try to do something from your KDE Connect phone app
3. See that your PC was disconnected

**Expected behavior**

GSConnect should work in lockscreen as it was with GNOME 42

**Support Log**

https://pastebin.com/JFGVtUwv

**System Details (please complete the following information):**

 - **GSConnect version:** 54
   - **Installed from:** Fedora Linux repos
 - **GNOME/Shell version:** 43.1
 - **Distro/Release:** Fedora Silverblue 37.20221113.0

**GSConnect environment (if applicable):**

 - **Paired Device(s):** Pixel 5a
 - **KDE Connect app version:** 1.19.1

andyholmes:
See https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1475#issuecomment-1271910239 for details about why this changed.

MacTavishAO:
Removing [this line](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/main/src/extension.js#L291) should help you with that. Maybe this behaviour should be bound to a boolean @andyholmes?

andyholmes:
Sure, we could add an option. Somewhere in the main preferences would probably work.

andyholmes:
For an interested contributor, this would involve:

1. Adding a boolean setting to [`org.gnome.Shell.Extensions.GSConnect.gschema.xml`](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/main/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml)
2. Adding a UI element, like a switch, probably to [`preferences/service.js`](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/main/src/preferences/service.js)
3. Check the setting before calling `this.service.stop()` at [`extension.js`, Line #293](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/1e856ae5ce6fc5fc3a38e5ccab90a794a035b5c0/src/extension.js#L293)

bluepostit:
Hi there, @andyholmes. I haven't built anything with GJS before, but I want to try to fix this issue as it affects me. 

[...]

PR #1537, opened 2022-12-10
Keep service alive when locked

bluepostit:
Currently in Gnome 43, the GS Connect service [stops when the screen is locked](https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1516). This means that it's impossible to unlock the screen via KDE Connect/GS Connect commands, as was previously possible.

This Pull Request aims to fix this, following the recommendations in #1516.

## Still To Do

- [ ] Make the switch component in the Preferences dialog reflect the actual state of the `keep-alive-when-locked` settings variable, both in its initial render and when it is toggled.
- [ ] Translations for the two new strings?
- [ ] Review of the UI components and their layout - might not be the best way to do this

### DIFF
--- a/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml
+++ b/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml
@@ -15,6 +15,9 @@ SPDX-License-Identifier: GPL-2.0-or-later
     <key name="show-indicators" type="b">
       <default>false</default>
     </key>
+    <key name="keep-alive-when-locked" type="b">
+      <default>true</default>
+    </key>
 
     <!-- Service Settings -->
     <key name="id" type="s">
--- a/data/ui/preferences-window.ui
+++ b/data/ui/preferences-window.ui
@@ -345,6 +345,65 @@ SPDX-License-Identifier: GPL-2.0-or-later
                           </packing>
                         </child>
                         <child>
+                          <object class="GtkLabel" id="behavior-when-locked-label">
+                            <property name="visible">True</property>
+                            <property name="can_focus">False</property>
+                            <property name="halign">start</property>
+                            <property name="margin_bottom">12</property>
+                            <property name="label" translatable="yes">Behavior When Locked</property>
+                            <attributes>
+                              <attribute name="weight" value="bold"/>
+                            </attributes>
+                          </object>
+                          <packing>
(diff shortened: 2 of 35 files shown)


andyholmes:
Looks pretty good so far. Now you just need to add a line to [`src/preferences/service.js`](https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/main/src/preferences/service.js#L226) like:

```js
this.add_action(this.settings.create_action('keep-alive-when-locked'));
```
And of course, uncomment the `<property name="action-name">win.keep-alive-when-locked<property>`, changing the action group from `settings.keep-alive-when-locked` to `win.keep-alive-when-locked`.

To update the translated strings, you can run `ninja -C _build gnome-shell-extension-gsconnect-update-po`, where `_build` is whatever build directory you used.

bluepostit:
Thanks, @andyholmes! 

I have made the changes. The setting seems to 'stick' now when I toggle it then re-open, and a few basic tests show that the service stays alive, only when the setting is turned on.

However, I'm not sure if the translation update worked properly. 

Previously, I had run `ninja -C _build install-zip` to build. So I ran `ninja -C _build org.gnome.Shell.Extensions.GSConnect-update-po` to update the translation strings. It's the most similar to the command you wrote in your previous comment, which didn't work (I got this error: `ninja: error: unknown target 'gnome-shell-extension-gsconnect-update-po'`

andyholmes:
> However, I'm not sure if the translation update worked properly.

Oh that's right. I got the command name a bit off, but that's correct. Crowdin and gettext break lines differently, but it's all the same once parsed and compiled.

> Also eg. the copyright notice at the top of [this file](https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1537/commits/082b23e9b4f56e575787f2b7a46362df30d2f197#diff-5d787d2060d290c305d0d7487a5b15ea2282b86e1c7e50869458d7e8c53613c0)

Ah yes, I noticed that the other day. We really can't put custom licenses in `.po` files because they'll just get wiped out. I'll add an exception for them in a new PR after merging this one.

I think this is more than good enough to get the ball rolling, so let's merge it into `main` and get some testing. Thanks for getting this fixed for everyone!

bluepostit:
Got it. Thanks for your assistance, @andyholmes! Glad to be able to help :)

PR #1557, opened 2023-01-29
RunCommand Plugin: add default commands

### LINKED ISSUE #1170: Command presets
**Is your feature request related to a problem? Please describe.**
Currently there's no command presets in GSConnect to select from, but you have to insert every command manually.

**Describe the solution you'd like**
In the Commands menu in GSConnect program, make it an option where we can select from some of the basic (or advanced) commands to be added, without having to enter them manually. Commands like:
- Shutdown
- Restart
- Lock screen
- Unlock screen 
- Suspend, etc.

 - GSConnect version: 45
   - Installed from: extensions.gnome.org

**Additional context**
In Kde Connect you can select from such command presets that i listed... i'm asking if GSConnect could have such a preset list, where by clicking on the option would add it. It would be very useful, especially for Linux users like myself, who don't know terminal commands, And it could also be useful for advanced users, because it's more convenient than typing every command.

pesader:
Add preset commands for session and power management. Their names match those used by GNOME Desktop itself (e.g. "Power off" instead of "Shutdown", "Restart" instead of "Reboot", etc). Since the commands were added as a single line in the XML file, here's a more human-readable list of them:

| ID | Name | Command |
|----|------|---------|
|lock|Lock| `xdg-screensaver lock`|
|logout|Log Out| `gnome-session-quit --logout`|
|restart|Restart|`gnome-session-quit --reboot --force`|
|poweroff|Power Off| `gnome-session-quit --power-off --force` |
|suspend|Suspend| `systemctl suspend` |

I marked this as a draft because:

1. The command to suspend uses `systemctl`, but I'm not sure every GNOME Desktop installation has that available. From what I've researched, GNOME depends only on `systemd-logind` but I'm not sure if that is the systemd component that is providing the command I've used.
2. Is there a way to add line breaks such that we have one command per line? Showing everything in a single line doesn't look good and some text editors may have trouble syntax highlighting it.

fixes #1170.

### DIFF
--- a/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml
+++ b/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml
@@ -150,7 +150,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
   <schema id="org.gnome.Shell.Extensions.GSConnect.Plugin.Presenter"/>
   <schema id="org.gnome.Shell.Extensions.GSConnect.Plugin.RunCommand">
     <key name="command-list" type="a{sv}">
-      <default>{}</default>
+      <default><![CDATA[{'lock': <{'name': 'Lock', 'command': 'xdg-screensaver lock'}>, 'restart': <{'name': 'Restart', 'command': 'gnome-session-quit --reboot --force'}>, 'logout': <{'name': 'Log Out', 'command': 'gnome-session-quit --logout'}>, 'poweroff': <{'name': 'Power Off', 'command': 'gnome-session-quit --power-off --force'}>, 'suspend': <{'name': 'Suspend', 'command': 'systemctl suspend'}>}]]></default>
     </key>
   </schema>
   <schema id="org.gnome.Shell.Extensions.GSConnect.Plugin.SFTP">


andyholmes:
> 1. The command to suspend uses systemctl [...]

I think that's okay? From what I can see GNOME Shell itself makes a direct D-Bus call to logind, and I don't think it's unfair to expect non-systemd users to report problems.

> 2. Is there a way to add line breaks such that we have one command per line?

Unfortunately not, AFAIK, since this is serialized GVariant and I don't think new-lines are allowed whitespace.

pesader:
Hey, thanks for timely feedback. Once I figure out why the CI is failing, I'll mark this PR as "Ready for review" :)

andyholmes:
Looks like you will have to wrap the default in `<![CDATA[` (content) `]]>`, otherwise the `<`/`>` throw off the parser.

pesader:
I would've taken ages to figure that out 😅 
You were right! [I ran the CI on my fork](https://github.com/pesader/gnome-shell-extension-gsconnect/actions/runs/4038367791) and it is passing all tests.

andyholmes:
LGTM, thanks!

PR #1560, opened 2023-01-31
Run Command Plugin: use systemctl for "Restart" and "Power Off" commands.

pesader:
This is a follow up to my previous PR, #1557.

Before this PR, the "Power Off", "Reboot", and "Log Out" commands displayed a confirmation dialog when ran. While this is a much welcome safety net when using the computer normally, it becomes annoying when running the commands remotely from a phone (requiring the user to go into Remote Input to click and confirm the operation).

Since we discussed it was okay to rely on systemd, I rewrote the commands for "Restart" and "Power Off" using `systemctl`.


pesader:
I tested running the "Log Out" command multiple times from my phone, and after a few runs every extension in my computer is disabled. It seems GNOME Shell is interpreting these logouts as an unintentional defect.

In that case, should we keep the "Log Out" command?

andyholmes:
> In that case, should we keep the "Log Out" command?

Hmm, not sure. Maybe keep it for now, let folks open bugs to see if its just a "you" issue?

pesader:
Sounds good :)

PR #1561, opened 2023-02-02
Fixed Spanish message for `Share Notifications`

alcarraz:
ditto


ferdnyc:
Thanks, @alcarraz! Yeah, it looks like we need a translator comment on that one, explaining that "share" is a verb in that context. I've added that in as well.

PR #1562, opened 2023-02-03
crowdin.yml: Add Finnish, Serbian (Latin) mappings

ferdnyc:
We had a request over @ Crowdin to add Finnish translations. Which we've had since 2021, but they weren't enabled in the Crowdin UI. I've changed that, and also enabled Serbian (Latin), which (as `sr-CS`) is how Crowdin lists our `sr@latin.po` locale. So, this also adds that mapping, and sorts the list of locale mappings in `crowdin.yml` to make it comprehensible.

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -9,28 +9,30 @@ files:
     languages_mapping:
       locale:
         ar-SA: ar
+        be: be
         ca-ES: ca
         ca: ca
-        fr: fr
+        cs: cs
         da: da
         de: de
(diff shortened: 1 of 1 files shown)


ferdnyc:
I'm not _entirely_ sure this is what's needed to get the existing Finnish translations to show up in Crowdin, but it can't hurt. (We should really document the translation-maintenance process(es) step-by-step in the Wiki at some point, for the reference of future maintainers... or even us current ones.)

Issue #1573, opened 2023-03-02
Add support for GNOME 44


Scotty-Trees:
### Describe your request

Greetings,

I know it's still a little bit early, but Gnome 44 alpha/beta are out now for testing purposes and I was wondering if/when the extension will be updated? Some extensions are already being updated to support the latest which is very nice to see. The final release for Gnome 44 should be on March 22nd, 2023 btw. Thank you for your time and thank you for making an amazing extension I couldn't live without! Cheers

### GSConnect version

54

### Installed from

Extension Manager

andyholmes:
For anyone interested in working on this, there is a guide for [Porting to GNOME 44](https://gjs.guide/extensions/upgrading/gnome-shell-44.html) available on [gjs.guide](https://gjs.guide).

sidevesh:
In case someone is taking a look at picking this up,
do consider https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1506 please.
Thanks.

yochananmarqos:
Well, especially since the tile on GNOME 44 is blank, I guess it's now a clean slate. 😜

The extension works otherwise, by the way.

retrixe:
One thing that could be improved (before releasing, possibly?) is adding the connected devices as a subtext like the updated Bluetooth sub-menu in GNOME 44:
![image](https://user-images.githubusercontent.com/6884828/226750887-820fd6f1-1cbe-40b7-839f-c7373c307ab4.png)
![image](https://user-images.githubusercontent.com/6884828/226750951-8dece960-6138-4ecc-a22a-a77ef5a7e6ef.png)
![image](https://user-images.githubusercontent.com/6884828/226751272-a657a029-6068-4778-a62a-7ccce8149cbb.png)
This would also solve #1506

maggu2810:
But isn't it a general improvement? How is this improvement related to the porting of existing functionality to GNOME 44?

yochananmarqos:
@retrixe It's already been done with the [Valent extension](https://github.com/andyholmes/gnome-shell-extension-valent):

![image](https://user-images.githubusercontent.com/5437803/226756038-6c02840a-4508-441c-b6c6-80a74b7004e4.png)

retrixe:
This was actually rather easy to do (and the code change is simple and shouldn't introduce any bugs), so I've made a pull request here with the necessary functionality to get this working: #1581 

[...]

PR #1577, opened 2023-03-11
Port to gnome 44

istvan-derda:
This patch intends to add support for gnome 44 and is very much work in progress. If you don't want WIP PRs in the repo, I'll happily retract it and submit when I'm ready. I don't have much experience with contributing to open source, not 100% sure about the ins and outs.

Status of the PR:
- [x] Extension installable on Gnome 44
- [x] QuickToggle looks as on Gnome 43
- [x] Quick toggle will be rendered above the background apps menu
- [x] No suspicious gnome-shell warnings logged when manually testing the extension

:exclamation: <del>The QuickToggle is appended _after_ the 'Background Apps' menu item. This looks rather weird, but not outlandish. </del>

<del>:exclamation: The QuickToggle is missing the icon it once had. I'm still looking into how it was originally displayed and how to bring it back.</del>

Solves #1573

### DIFF
--- a/src/extension.js
+++ b/src/extension.js
@@ -329,6 +329,12 @@ class ServiceIndicator extends QuickSettings.SystemIndicator {
         // Add the indicator to the panel and the toggle to the menu
         QuickSettingsMenu._indicators.insert_child_at_index(this, 0);
         QuickSettingsMenu._addItems(this.quickSettingsItems);
+
+        // Ensure the tile(s) are above the background apps menu
+        for (const item of this.quickSettingsItems) {
+            QuickSettingsMenu.menu._grid.set_child_below_sibling(item,
+                QuickSettingsMenu._backgroundApps.quickSettingsItems[0]);
+        }
     }
 
     destroy() {
--- a/src/extension.js
+++ b/src/extension.js
@@ -41,7 +41,7 @@ const ServiceToggle = GObject.registerClass({
 
     _init() {
         super._init({
-            label: 'GSConnect',
+            title: 'GSConnect',
             iconName: 'org.gnome.Shell.Extensions.GSConnect-symbolic',
             toggleMode: true,
         });
--- a/src/extension.js
+++ b/src/extension.js
@@ -45,7 +45,7 @@ const ServiceToggle = GObject.registerClass({
             toggleMode: true,
         });
 
-        this.set({iconName: 'org.gnome.Shell.Extensions.GSConnect-symbolic'})
+        this.set({iconName: 'org.gnome.Shell.Extensions.GSConnect-symbolic'});
 
         // Set QuickMenuToggle header.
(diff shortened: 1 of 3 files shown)


istvan-derda:
I spent some time reading the gnome-shell code - as far as I can tell there is no good solution for inserting the quick toggle reliably before the 'Background Apps' QuickSettingsItem. Other than that this patch is working fine on Gnome 44 :)

src/extension.js
```diff
@@ -41,7 +41,7 @@ const ServiceToggle = GObject.registerClass({
 
     _init() {
         super._init({
-            label: 'GSConnect',
+            title: 'GSConnect',
```
DaPigGuy, line 44: This breaks compatibility for GNOME 43

DaPigGuy (changes requested):
image

The quick settings toggle icon seems to have disappeared on GNOME 44...

src/extension.js
daniellandau: I didn't test yet, thanks for the testing help! In general we only support the latest shell version, so  the correct solution is to drop 43 from metadata.json

istvan-derda: Thanks for catching that, I totally missed it. @daniellandau If you have the time to fix the missing icon feel free - otherwise I'll try to come back to this when I have time. :)

istvan-derda:
> **GNOME 44** image
> 
> **GNOME 43** image
> 
> The quick settings icon seems to have disappeared on GNOME 44...

should be fixed with the latest commit. Thanks for catching it though!

image

daniellandau:
Took a while to manage finding a 'background app', but yes indeed that's a bad place for the quicktoggle. Otherwise looking really good and I understood in Matrix that you should have an idea already on how to fix it. (Might want to contribute the fix to caffeine too BTW)

istvan-derda:
> Took a while to manage finding a 'background app', but yes indeed that's a bad place for the quicktoggle. Otherwise looking really good and I understood in Matrix that you should have an idea already on how to fix it. (Might want to contribute the fix to caffeine too BTW)

Yes, @andyholmes helped me out there. That behavior is now fixed :)

src/extension.js
```diff
@@ -41,11 +41,12 @@ const ServiceToggle = GObject.registerClass({
 
     _init() {
         super._init({
-            label: 'GSConnect',
-            iconName: 'org.gnome.Shell.Extensions.GSConnect-symbolic',
+            title: 'GSConnect',
             toggleMode: true,
         });
 
+        this.set({iconName: 'org.gnome.Shell.Extensions.GSConnect-symbolic'})
```
DaPigGuy, line 48:
```suggestion
        this.set({iconName: 'org.gnome.Shell.Extensions.GSConnect-symbolic'});
```

andyholmes:
LGTM, let's merge this so we can get some testing before Wednesday when GNOME 44 drops.

Thanks for the work everyone!

PR #1581, opened 2023-03-21
Set Quick Settings tile subtitle on GNOME 44

### LINKED ISSUE #1506: Better quick settings tile on Gnome 43
Currently the quick settings tile on Gnome 43 just says GSConnect which looks a bit out of place and doesn't provide any useful info at glance,
can it be replaced with maybe the device battery, network signal etc or maybe how many devies are connected if multiple devices are connected.

retrixe:
Closes #1506 
Addresses my comment here: https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1573#issuecomment-1478645384

I based this off of the old code from GNOME 42 (which was removed in the GNOME 43 update, since this wasn't available back then, of course) and [the code of the improved Bluetooth Quick Settings tile in GNOME 44](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/status/bluetooth.js#L364).

Preview:
image
image
image

The code changes here are simple, so they shouldn't introduce any bugs. It would be great if we can get this merged prior to releasing the GNOME 44 version.

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: gsconnect\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-12-10 23:37+0100\n"
+"POT-Creation-Date: 2023-03-22 10:01+0530\n"
 "PO-Revision-Date: 2022-12-25 18:09\n"
 "Last-Translator: \n"
 "Language-Team: Arabic\n"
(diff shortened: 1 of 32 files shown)


andyholmes (changes requested):
This looks good, but needs the pot file updated for the new strings, by running ninja:

```
ninja -C _build org.gnome.Shell.Extensions.GSConnect-update-po
```

andyholmes (approved):
Thanks!

PR #1583, opened 2023-03-24
Update QuickMenuToggle header to Devices

sidevesh:
Change the GSConnect label in Quick settings toggle to just Devices, similar to Valent's gnome extension https://github.com/andyholmes/gnome-shell-extension-valent/blob/e7f759047c45833cd211ef18a8554008cb1b8b12/src/extension/status.js#L255 This makes it look better integrated with the system.

### DIFF
--- a/src/extension.js
+++ b/src/extension.js
@@ -48,7 +48,7 @@ const ServiceToggle = GObject.registerClass({
         this.set({iconName: 'org.gnome.Shell.Extensions.GSConnect-symbolic'});
 
         // Set QuickMenuToggle header.
-        this.menu.setHeader('org.gnome.Shell.Extensions.GSConnect-symbolic', 'GSConnect',
+        this.menu.setHeader('org.gnome.Shell.Extensions.GSConnect-symbolic', 'Devices',
             _('Sync between your devices'));
 
         this._menus = {};


andyholmes:
I'm okay with this, but I'm going pull in Daniel and Frank for their thoughts.

daniellandau:
As I said on the issue, IMO GSConnect is fine as a label.

ferdnyc:
Sorry, missed this until now. But for the record, I'm in agreement that "Devices" is too generic. Rather than that, I'd prefer to see the QS button display the actual connected device _name_. (As somewhat suggested by @sidevesh in #1506. Though as I noted there, one immediate complication is: How do multiple paired devices work?)

At the time I was pointing to the Bluetooth QS widget as a possible source of inspiration for ways to handle this issue, since it was the best potential analogue to GSConnect's needs. Joke was on me, though, because the Bluetooth QS button is actually a terrible example of anything. Its ONLY function is to turn the radio on/off entirely, and to even access a device list you have to ignore it and  go to the Control Center instead.

I agree some sort of device-name display would better fit with the built-in widgets (the way the network button shows the current profile name, ditto the power-profile button...), but there appear to be significant barriers to that approach, and once you take it off the table, just "GSConnect" is the least worst remaining bad option.

PR #1592, opened 2023-04-04
Fix deduplication logic for fdo notifications

### LINKED ISSUE #1591: GSConnect ignores some/all freedesktop Notifications
### Describe the bug

In a discussion (#1586), @aplumez asked why GSConnect wasn't registering / able to relay notifications created by Google Chrome. It turns out, <s>we're not looking at the correct field(s) of the notification body, to match those notifications up to their source application.</s> (See below.)

### Steps to reproduce

1. Visit https://www.bennish.net/web-notifications.html using Google Chrome
2. Authorize notifications for the site, and then use the Show button to create a test notification.
3. Look for Chrome in a device's Notifications list — it will not appear

### Expected behavior

GSConnect recognizes the test notification as coming from Chrome, and adds it to the list of applications for which notifications can be shared.

### GSConnect version

54

### Installed from

other source

### GNOME Shell version

43.3

Fedora 37

### Paired device(s)

--

### KDE Connect app version

--

### Plugin(s)

_No response_

### Support log

_No response_

### Screenshots

_No response_

### Notes

Sniffing notification traffic with `dbus-monitor` proves enlightening:

```console
$ dbus-monitor --session "interface='org.freedesktop.Notifications'"
method call time=1680638594.410850 sender=:1.21695 -> destination=org.freedesktop.Notifications serial=250 path=/org/freedesktop/Notifications; interface=org.freedesktop.Notifications; member=Notify
   string "Google Chrome"
   uint32 0
   string "file:///tmp/.com.google.Chrome.3F5zhp"
   string "Notification #8"
   string "www.bennish.net

This is the text body of the notification. 
Pretty cool, huh?"
   array [
      string "default"
      string "Activate"
      string "settings"
      string "Settings"
   ]
   array [
      dict entry(
         string "urgency"
         variant             uint32 1
      )
      dict entry(
         string "desktop-entry"
         variant             string "google-chrome"
      )
      dict entry(
         string "image_path"
         variant             string "/tmp/.com.google.Chrome.TGa9S6"
      )
      dict entry(
         string "image-path"
         variant             string "/tmp/.com.google.Chrome.TGa9S6"
      )
   ]
   int32 -1
```

The first argument is what GSConnect detects as the `appName`, and that's set to `"Google Chrome"`. But what we _**should**_ be looking for (preferentially, at least, if present) is the [standard](https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html#hints) entry in the `hints` dictionary: `desktop-entry`.

The `desktop-entry` hint is set to `google-chrome`, which corresponds to the `google-chrome.desktop` file that's used to identify the application.

ferdnyc:
Commit 4734c8feb82ff9f955b774ed34c8b76b15fa4ca4 added a check for notification destination (intended as a deduplication measure) that was accidentally rejecting _all_ copies of (some? all?) freedesktop notifications. This relaxes that check.

Fixes #1591

### DIFF
--- a/src/service/components/notification.js
+++ b/src/service/components/notification.js
@@ -251,14 +251,19 @@ const Listener = GObject.registerClass({
             // libnotify
             } else if (name === 'Notify') {
                 const message = invocation.get_message();
+                const destination = message.get_destination();
 
-                if (this._fdoNameOwner === undefined) {
-                    this._fdoNameOwner = await this._getNameOwner(
-                        'org.freedesktop.Notifications');
+                // Deduplicate notifications; only accept messages
+                // directed to the notification bus, or its owner.
+                if (destination !== 'org.freedesktop.Notifications') {
+                    if (this._fdoNameOwner === undefined) {
+                        this._fdoNameOwner = await this._getNameOwner(
+                            'org.freedesktop.Notifications');
+                    }
+
+                    if (this._fdoNameOwner !== destination)
+                        return;
                 }
-
-                if (this._fdoNameOwner !== message.get_destination())
-                    return;
 
                 // Try to brute-force an application name using DBus
                 if (!this.applications.hasOwnProperty(parameters[0])) {
@@ -441,4 +446,3 @@ const Listener = GObject.registerClass({
  * The service class for this component
  */
 var Component = Listener;
-


ferdnyc:
You are a nitpicky bastard, eslint. :stuck_out_tongue_closed_eyes:

andyholmes (approved):
LGTM

PR #1596, opened 2023-04-05
Add startup permissions checks

ferdnyc:
Over in discussion #1586, when @aplumez attempted to use one of the GitHub Actions artifact zip files as a GSConnect installer, we discovered that there's an issue with file permissions. It seems that GitHub strips the executable bit off of all the files it adds to the artifact zip. That causes the installed GSConnect not to be able to launch its preferences, or start `daemon.js` or `nativeMessagingHost.js`

Since installable Artifacts are still a very useful testing tool, this PR addresses the issue with an additional check performed during the extension `_init()`: `Utils.ensurePermissions()`. Analogous to the existing `Utils.installService()` call, `ensurePermissions()` will (iff the extension is installed as a user extension) check the executable bits on those three files, and if they're not set, will attempt to set them.

Doing that startup check ensures that a GSConnect installed from a GitHub Artifacts zip, or one that's been misinstalled through some other means, will be made functional again, instead of creating difficult-to-diagnose problems.

### DIFF
--- a/src/extension.js
+++ b/src/extension.js
@@ -364,6 +364,14 @@ var serviceIndicator = null;
 var lockscreenInput = null;
 
 function init() {
+    // If installed as a user extension, this checks the permissions
+    // on certain critical files in the extension directory
+    // to ensure that they have the executable bit set,
+    // and makes them executable if not. Some packaging methods
+    // (particularly GitHub Actions artifacts) automatically remove
+    // executable bits from all contents, presumably for security.
+    Utils.ensurePermissions();
+
     // If installed as a user extension, this will install the Desktop entry,
     // DBus and systemd service files necessary for DBus activation and
     // GNotifications. Since there's no uninit()/uninstall() hook for extensions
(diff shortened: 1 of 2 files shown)


ferdnyc:
I'd also like to note, as a side observation, that the `Gio.File` / `Gio.FileInfo` documentation on file attributes is so obtuse, and so devoid of examples or explanations involving real-world scenarios, that I ended up having to go read the Nautilus code just to find out how a `GFile` object can be used to change the actual disk file's permission bits. :roll_eyes:

andyholmes:
> I'd also like to note, as a side observation, that the `Gio.File` / `Gio.FileInfo` documentation on file attributes

Yeah, probably we should have some docs for this, at least for GJS. If you want to open an issue on gjs.guide I plan on spending some time there next month or so pumping some content into it :slightly_smiling_face:

ferdnyc:
@andyholmes 
> Yeah, probably we should have some docs for this, at least for GJS. If you want to open an issue on gjs.guide I plan on spending some time there next month or so pumping some content into it slightly_smiling_face

[Done.](https://gitlab.gnome.org/ewlsh/gjs-guide/-/issues/60) I was not... _especially_ gentle. (Though, maybe for me, I was.) :angel:

PR #1614, opened 2023-04-29
Use a more descriptive label for extension-wide keepalive setting

### LINKED ISSUE #1613: "Keep Alive" option isn't very clear
### Describe the bug

In GSConnect settings, there is a option called "Keep Alive" with the title "Behavior when locked". However, these strings aren't very clear. What will be kept alive, KDE Connect or GSConnect or both? And what "alive" means? Does "locked" apply for the phone or for the desktop/notebook? 

### Steps to reproduce
Pair a device and then go open the "Mobile Settings" item from GSConnect submenu from quick settings.

### Expected behavior

_No response_

### GSConnect version

55

### Installed from

GNOME Extensions website

### GNOME Shell version

gnome-shell-44.0-4.fc38.x86_64

Fedora 38

### Paired device(s)

S20 FE

### KDE Connect app version

1.24.3

### Plugin(s)

_No response_

### Support log

_No response_

### Screenshots

Screenshot from 2023-04-25 17-43-04

### Notes

_No response_

ferdnyc:
The renamed section title is based on the expectation that there could be other extension-wide settings, eventually.

image

Fixes #1613

### DIFF
--- a/data/ui/preferences-window.ui
+++ b/data/ui/preferences-window.ui
@@ -370,18 +370,6 @@ SPDX-License-Identifier: GPL-2.0-or-later
                             <property name="margin_bottom">20</property>
                             <property name="spacing">24</property>
                             <child>
-                              <object class="GtkLabel" id="keep-alive-when-locked-label">
-                                <property name="visible">True</property>
-                                <property name="can_focus">False</property>
-                                <property name="halign">start</property>
-                                <property name="margin_bottom">12</property>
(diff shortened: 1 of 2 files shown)


ferdnyc:
> image

(About the control order change: Unless the switch is embedded in a full-width row inside a list widget, placing it before the text — in the manner of a radio button or check box — tends to be cleaner, because it means that multiple switches will be aligned. (When there are multiple switches.) Also, the switch won't change position with the length of the descriptive text.)

andyholmes (approved):
LGTM

PR #1622, opened 2023-05-04
(Even-more-radical) translations fixes

### LINKED ISSUE #1620: Translation files have wrong names
### Describe the bug

Many translations are duplicated or have an unnecessary country tag which limits their use, e.g. `ar_SA.po` won’t be used when the system locale is `ar_KW`. To make the most out of the translations you have and comply with other projects (like [GNOME Shell](https://gitlab.gnome.org/GNOME/gnome-shell/-/tree/main/po)), here are the changes you should make:

`ar.po` and `ar_SA.po` → `ar.po`,

`fa-IR.po`, `fa_IR.po` and `fa.po` → `fa.po`, 

`fi_FI.po` and `fi.po` → `fi.po`,

`fy-NL.po` and `fy_NL.po` → `fy.po`,

`id-ID.po` and `id_ID.po` → `id.po`,

`ko-KR.po`, `ko_KR.po` and `ko.po` → `ko.po`,

`nl_BE.po` and `nl_NL.po` → `nl.po`,

`pt_PT.po` → `pt.po` (BUT `pt_BR.po` should stay as-is),

`sr_CS.po` and `sr@latin.po` → `sr@latin.po`,

`sv.po`, `sv-SE.po` and `sv_SE.po` → `sv.po`.

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### GSConnect version

Newest git

### Installed from

GitHub releases (zip file)

### GNOME Shell version

_No response_

_No response_

### Paired device(s)

_No response_

### KDE Connect app version

_No response_

### Plugin(s)

_No response_

### Support log

_No response_

### Screenshots

_No response_

### Notes

_No response_

ferdnyc:
In #1620 @piotrdrag brought up the mess we've currently got going on in the `po/` directory, and with translation naming in general, prompting me to (once again) revisit the issue. Admittedly, I'd been dragging my feet on returning to it, because it always makes my head spin.

(And I never seem to know, even after-the-fact, whether I've ended up making things _better_ or _**worse**_ for our translators. And/or our users who rely on localization. I _hope_ better! But `¯\_(ツ)_/¯`!)

You can read [my long comment at the issue](https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1620#issuecomment-1534868871) for the rationale behind a lot of this, as a starting point. But actually, this PR deviates from the plan outlined there, going a few steps further in ways I'll detail now:

#### New default filename template (AGAIN)

When I last changed the Crowdin configuration (in #1564), I took a cautious approach to addressing the translation filename template, replacing `%locale%` (Crowdin's default `xx-YY` language names) with `%locale_with_underscore%` to have Crowdin generate `xx_YY.po` filenames instead.

But (as I knew even at the time), we don't really **want** `xx_YY.po` filenames any more than we want `xx-YY.po` filenames. Except in a very few special cases (`zh_*.po`, `sr@latin.po`, `pt_BR.po`, `nl_NE.po`), what we want are `xx.po` filenames, which correspond to Crowdin's `%two_letters_code%` placeholder.

On the strength of Crowdin's own [GetText Project example](https://developer.crowdin.com/configuration-file/#example-configurations) from their documentation, I'm now more convinced than ever that the _correct_ configuration for our needs is to use `po/%two_lettters_code%.po` as our translation output filename template, and then add override mappings **only** for the relatively few languages where `%two_letters_code%` is meant to mean something else.

In the updated configuration file, that set of overridden mappings is:

```yml
    languages_mapping:
      two_letters_code:
        nl-BE: nl_BE
        nl-NL: nl
        pt-BR: pt_BR
        pt-PT: pt
        sr-CS: sr@latin
        sr-SP: sr
        zh-CN: zh_CN
        zh-TW: zh_TW
```

...Using this configuration _should_, based on all of the information I've been able to glean from Crowdin's docs and interface, cause Crowdin to add _new_ languages as `xx.po` automatically, and require no intervention on the name mapping except when new cases arise where that's **not** what we want.

#### LINGUAS (AGAIN)

We also have to remember to add the new code — two-letter or otherwise — to the `po/LINGUAS` file that drives the translation compile step of the build. Every damn time.

As I note in one of the commit messages here, it really seems like that _shouldn't_ be necessary, since the contents of the file should effectively always match the output of this command:

```sh
cd po; ls *.po | sed -e 's/\.po$//' > LINGUAS
```
...Since the contents _are_ so rigidly defined and easily scriptable, it seems like the sort of thing that shouldn't require a PR to the repo to update. Why can't the build system just generate that listing on-the-fly instead, so that we don't have to manage that extra file?

That question is not satisfactorily answered, from my perspective, but for now I've set it aside and just stuck with the devil (`po/LINGUAS`) we know. The PR does update it, by simply running the command above and committing the results.

(HOPEFULLY...)
Fixes #1620


andyholmes:
Looks pretty good to me. FWIW the `sr`/sr@latin` are contributed by someone directly (they didn't want to use Crowdin), although I haven't seen them in awhile they seemed pretty confident with their stuff.

`cd po; ls *.po | sed -e 's/\.po$//' > LINGUAS` could be modified a bit to `diff` the existing file in a CI job, if that's wanted.

ferdnyc:
@andyholmes 

> FWIW the `sr`/sr@latin` are contributed by someone directly (they didn't want to use Crowdin), although I haven't seen them in awhile they seemed pretty confident with their stuff.

Hmm, that's helpful info, thanks.

It's _entirely_ possible we (read: **I**) screwed that up on them, in the last round of renames & resyncs. I'll try to dig through the history and see where the cyrillic strings slipped in.

andyholmes:
Is what's here good to merge? I don't see any problems myself, and there's time to double check before rolling a release.

ferdnyc:
I wrote, re: the Cyrillic in `sr@latin.po`:

> It's entirely possible we (read: I) screwed that up on them, in the last round of renames & resyncs. I'll try to dig through the history and see where the cyrillic strings slipped in.

I finally found time to look at this. Turns out, I didn't do it! \o/

...But it's still very likely my fault, for misusing/misconfiguring Crowdin.

In commit c7358715773cf6c20abe1426d3a8f39a10a1c1c2 I painstakingly migrated all of the Latin-script Serbian strings to the `sr@latin.po` file.

Three hours later, in 01fc4ab8db0d6db817458c364438095d541cb0d6, the very next Crowdin update dumped a bunch of Cyrillic text into that file, showing that Crowdin still considered it the output file for the Serbian translations.

There's probably a "right way" to get Crowdin to re-**import** the _Latin_-script translations properly, but my skills with its tools are (as I believe I've mentioned multiple times already, just in _this_ PR) not really up to the task.

Still, that's _already_ broke, and this PR won't make it any broke-r. (It might even help! ...I doubt it.) Something that can be fixed after this goes in just as easily as before, though, for sure. I don't see any reason not to merge this, and then afterwards hopefully just pick up the pieces of whatever breaks.

(I do apologize to @Xabre666 for making a mess of their painstakingly-prepared translations, though.)

I've just rebased the branch on `main` and incorporated a few translation updates, as well. Should be good to go.

andyholmes (approved):

PR #1632, opened 2023-05-27
preferences-window: Improve the description of the "keep alive" option

automerisnaranja:
Follow-up from https://github.com/GSConnect/gnome-shell-extension-gsconnect/commit/45e13005170289ff6f04abc6f491ea439e9e9d6d

The option description is currently phrased as "GSConnect remains active when GNOME Shell is locked". However, this text has 2 problems:

- Options descriptions are usually phrased in the imperative mood, but this one isn't
- "GNOME Shell" can be a technical term for people who don't know what a shell is

This change addresses the 2 problems above by using "Keep GSConnect active when this device is locked".

Fixes https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1613#issuecomment-1561984145

### DIFF
--- a/data/ui/preferences-window.ui
+++ b/data/ui/preferences-window.ui
@@ -387,7 +387,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
                                 <property name="can_focus">False</property>
                                 <property name="halign">start</property>
                                 <property name="margin_bottom">12</property>
-                                <property name="label" translatable="yes">GSConnect remains active when GNOME Shell is locked</property>
+                                <property name="label" translatable="yes">Keep GSConnect active when this device is locked</property>
                                 <accessibility>
                                   <relation type="label-for" target="keep-alive-when-locked"/>
                                 </accessibility>


ferdnyc:
> * "GNOME Shell" can be a technical term for people who don't know what a shell is
> 
> This change addresses the 2 problems above by using "Keep GSConnect active when this device is locked".

Hmmm. I worry that this runs back into one of the original problems that was brought up in #1613: is it then clear enough _which_ "device" the text is referring to? For users who don't think of their computer as "this device", and only have one device _paired_ with their computer, they may take it to be referring to the paired phone/tablet/whatever "device".

Saying "GNOME Shell" unequivocally ties the statement to the _local_ system, the one where GSConnect is running. (As opposed to any Android device that may be paired.)

And if there really are that many people out there running GNOME Shell, who somehow _don't know what that is_, I'm sure the GNOME project would be thrilled if we helped to teach them. (They seem weirdly into brand recognition, for a project with no profit motive driving that goal.)

I mean, this IS the `gnome-shell-extension-gsconnect` repo... feels kind of like that train has sailed already, no?

...This is purely IMHO, though. My intention here is to initiate a discussion, not issue any sort of decree. Thoughts?

ferdnyc:
(Note to other maintainers: Before merging this, the translation source `.pot` file still needs to be regenerated with the updated text.)

andyholmes:
Maybe "Keep GSConnect active when the screen is locked"? I think either will work though.

Issue #987, opened 2020-10-02
Review upstream security advisory


andyholmes:
The KDE Project issued a security advisory, which should be reviewed to ensure none of the issues affect GSConnect.

https://kde.org/info/security/advisory-20201002-1.txt

andyholmes:
These seem to be the issues relevant to GSConnect:

* [Limit identity packets to 8KiB](https://invent.kde.org/network/kdeconnect-kde/-/commit/b496e66899e5bc9547b6537a7f44ab44dd0aaf38)
* [Do not let lanlink connections stay open for long without authenticating ](https://invent.kde.org/network/kdeconnect-kde/-/commit/5310eae85dbdf92fba30375238a2481f2e34943e)
* [Limit number of connected sockets from unpaired devices](https://invent.kde.org/network/kdeconnect-kde/-/commit/ae58b9dec49c809b85b5404cee17946116f8a706) (maybe)
* [Limit the ports we try to connect to to the port range of KDE Connect](https://invent.kde.org/network/kdeconnect-kde/-/commit/85b691e40f525e22ca5cc4ebe79c361d71d7dc05)
* [Do not replace connections for a given deviceId if the certs have changed](https://invent.kde.org/network/kdeconnect-kde/-/commit/48180b46552d40729a36b7431e97bbe2b5379306) (maybe)

PR #1638, opened 2023-06-11
GSconnect vulnerable to man-in-the-middle attack due to compatibility [Potential Fix] 

### LINKED ISSUE #1606: GSconnect vulnerable to man-in-the-middle attack due to compatibility
### Describe the bug

The following bug states some problem:
https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/930
and it seems it's the reason to undermine all users security is that some  users are experienciing issues with legacy devices. And by those are users who didn't update software in their phones since years 
This is by the line:
```
// Allow TLSv1.0 certificates
// See https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/930
imports.gi.GLib.setenv('G_TLS_GNUTLS_PRIORITY', 'NORMAL:%COMPAT:+VERS-TLS1.0', true
```
in:
> /$HOME/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js

### Steps to reproduce

_No response_

### Expected behavior

There should be an option to adjust TLS options but compromising security of service which is exposed on all interfaces(listen 0.0.0.0) should never be default setting

### GSConnect version

50

### Installed from

GNOME Extensions website

### GNOME Shell version

_No response_

_No response_

### Paired device(s)

_No response_

### KDE Connect app version

_No response_

### Plugin(s)

_No response_

### Support log

_No response_

### Screenshots

_No response_

### Notes

_No response_

mrdev023:
I enable all version of TLS like in https://invent.kde.org/network/kdeconnect-android/-/commit/e2dbc39e3a73cc6c7cf1c27362e4f2b768e88f42

It's my first contribution.

This PR can fix #1606

### DIFF
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -8,7 +8,7 @@
 
 // Allow TLSv1.0 certificates
 // See https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/930
-imports.gi.GLib.setenv('G_TLS_GNUTLS_PRIORITY', 'NORMAL:%COMPAT:+VERS-TLS1.0', true);
+imports.gi.GLib.setenv('G_TLS_GNUTLS_PRIORITY', 'NORMAL:%COMPAT:+VERS-ALL', true);
 
 imports.gi.versions.Gdk = '3.0';
 imports.gi.versions.GdkPixbuf = '2.0';


src/service/daemon.js
```diff
@@ -8,7 +8,7 @@
 
 // Allow TLSv1.0 certificates
 // See https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/930
```
andyholmes, line 10:
```suggestion
```

src/service/daemon.js
```diff
@@ -8,7 +8,7 @@
...
-imports.gi.GLib.setenv('G_TLS_GNUTLS_PRIORITY', 'NORMAL:%COMPAT:+VERS-TLS1.0', true);
+imports.gi.GLib.setenv('G_TLS_GNUTLS_PRIORITY', 'NORMAL:%COMPAT:+VERS-ALL', true);
```
andyholmes, line 11:
```suggestion
```

andyholmes (changes requested):
Thanks for the contribution!

If you want to remove the exception for TLS 1.0, you should just removed this entire block and let `glib-networking` handle keeping up with current security notices.

However, it is worth pointing out again that this will very much not help with MITM attacks. #987 contains a number of far more critical vulnerabilities.

bernieEV:
I have Ubuntu 22.04LTS with GSconnect and Android14 with KDEconnect
All is working fine. I did a wireshark trace on ethernet and found **TLSv1.2**
I want to limit TLS to version 1.3 only. how to ?

andyholmes:
If you want to restrict to TLSv1.3 you can muck around with this line:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/gnome-43/src/service/daemon.js#L11

This has been removed in the latest version, so you'll have to carry a patch yourself. The upstream documentation for this is here: [GnuTLS Priority Strings](https://gnutls.org/manual/html_node/Priority-Strings.html).

PR #1656, opened 2023-07-14
refactor: port to `Gio._promisify()`

andyholmes:
This tech-preview isn't going anywhere, and in fact will become a part
of GJS soon. Port to the more modern async-await patterns throughout
the service.

Depends on #1655

### DIFF
--- a/src/service/__init__.js
+++ b/src/service/__init__.js
@@ -14,6 +14,46 @@ const GLib = imports.gi.GLib;
 const Config = imports.config;
 
 
+// Promise Wrappers
+try {
+    const {EBook, EDataServer} = imports.gi;
+
+    Gio._promisify(EBook.BookClient, 'connect');
+    Gio._promisify(EBook.BookClient.prototype, 'get_view');
+    Gio._promisify(EBook.BookClient.prototype, 'get_contacts');
+    Gio._promisify(EDataServer.SourceRegistry, 'new');
+} finally {
+    // Silence import errors
+}
+
+Gio._promisify(Gio.AsyncInitable.prototype, 'init_async');
+Gio._promisify(Gio.DBusConnection.prototype, 'call');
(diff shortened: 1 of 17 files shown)


src/service/__init__.js
```diff
@@ -14,6 +14,46 @@ const GLib = imports.gi.GLib;
...
+Gio._promisify(Gio.AsyncInitable.prototype, 'init_async');
+Gio._promisify(Gio.DBusConnection.prototype, 'call');
+Gio._promisify(Gio.DBusProxy.prototype, 'call');
+Gio._promisify(Gio.DataInputStream.prototype, 'read_line_async',
+    'read_line_finish_utf8');
+Gio._promisify(Gio.File.prototype, 'delete_async');
+Gio._promisify(Gio.File.prototype, 'enumerate_children_async');
+Gio._promisify(Gio.File.prototype, 'load_contents_async');
+Gio._promisify(Gio.File.prototype, 'mount_enclosing_volume');
+Gio._promisify(Gio.File.prototype, 'query_info_async');
```
ferdnyc, line 38: When you say this "will be part of GJS soon", does that mean that this will happen automatically for the async functions? Because, _for the most part_ it looks like using Promisify makes the code a lot more concise (which is great), except that there's all this nonsense here to take its place. (While there's a lot **less** of it, compared to what it's replacing, it's still ugly boilerplate code all the same.)

andyholmes: Correct, it will become a part of the GObject-Introspection infrastructure once all the pieces have been put together, such as [girepository: Add APIs for sync, async, and finish function annotations](https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/404).

andyholmes:
I'm going to go ahead and take that as a review :)

Hopefully with this and other recent fixes it may be possible to actually generate coverage for the tests. I think that'll probably be good to have, considering the inertia this project has despite the slow development rate.

Issue #1640, opened 2023-06-12
The button Settings on the dock becomes inaccessible with too many devices or on lower display resolutions


bekopharm:
### Describe the bug

The button `Settings` at the bottom of the expanded device list on the dock becomes inaccessible if too many devices are connected. The expanded list of devices has no scrollbar. The individual devices can not be collapsed to make space.

Workaround is to use the `gear` symbol in `extension-manager` to configure `GSConnect` again. That's a lot of extra clicks though and I use that button a lot to configure which device gets the clipboard synced and which not.

### Steps to reproduce

1. Go to dock
2. Click on `GSConnect` to expand connected devices

### Expected behavior

A scrollbar should appear on mouse wheel or scroll gesture or listed devices should be able to collapse so the `Settings` button at the bottom show up again.

### GSConnect version

54

### Installed from

GNOME Extensions website

### GNOME Shell version

43.5

### Linux distribution/release

Fedora 37

### Paired device(s)

many

### KDE Connect app version

any

### Screenshots

![gnome-shell-screenshot-g3zins](https://github.com/GSConnect/gnome-shell-extension-gsconnect/assets/5016428/f063f096-f0f3-4d8d-8040-bd960df4edae)

frikisama:
You don't even need that many devices, the menu is cut with only one too, if the resolution is low and the menu full enough. Here is a screenshot of my device, with its 1600p display at 2x desktop scale. I guess 800p would have the same problem.

![image](https://github.com/GSConnect/gnome-shell-extension-gsconnect/assets/10727644/07f7b4d9-ef17-4dce-960a-6aa253f97207)

bekopharm:
Always boggles my mind that someone would indeed run Gnome on such low resolutions but then I remember that some use this indeed on phones and similar nowadays.

Yes, that's exactly the same problem 👍

frikisama:
Quick update: there is a related issue on the Gnome Shell issue tracker.

https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6056

ferdnyc:
I thought there was already an older issue open about this, but I can't find it.

As @frikisama notes, it's largely a GNOME issue — the menu should be scrollable if its contents are taller than the screen, but it's not. **That** aspect, I don't think we could solve in GSConnect.

[...]

PR #1664, opened 2023-08-10
Hide Waydroid devices from the quick settings panel

sidevesh:
Until waydroid resolves this https://github.com/waydroid/waydroid/issues/211, GSConnect is a really good and seamless alternative to get the notifications to show up natively,
but the Waydroid device showing up in quick settings looks ugly and sort of exposes the "implementation detail" of this workaround,
skipping the Waydroid device in quick settings makes things look neat.
The Waydroid device will still show up in the settings menu so no issues with configuring this.
I know this is "hacky", raising this mostly to get views from GSConnect team on this and also to help anyone else who might want to achieve the same thing as this.

### DIFF
--- a/src/extension.js
+++ b/src/extension.js
@@ -132,9 +132,10 @@ const ServiceToggle = GObject.registerClass({
     }
 
     _sync() {
-        const available = this.service.devices.filter(device => {
+        const availableBeforeFiltering = this.service.devices.filter(device => {
             return (device.connected && device.paired);
         });
+        const available = availableBeforeFiltering.filter(device => !(device.name && device.name.toLowerCase().includes('waydroid')));
         const panelMode = this.settings.get_boolean('show-indicators');
 
         // Hide status indicator if in Panel mode or no devices are available


ferdnyc:
@sidevesh Interesting.

Knowing nothing about Waydroid, if people are using GSConnect in this way I'm wondering if, rather than special-casing, it wouldn't be better to just add a new preference for _all_ devices, controlling whether or not they're visible in the menu.

Given #1640, that would also provide users at least a partial workaround because they could at least choose _which_ device to display (by hiding others), in cases where multiple devices don't fit in the available screen height.

sidevesh:
Option to hide devices from quick settings does make sense, one question I guess will be that if there are devices connected but none selected then what should the quick settings tile show

ferdnyc:
@sidevesh 

> Option to hide devices from quick settings does make sense, one question I guess will be that if there are devices connected but none selected then what should the quick settings tile show

I'd argue that the _tile_ should be unchanged — even if a device's menu is hidden, it should show the name of the device if it's the only one connected, and "<var>n</var> connected" if there are multiples.

It still makes sense to report the status of _all_ connected devices, so that the user can **know** when hidden ones are connected. (Or not connected — i.e. if a hidden device falls off the network, the change in the tile will be a means for the user to "notice" that.)

IOW, my vision of "hiding" is not so much that the _device_ is hidden, merely that its menu is not displayed. (In fact, the preference name I went with in my WIP attempt at implementing this was `menu-visible`, with a display name of "Visible in Menu" (default enabled), rather than anything involving "hidden".)

sidevesh:
@ferdnyc while that makes sense, it does not work for this hack,
showing the Waydroid device in quick settings tile if its the only one connected will as I said expose the implementation detail of this hack,
but anyways this is just a hack and wasn't meant to be addressed by gsconnect, just pushed this to share this with others.
Hiding the deviecs officially can be a separate issue, closing this.

PR #1679, opened 2023-09-17
GNOME 45 port of Shell extension

### LINKED ISSUE #1665: GNOME 45 support
### Describe your request

There are breaking changes which will require a separate release for 45 and <=44:

See https://gjs.guide/extensions/upgrading/gnome-shell-45.html

### Proposed solution

_No response_

### Alternatives

_No response_

### GSConnect version

N/A

### Installed from

OS package manager

### GNOME Shell version

_No response_

_No response_

### Additional context

_No response_

retrixe:
- Migrate code over to ESM.
- Migrate prefs.js and extension.js to export needed objects.
- Drop 'use strict' from ES Modules since it's unneeded as per JS spec.
- Drop gettext and getIcon helpers on Extension.
- Drop usages of ByteArray in favour of TextDecoder.
- Drop `var` in favour of `export let`.
- Remove `__init__.js`, move its setup code to a function in `utils.js`
- Work around GNOME Shell not exporting some values to `notification.js`
- Add workarounds for config.js/remote.js to be imported from ES modules
- Use new Quick Settings API to add tile.
- Update ESLint settings to support ES modules and newer JS features
- Update _ornamentLabel name, fixes quick settings

Fixes #1665


src/extension.js
```diff
@@ -2,33 +2,34 @@
...
+import * as Clipboard from './shell/clipboard.js';
+import './config.js';
+import * as Device from './shell/device.js';
+import * as Keybindings from './shell/keybindings.js';
+import * as Notification from './shell/notification.js';
+import * as Input from './shell/input.js';
+import './utils/remote.js';
+// FIXME: As a workaround, we are using globalThis for module interop.
+const Config = globalThis.GSConnectLegacyExports.Config;
+const Remote = globalThis.GSConnectLegacyExports.Remote;
```
andyholmes, line 30: This would indeed be nice to avoid, but I doubt there will be any conflict.

src/extension.js
```diff
@@ -360,53 +354,56 @@ class ServiceIndicator extends QuickSettings.SystemIndicator {
...
-    // portal for use on Wayland
-    Clipboard.watchService();
-}
-
-
-function enable() {
-    serviceIndicator = new ServiceIndicator();
-    Notification.patchGtkNotificationSources();
+export let serviceIndicator = null;
+export let lockscreenInput = null;
```
andyholmes, line 358: I don't think these need to be exported, you can probably just make them variables of the `GSConnectExtension` class now.

andyholmes (changes requested):
Nicely done!

I didn't look too deep, since there's just a lot of maintenance to do everywhere in GSCOnnect, so just a few tweaks and I think this will be good to merge.

However, the `metadata.json` will have to be updated to `"45"` as the `shell-version`.

src/extension.js
retrixe: `serviceIndicator` is being used from another class, but I've moved `lockscreenInput` in

src/extension.js
retrixe: There are 2 alternative options I've thought of, one would be to clone `{config,remote}.js` into `{config,remote}.mjs` (`.mjs` being the Node.js extension for ES modules), and another is to enable legacy imports in the files that need `Config`/`Remote` (each file seems to need the `searchPath` modified individually). Instead of cloning the entire files, the `.mjs` files could use legacy imports to import the legacy modules and re-export them as ESM.

I've pushed 2 commits, the 1st one enables legacy imports in the individual files and uses those for imports, the 2nd one reverts that and has `.mjs` files which re-export the contents of `config.js` and `remote.js`, based on your input I can pick either of those approaches, unless you have any other idea? I think the latter approach is probably cleanest.

retrixe:
It seems I'm experiencing a major issue on Fedora 39 + GNOME 45, where my device is seemingly unable to receive information from GSConnect despite being connected. I can see current info about the phone in my GSConnect menu and receive notifications and everything else, but I can't see laptop battery info, media players, receive files from laptop, etc from the phone. These did work fine with GNOME 44 (and Valent w/ GNOME 45 port works fine too).

I'm not sure what the root cause could be, seeing that the service and preferences are separate (and using the buttons there don't work either), and there are seemingly no relevant errors or stack traces in `journalctl -xe` or the generated support log.

andyholmes:
> It seems I'm experiencing a major issue on Fedora 39 + GNOME 45, where my device is seemingly unable to receive information from GSConnect despite being connected...

I would guess this is because GSConnect can now run on multiple ports, while an older instance may bind port 1716 and disallow re-use. Short answer: make sure you kill all existing instances of GSConnect and only the version you're working on is running.

src/extension.js
andyholmes: Yeah, I'd agree. It's also probably easier to remove if/when the rest of the extension is ported to ESM.

retrixe:
This doesn't seem to be the case, I only have a single GSConnect daemon running, and the issue persists after a system restart as well :/

andyholmes:
> This doesn't seem to be the case, I only have a single GSConnect daemon running, and the issue persists after a system restart as well :/

Yeah, I can confirm this. I don't really know what's causing it, since backing up to `main` branch it seems to work fine. I guess the service may be importing something shared that's getting it confused?

retrixe:
It's not the `.mjs` files causing it, the imports are read correctly in the preferences and service from my logging, and the only imports that are shared appear to be the aforementioned files i.e. `config.js` and `remote.js`.

Nothing outside `src/shell` is importing from `src/shell`, same goes for `extension.js` and `prefs.js` (used regex `imports\.(?!gi)(?!system)(?!searchPath)(?!config)(?!utils)(?!preferences)(?!byteArray)(?!format)(?!service)(?!tweener)(?!remote)(?!gettext)(?!wl_clipboard)` to check)

Perhaps it's a change in gjs/gi libraries that affects something in the service?

andyholmes:
> Perhaps it's a change in gjs/gi libraries that affects something in the service, will try to look into it further :/

I don't believe so, I just pulled from `main` and restarted the daemon and it worked fine. Logs from kdeconnect-android may be more enlightening though.

andyholmes:
Actually, seems to work for me now. Is nothing being printed in the log that's any help?

retrixe:
Well, this is really weird, I just restarted KDE Connect on my phone and the daemon works fine now lol
Switching the daemon off and then on also reconnects properly (although there is a slight delay before the phone starts responding/receives any requests to e.g. Ring in the first place), not sure how this bug came about in the first place, but I guess this has nothing to do with this PR after all.

FWIW yes, I was using `debug` and the only logs I saw were about sending/receiving data and few unrelated errors wrt SFTP. Not sure how to fetch logs from kdeconnect-android (`adb logcat` doesn't seem to have many relevant logs), but the bug isn't occurring anymore and I'm not able to repro it anymore

andyholmes:
Okay, cool, is there anything left to resolve? I think this is good enough to get some testing. GNOME 45 is released in a few days, so it would good to get some broader testing.

andyholmes (approved):

retrixe:
Nope, everything is complete and ready to merge 👍

andyholmes:
Thanks so much for working through this!

kloczek:
Is it anything else still on outstanding list which holds new version released with NOME 45 support? 🤔

andyholmes:
#1694 needs to be fixed, probably as a part of #1683, although it could just be patched.

kloczek:
Thank you to let me know 👍

PR #1681, opened 2023-09-23
refactor: remove the "photo" plugin

andyholmes:
This plugin was deprecated and removed in all clients.

See also: https://bugs.kde.org/show_bug.cgi?id=474121


andyholmes:
Going to go ahead and merge this, since it's official upstream now.

PR #1682, opened 2023-09-23
Prepare service/preferences for ES Modules

retrixe:
- Replace all calls to byteArray module
- Replace \_\_init\_\_.js with files imported as needed
- Share common init code, lint exceptions cleanup
- Migrate WebExtension file, share common func

I haven't replaced tweener and format, we can just keep importing them via the legacy `imports` object even in ESM
I'll be pushing a PR soon which converts the service and preferences to ES Modules fully


src/extension.js
```diff
@@ -159,7 +159,7 @@ const ServiceToggle = GObject.registerClass({
                 '%d Connected',
                 '%d Connected',
                 available.length
-            ).format(available.length);
+            ).replace('%d', available.length);
```
andyholmes, line 162: I'm not the biggest fan of this. It's sort of what GJS does anyways, but I think we just suck it up and do `String.prototype.format = imports.format.format` in `setup.js` like GNOME Shell  does.

src/preferences/service.js
```diff
@@ -19,7 +19,7 @@ const Remote = imports.utils.remote;
 /*
  * Header for support logs
  */
-const LOG_HEADER = new GLib.Bytes(`
+const LOG_HEADER = new TextEncoder().encode(`
```
andyholmes, line 22:
```suggestion
const LOG_HEADER = new GLib.Bytes(`
```
This should stay `GLib.Bytes`, since it's passed to `Gio.OutputStream.write_bytes_async()`.

src/service/plugins/sms.js
```diff
@@ -85,7 +85,7 @@ var Metadata = {
  *
  * TEXT_MESSAGE: Has a "body" field which contains pure, human-readable text
  */
-var MessageEvent = {
+var MessageEventType = {
```
andyholmes, line 88: Seems unrelated?

src/shell/device.js
```diff
@@ -122,7 +122,7 @@ export const Battery = GObject.registerClass({
         if (time === 0)
             // TRANSLATORS: When no time estimate for the battery is available
             // EXAMPLE: 42% (Estimating…)
-            return _('%d%% (Estimating…)').format(level);
```
andyholmes, line 125: These don't need to be changed, since GNOME Shell includes the `String.prototype.format` override.

andyholmes (changes requested):
Looks pretty good!

src/service/plugins/sms.js
retrixe: The reason for this is that `MessageEvent` is defined in the `shared-node-browser` env

It seems gjs doesn't provide `MessageEvent`, but it's just easier and imo preferable to rename this variable instead of conflicting with a standardised API name

andyholmes (approved):
LGTM, ready to merge on your end?

retrixe:
Should be good enough, I think there are some more bits of code that could be pulled from #1683 though

Once #1683 is in a functional shape (currently it freezes on startup for some reason after applying some bug fixes locally, will debug when I'm free), I'll pull in any remaining changes not tied to ESM and re-request a review

andyholmes (approved):
LGTM as a stepping stone.

I assume `setup.mjs` and `setup.js` will converge in the next PR?

retrixe:
> I assume `setup.mjs` and `setup.js` will converge in the next PR?
Yeah, I've mostly completed the work here #1683

src/utils/setup.mjs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+import GLib from 'gi://GLib';
+
+const [filename] = GLib.filename_from_uri(import.meta.url);
+const dirname = GLib.path_get_dirname(filename);
+imports.searchPath.unshift(dirname);
+
+export const {setup, setupGettext} = imports.setup;
+export default setup;
```
ferdnyc, line 12: OOC, why does `setup.mjs` here export `setup`, but `remote.mjs` (not in this PR) export `Remote`?

Does the difference in casing have any conceptual significance? Can they be used interchangeably (case-insensitively), or does it make a practical difference? Should we settle on a standard capitalization style?

retrixe: `setup` is a function name, hence it's exported as lowercase, whereas `Remote` is a namespace, hence capitalised. This is essentially the same naming convention I see throughout the rest of the extension and most JavaScript code in general.

Since it's a default export, `Remote` isn't a name exposed to modules that import it anyway, e.g. `import remote` or `import Remote from './remote.mjs'` are both valid, dynamic import works as `(await import('./remote.mjs')).default`

IIRC, in #1683 I export the members directly since the bridge is no longer required

PR #1683, opened 2023-09-24
Migrate to ES Modules

retrixe:
This converts the service and preferences over to ES Modules as well

Remaining problems to tackle:
- [x] Ensure daemon and preferences load properly
- [x] Fix remaining dirty hacks
- [x] Tests have not been migrated

Requires #1682 to be merged first


Totto16:
any updates or help needed?

retrixe:
> any updates or help needed?

I've been a bit busy IRL, but now I'm free again to work on this PR so I'll work on tackling the remaining issues.

Currently, I'm slightly unsure how to tackle the remaining issues left ideally that are in the last commit (one is the PulseAudio `Gvc` module being unnecessarily imported in the preferences dialog due to the import tree, another is a circular import with the `Lan` channel that I've worked around but imo is not the most ideal solution...). I think some code restructuring will be necessary to fix these. Apart from that, migrating tests is its own separate piece of work as well.

retrixe (commented):
Are you sure this works correctly? I was using dynamic imports earlier and ended up reverting them in a subsequent commit because there's no mainloop at first when the imports are being resolved, which prevents top-level await from functioning: https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1683/commits/dcae46d9a77ef673b59a38641eb957fc477b2477

Also, the correct syntax should be `(await import('module')).default` to get the default export

andyholmes:
> Are you sure this works correctly? I was using dynamic imports earlier and ended up reverting them in a subsequent commit because there's no mainloop at first when the imports are being resolved, which prevents top-level await from functioning: [dcae46d](https://github.com/GSConnect/gnome-shell-extension-gsconnect/commit/dcae46d9a77ef673b59a38641eb957fc477b2477)

GJS does supports top-level await, although maybe we have to use [`Gio.Application.runAsync()`](https://gjs-docs.gnome.org/gjs/overrides.md#gio-application-runasync).
 
> Also, the correct syntax should be `(await import('module')).default` to get the default export

Ah right, sorry I forgot about the `default` bit.

This worked for me, does it for you?

```diff
index ad127cbe..2ff2e119 100755
--- a/src/gsconnect-preferences
+++ b/src/gsconnect-preferences
@@ -99,4 +99,4 @@ const Preferences = GObject.registerClass({
     }
 });
 
-(new Preferences()).run([system.programInvocationName].concat(ARGV));
+(new Preferences()).runAsync([system.programInvocationName].concat(ARGV));
diff --git a/src/service/components/contacts.js b/src/service/components/contacts.js
index 5bff3776..3070f408 100644
--- a/src/service/components/contacts.js
+++ b/src/service/components/contacts.js
@@ -14,9 +14,9 @@ let EBookContacts = null;
 let EDataServer = null;
 
 try {
-    EBook = imports.gi.EBook;
-    EBookContacts = imports.gi.EBookContacts;
-    EDataServer = imports.gi.EDataServer;
+    EBook = (await import('gi://EBook')).default;
+    EBookContacts = (await import('gi://EBookContacts')).default;
+    EDataServer = (await import('gi://EDataServer')).default;
 } catch (e) {
     HAVE_EDS = false;
 }
diff --git a/src/service/components/sound.js b/src/service/components/sound.js
index faf4ca32..f663d859 100644
--- a/src/service/components/sound.js
+++ b/src/service/components/sound.js
@@ -8,7 +8,7 @@ import GLib from 'gi://GLib';
 
 let GSound = null;
 try {
-    GSound = imports.gi.GSound;
+    GSound = (await import('gi://GSound')).default;
 } catch (e) {}
 
 
diff --git a/src/service/daemon.js b/src/service/daemon.js
index 2aa88295..7ea1de52 100755
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -696,5 +696,5 @@ const Service = GObject.registerClass({
     }
 });
 
-(new Service()).run([system.programInvocationName].concat(ARGV));
+(new Service()).runAsync([system.programInvocationName].concat(ARGV));
 
diff --git a/src/service/manager.js b/src/service/manager.js
index 9c26f220..4d7481fe 100644
--- a/src/service/manager.js
+++ b/src/service/manager.js
@@ -370,7 +370,7 @@ const Manager = GObject.registerClass({
             if (unpaired.length === 3)
                 this.discoverable = false;
 
-            device = new Device.Device(packet);
+            device = new Device(packet);
             this._exportDevice(device);
             this.devices.set(device.id, device);
```

retrixe:
I think I reverted the code earlier because top-level await returns the control flow and begins loading other modules, causing errors when using top-level await in `service/init.js` (which I just tried, it does cause issues), but in the places where your patches are used, it seems to be fine, I think

I've added the `Device.Device` fix to the PR, `runAsync` seems to be needed for the service to work properly as well when I add top-level awaits (from what I see [here](https://gitlab.gnome.org/GNOME/gjs/-/merge_requests/732), they use top-level await on `runAsync` but it works either way?)

andyholmes:
> `runAsync` seems to be needed for the service to work properly as well when I add top-level awaits (from what I see [here](https://gitlab.gnome.org/GNOME/gjs/-/merge_requests/732), they use top-level await on `runAsync` but it works either way? I've added it just in case)

Right, I believe top-level await can stall a GJS process if `runAsync()` isn't used (in some cases). Generally `GLib.MainLoop.run()` or `Gio.Application.run()` takes ownership of the default `GLib.MainContext` and responsibility for running a loop (or iterating the context), but that can conflict with other Promises.

I'm not sure how this plays out in GSConnect though, so I can't explain why things are resolving out of order wrt `init.js`. I'm also not sure why `Gvc` or `pulseaudio.js` is getting imported in the preferences? The last commit is not very elegant, but if it gets it working I guess it's fine.

src/service/init.js
```diff
@@ -2,27 +2,25 @@
...
 
 // Promise Wrappers
-try {
-    const {EBook, EDataServer} = imports.gi;
-
+import('gi://EBooks').then(EBook => {
     Gio._promisify(EBook.BookClient, 'connect');
     Gio._promisify(EBook.BookClient.prototype, 'get_view');
     Gio._promisify(EBook.BookClient.prototype, 'get_contacts');
+}).catch(() => {});
```
andyholmes, line 20:
```suggestion
import('gi://EBook').then(module => {
    const EBook = module.default;
    Gio._promisify(EBook.BookClient, 'connect');
    Gio._promisify(EBook.BookClient.prototype, 'get_view');
    Gio._promisify(EBook.BookClient.prototype, 'get_contacts');
}).catch(console.debug);
```

src/service/init.js
```diff
@@ -2,27 +2,25 @@
...
+import('gi://EDataServer').then(EDataServer => {
     Gio._promisify(EDataServer.SourceRegistry, 'new');
```
andyholmes, line 22:
```suggestion
import('gi://EDataServer').then(module => {
    const EDataServer = module.default;
    Gio._promisify(EDataServer.SourceRegistry, 'new');
```

src/service/init.js
```diff
@@ -2,27 +2,25 @@
...
-} catch (e) {
-    // Silence import errors
-}
+}).catch(() => {});
```
andyholmes, line 23:
```suggestion
}).catch(console.debug);
```

andyholmes (changes requested):
Some import fixes

retrixe:
From what I read, it's normal JavaScript behaviour, when a top-level await is encountered then control flow is returned to the importing module, which can then start loading the next module in parallel.
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await
You can use the await keyword on its own (outside of an async function) at the top level of a module. This means that modules with child modules that use await will wait for the child modules to execute before they themselves run, all while not blocking other child modules from loading.

In `init.js` if I use top-level await, then control flow goes back to `daemon.js`, which then imports `manager.js`, importing `device.js` and `plugins/index.js` in turn
Preferences imports `components/pulseaudio.js` through a dependency chain as well, `preferences/device.js` -> `plugins/index.js` -> most plugins import `components/index.js` -> `components/pulseaudio.js`

Also oops thanks for pointing out the accidental missing `default` lol

andyholmes:
Ah, so it's more like "defer" then. Well, I guess that can be cleaned up later, not very much of GSConnect was written to support modern async features so this is probably to be expected.

retrixe:
Some tests are inexplicably failing and I haven't been able to figure out why, it's because the `mockComponents` function seems to not work (properties simply disappear from the object when re-assigned...) There's also some issue with the MPRIS component test where it can't access DBUS for some reason.

It's the only blocker preventing this PR from being un-drafted, help is appreciated

retrixe:
That DBUS property is defined in setup.js file, then gets stripped later down the line somehow once the control flow returns to the test file, which makes no sense to me at all
The same happens inside the mockComponents() function

andyholmes:
Okay, so it looks like some of the paths just might not be pointing to the correct parent, like

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/3443e2767fcbfa795e8da7acc3767b1596e49c9f/installed-tests/suites/backends/testLanBackend.js#L8

Which I believe should be `../../fixtures/utils.js`?

I also notice that there are two `config.js` files in play, and I'm not sure how those interact. For example, I believe this will end up being relative to e.g. `/usr/local/share/share/gnome-shell/extensions/gsconnect@andyholmes.github.io`:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/3443e2767fcbfa795e8da7acc3767b1596e49c9f/src/service/init.js#L11

While here, in the installed-tests suite, will be passed e.g. `/usr/local/share/installed-tests/gsconnect`:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/3443e2767fcbfa795e8da7acc3767b1596e49c9f/installed-tests/minijasmine#L120

src/service/components/pulseaudio.js
```diff
@@ -2,40 +2,43 @@
...
+let Gvc = null;
+try {
+    // Add gnome-shell's typelib dir to the search path
+    const typelibDir = GLib.build_filenamev([Config.GNOME_SHELL_LIBDIR, 'gnome-shell']);
+    GIRepository.Repository.prepend_search_path(typelibDir);
+    GIRepository.Repository.prepend_library_path(typelibDir);
 
-const Gvc = imports.gi.Gvc;
+    Gvc = (await import('gi://Gvc')).default;
+} catch (e) {}
```
ferdnyc, line 22: Is it possible that having this in a try/catch, combined with `await import()`, is causing failures to resolve imports when they're needed?

src/service/components/pulseaudio.js
```diff
@@ -110,7 +113,7 @@ class Stream {
  * The Mixer class uses GNOME Shell's Gvc library to control the system volume
  * and offers a few convenience functions.
  */
-const Mixer = GObject.registerClass({
+const Mixer = !Gvc ? null : GObject.registerClass({
     GTypeName: 'GSConnectAudioMixer',
 }, class Mixer extends Gvc.MixerControl {
     _init(params) {
```
ferdnyc, line 119: Because, previously, `Mixer` would always be defined.  But now, the code can cause it to initialize as `null` if `Gvc` hasn't resolved...

src/service/components/pulseaudio.js
```diff
@@ -265,4 +268,4 @@ const Mixer = GObject.registerClass({
 /**
  * The service class for this component
  */
-var Component = Mixer;
+export default Mixer;
```
ferdnyc, line 271: Which means the default export from the module is `null`.

Which appears to be exactly what we're seeing in the tests:

```text
not ok 7 The telephony plugin can lower and restore the volume when
 the phone is ringing
# Message: Error: <spyOn> : could not find an object to spy upon for
 lowerVolume()\nUsage: spyOn(<object>, <methodName>) in
 file:///usr/libexec/installed-tests/gsconnect/jasmine.js (line 7936)
# Stack:
#   @file:///usr/libexec/installed-tests/gsconnect/plugins/testTelephonyPlugin.js:151:18
#   setTimeout/source<@resource:///org/gnome/gjs/modules/esm/_timers.js:72:9
#   _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34
# Message: TypeError: localMixer is undefined in
 file:///usr/libexec/installed-tests/gsconnect/plugins/testTelephonyPlugin.js (line 162)
# Stack:
#   @file:///usr/libexec/installed-tests/gsconnect/plugins/testTelephonyPlugin.js:162:13
#   setTimeout/source<@resource:///org/gnome/gjs/modules/esm/_timers.js:72:9
#   _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34
```

src/service/components/index.js
```diff
@@ -18,17 +41,17 @@ const Default = new Map();
...
     try {
         let info = Default.get(name);
 
         if (info === undefined) {
-            const module = imports.service.components[name];
+            const module = components[name];
 
             info = {
-                instance: new module.Component(),
+                instance: new module.default(),
```
ferdnyc, line 54: Now telephony's attempt to `acquire()` an instance of the `pulseaudio` component module (which exports its `Mixer`) as its `_mixer` object seems to be ending up with a `null`:

```js
const TelephonyPlugin = GObject.registerClass({
    GTypeName: 'GSConnectTelephonyPlugin',
}, class TelephonyPlugin extends Plugin {

    _init(device) {
        super._init(device, 'telephony');

        // Neither of these are crucial for the plugin to work
        this._mpris = Components.acquire('mpris');
        this._mixer = Components.acquire('pulseaudio');
    }
```

Or perhaps `acquire()` needs to be an `async` function?

ferdnyc:
> ~Okay, so it looks like some of the paths just might not be pointing to the correct parent, like~.
> 
> https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/3443e2767fcbfa795e8da7acc3767b1596e49c9f/installed-tests/suites/backends/testLanBackend.js#L8
> 
> ~Which I believe should be `../../fixtures/utils.js`?~ EDIT: sorry, I forgot these actually get installed.

Maybe... but if they're _**not**_ installed, it seems like this causes problems. And the tests didn't _used_ to require installation.

On the `main` branch, I can run the tests right from the build dir:

```console
$ cd /path/to/gsconnect
$ meson setup _build
$ meson compile -C _build
$ meson test -C _build
```

...everything succeeds except the clipboard test.

On this PR's branch, if I try that, all of the functionality tests abort with failures to find `installed-tests/suites/fixtures/utils.js`:

```console
$ meson test -C _build
[1/1] Generating installed-tests/prepare-tests with a custom command
[1-4/27] 🌑 gsconnect:data / Validate desktop file                               
1/27 gsconnect:data / Validate desktop file                       OK              0.01s
[2-4/27] 🌒 gsconnect:data / Validate metainfo file                             
[2-5/27] 🌓 gsconnect:data / Validate metainfo file                              
2/27 gsconnect:data / Validate desktop file (Preferences)         OK              0.01s
[3-5/27] 🌔 gsconnect:data / Validate metainfo file                             
[3-6/27] 🌕 gsconnect:data / Validate metainfo file                              
3/27 gsconnect:data / Validate metainfo file                      OK              0.04s
[4-6/27] 🌖 gsconnect:core / Device                                             
[4-7/27] 🌗 gsconnect:core / Device                                              
4/27 gsconnect:lint / Python Black (Nautilus Extension)           OK              0.32s
[5-7/27] 🌘 gsconnect:core / Device                                             
[5-8/27] 🌑 gsconnect:core / Device                                              
5/27 gsconnect:lint / Python Flake8 (Nautilus Extension)          OK              0.40s
[6-8/27] 🌒 gsconnect:core / Device                                             
[6-9/27] 🌓 gsconnect:core / Device                                              
6/27 gsconnect:core / Device                                      ERROR           0.48s   killed by signal 5 SIGTRAP
>>> GJS_PATH=.../_build/installed-tests/src NO_AT_BRIDGE=1 
GSETTINGS_BACKEND=memory
 MALLOC_PERTURB_=1 GSCONNECT_TEST=1 
 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 
 GSETTINGS_SCHEMA_DIR=.../_build/installed-tests 
 G_DEBUG=fatal-warnings,fatal-criticals 
 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 
 .../installed-tests/minijasmine 
.../_build/../installed-tests/suites/core/testDevice.js

[7-9/27] 🌔 gsconnect:core / Plugin                                             
[7-10/27] 🌕 gsconnect:core / Plugin                                             
7/27 gsconnect:core / Manager                                     ERROR           0.56s   killed by signal 5 SIGTRAP
>>> GJS_PATH=.../_build/installed-tests/src NO_AT_BRIDGE=1 
GSETTINGS_BACKEND=memory GSCONNECT_TEST=1 
GSETTINGS_SCHEMA_DIR=.../_build/installed-tests 
UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 
MALLOC_PERTURB_=128 
G_DEBUG=fatal-warnings,fatal-criticals 
ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 
.../installed-tests/minijasmine 
.../_build/../installed-tests/suites/core/testManager.js


[...]

Ok:                 5   
Expected Fail:      0   
Fail:               22  
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   

Full log written to .../_build/meson-logs/testlog.txt
```

...And in the test log:

```text
==================================== 6/27 ====================================
test:         gsconnect:core / Device
start time:   21:03:47
duration:     0.48s
result:       killed by signal 5 SIGTRAP
command:      GJS_PATH=.../_build/installed-tests/src NO_AT_BRIDGE=1 GSETTINGS_BACKEND=memory MALLOC_PERTURB_=1 GSCONNECT_TEST=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 GSETTINGS_SCHEMA_DIR=.../_build/installed-tests G_DEBUG=fatal-warnings,fatal-criticals ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 .../installed-tests/minijasmine .../_build/../installed-tests/suites/core/testDevice.js
----------------------------------- stderr -----------------------------------

(gjs:357516): Gjs-CRITICAL **: 16:03:47.918: JS ERROR:
 ImportError: Unable to load file from:
 file://.../installed-tests/suites/fixtures/utils.js
 (Error opening file .../installed-tests/suites/fixtures/utils.js:
 No such file or directory)
loadURI@resource:///org/gnome/gjs/modules/internal/loader.js:88:21
loadURI@resource:///org/gnome/gjs/modules/internal/loader.js:290:30
resolveModule@resource:///org/gnome/gjs/modules/internal/loader.js:179:33
moduleResolveHook@resource:///org/gnome/gjs/modules/internal/loader.js:339:29
```

andyholmes:
Ah, you're right:
https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/43594cc23826fdc6b17acd04b67b34338acf78d4/installed-tests/meson.build#L29-L41

That should setup tests correctly for the uninstalled case. Then each suite should setup the rest for installed tests:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/43594cc23826fdc6b17acd04b67b34338acf78d4/installed-tests/suites/core/meson.build#L28-L43

src/service/components/index.js
andyholmes: I think this is fixed by my comment [here](https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1683#issuecomment-1963095964)

andyholmes:
Considering the amount of work that went into this, and the fact the test suite is pretty dusty anyways, how do we feel about just pushing forward after a re-base?

If the tests can be run uninstalled that seems good enough, and there have been any number of updates to [jasmine-gjs](https://github.com/ptomato/jasmine-gjs) since *[checks watch]* the last Olympics that probably make ESModules more viable in tests.

daniellandau:
That sounds good to me

retrixe:
I can rebase this later tonight and look into the proposed test fixes 👍

retrixe:
Re: https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1683#issuecomment-1968099048 these config.js files have identical contents for me, I don't think we can deduplicate these because the installed tests import `config.js` to get the `PACKAGE_DATADIR` to import the rest of the files from (this also seems to be the case currently on `main` too, so I don't think it's a problem for this PR in question)

Re: running the tests uninstalled, I'm not sure how we can solve this without using dynamic imports to pull in fixtures because of how the build system seems to be placing the tests? I just tried substituting `../fixtures` with `../../fixtures` to see if that works, but I run into an issue with `config.js`.

The tests are running fine installed currently after applying https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1683#issuecomment-1963095964 ( :tada: ), only running into an issue with one of the clipboard plugin tests and the MPRIS component test suite, I'll look into these later today. With regards to landing it, I think it should easily be doable for the GNOME 46 release unless these test failures are elusive to solve, lol

andyholmes:
> The tests are running fine installed currently after applying https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1683#issuecomment-1963095964 ( 🎉 ), only running into an issue with one of the clipboard plugin tests and the MPRIS component test suite, I'll look into these later today.

If you're running this locally, I believe the clipboard test really only ever worked on X11. It's pretty much expected to fail outside of tightly controlled test environment, so no worries there. Might be the same for the MPRIS test, but it's been awhile :)

retrixe:
Nope, I'm running them inside the Docker container the CI uses with the same commands it uses

Seems like the two `config.js` files are the culprit with the MPRIS failure, the `DBUS` property is initialised on the `config.js` that sits alongside the rest of the GSConnect extension. The other tests are fine using the `config.js` in the installed-tests folder because they only pull `PACKAGE_DATADIR` and some other odd string info bits from it

I've left a FIXME comment in there because it might be a better idea to at least rename that config.js file, if not remove it

andyholmes:
The MPRIS tests are useful, because it's one the plugins associated with a specification, but honestly most players don't implement it properly anyways (I think the Rust crate floating around returns the position as a string :slightly_frowning_face:).

Since that's being poked at in #1779 and #1780, I'd be okay with marking it `skip` for now if everyone else is.

retrixe:
The MPRIS test is working now in the latest commit, but the solution is a bit ugly

https://github.com/retrixe/gnome-shell-extension-gsconnect/blob/c7b0c26e08ba5d86abd1d87b1df0cc461557de67/installed-tests/fixtures/mpris.js#L8-L18

I've pushed a commit to address this (and https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1683#issuecomment-1963075203 by extension) by having the test `config.js` copy re-export the extension's `config.js` file, let me know what you think of this approach

As for the clipboard test (`The clipboard plugin will not pull content when not allowed`), apparently it's because the component mocking function is actually broken vs `main`, which I've resolved in the latest commit, I can't think of a cleaner approach to replace the original functions

andyholmes (approved):
I think the function overrides are fine for now. Those are not hot-paths and if they were they'd get JIT-compiled out.

If @daniellandau agrees, I think this is good to merge. If not, we can do a release -> merge -> short test period -> release. The nature of the `metadata.json` `version` field is clearer now, so the [`version-name`](https://gjs.guide/extensions/overview/anatomy.html#version-name) field can be used to do point releases.

src/service/components/sound.js
```diff
@@ -2,23 +2,26 @@
...
+try {
+    GSound = (await import('gi://GSound')).default;
+} catch (e) {}
 
 
 /*
  * Used to ensure 'audible-bell' is enabled for fallback
  */
-const WM_SETTINGS = new Gio.Settings({
+/* const WM_SETTINGS = new Gio.Settings({
```
daniellandau, line 18: Ia this variable not used anywhere?

src/utils/remote.js
```diff
@@ -2,16 +2,14 @@
...
-const GObject = imports.gi.GObject;
+import Gio from 'gi://Gio';
+import GLib from 'gi://GLib';
+import GObject from 'gi://GObject';
 
 const SERVICE_NAME = 'org.gnome.Shell.Extensions.GSConnect';
 const SERVICE_PATH = '/org/gnome/Shell/Extensions/GSConnect';
 const DEVICE_NAME = 'org.gnome.Shell.Extensions.GSConnect.Device';
-const DEVICE_PATH = '/org/gnome/Shell/Extensions/GSConnect/Device';
+// const DEVICE_PATH = '/org/gnome/Shell/Extensions/GSConnect/Device';
```
daniellandau, line 12: Here too commented out code

daniellandau:
I added a couple of comments. I'd prefer merge before release: I'm pretty sure we get orders of magnitude more community testing now than after a release with 46 support.

src/service/components/sound.js
retrixe: Right, these comments are because ESLint complains about unused global functions/variables which aren't explicitly exported now since we're on ESM

I can drop all 3 of these if you want, but yeah none of these are referenced anywhere

andyholmes:
Just need those two remaining dead code bits removed, and looks good to merge :+1:

andyholmes:
Here we goooo

PR #1695, opened 2023-10-23
shell/utils: Fix Gio bitfield value

ferdnyc:
I got the name of a bitfield member wrong, when using `Gio.File.query_info`: the `FileQueryInfoFlags` member is named `NOFOLLOW_SYMLINKS`, not `NO_FOLLOW_SYMLINKS`.

I'm not 100% sure this makes any difference. (Most likely, it just means the calls are made with flags of `0` (`NONE`) instead. Which is fine, since the files being examined _aren't_ symlinks.) But, worst case scenario is that the calls fail and the permissions on the target files aren't fixed. So, better to use the correct name.


andyholmes (approved):
LGTM!

PR #1699, opened 2023-10-29
Quick fixes for GNOME 45 compatibility

daniellandau:
I couldn't get the code in https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1683 to a state where the main UI would open, with the most blocking part being Gvc, which seems to have been originally difficult as well https://stackoverflow.com/questions/46674935/is-it-possible-to-import-the-gvc-typelib-outside-of-the-gnome-shell-environment

With this small change at least some things work and we could maybe put out a release, though for example remote input from phone to computer doesn't yet.

### DIFF
--- a/src/service/components/input.js
+++ b/src/service/components/input.js
@@ -381,29 +381,6 @@ class Controller {
         return reply.deepUnpack()[0];
     }
 
-    async _createScreenCastSession(sessionId) {
-        if (this.connection === null)
-            throw new Error('No DBus connection');
-
-        const options = new GLib.Variant('(a{sv})', [{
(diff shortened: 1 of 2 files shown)


daniellandau:
I couldn't see any adverse effects from removing the screencastsession, and remote input started working after removing that.

retrixe:
#1683 should work ootb barring the newest commit which stops the service from working
The changes in this PR are needed there too to fix the optional dependency problem so :+1:

andyholmes:
LGTM

andyholmes (approved):

PR #1700, opened 2023-10-29
Fixed Import Error

Hennsie:
Fix an error where the service isn't starting. The reason for this was that the try...finally block wasn't always catching the exception. An essential catch has been added. Now the service is running and everything works fine.


retrixe:
Er, #1699 addresses this

PR #1701, opened 2023-10-29
fix: cleanup GFile symlink queries

### LINKED ISSUE #1698: Question about daemon.js usage
### Describe the bug

I tested the daemon script by running: `./deamon.js --ring -d [id]` which works!

However, `./deamon.js --notification=hello -d [id]` does not. Both commands do emit an error.

```
(gjs:625179): GLib-GIO-CRITICAL **: 00:16:19.046: GFileInfo created without standard::symlink-target

(gjs:625179): GLib-GIO-CRITICAL **: 00:16:19.046: file ../gio/gfileinfo.c: line 2062 (g_file_info_get_symlink_target): should not be reached
```

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### GSConnect version

nightly

### Installed from

GitHub releases (zip file)

### GNOME Shell version

45

Fedora 39

### Paired device(s)

Pixel 7

### KDE Connect app version

1.29.0

### Plugin(s)

_No response_

### Support log

_No response_

### Screenshots

_No response_

### Notes

_No response_

andyholmes:
Cleanup the symlink bootstrapping in the initialization of `daemon.js` and `gsconnect-preferences`.

Changes to the way GFileInfo sattributes are set has resulted in errant criticals if some attributes are not checked before accessed.

closes #1698

### DIFF
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -23,15 +23,10 @@ const Gtk = imports.gi.Gtk;
 
 // Bootstrap
 function get_datadir() {
-    const m = /@([^:]+):\d+/.exec(new Error().stack.split('\n')[1]);
-    let path = null;
-    const symbolic_link = Gio.File.new_for_path(m[1])
-        .query_info('standard::*', null, null)
-        .get_symlink_target();
-    if (symbolic_link)
-        path = symbolic_link;
-    else
-        path = m[1];
+    let [, path] = /@([^:]+):\d+/.exec(new Error().stack.split('\n')[1]);
+    const info = Gio.File.new_for_path(path)
+        .query_info('standard::*', Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS, null);
+    path = info.get_is_symlink() ? info.get_symlink_target() : path;
 
     return Gio.File.new_for_path(path).get_parent().get_parent().get_path();
 }
--- a/src/gsconnect-preferences
+++ b/src/gsconnect-preferences
@@ -24,15 +24,10 @@ const Gtk = imports.gi.Gtk;
 
 // Bootstrap
 function get_datadir() {
-    const m = /@([^:]+):\d+/.exec(new Error().stack.split('\n')[1]);
-    let path = null;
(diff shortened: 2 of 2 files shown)


src/service/daemon.js
```diff
@@ -23,15 +23,10 @@ const Gtk = imports.gi.Gtk;
...
-    if (symbolic_link)
-        path = symbolic_link;
-    else
-        path = m[1];
+    let [, path] = /@([^:]+):\d+/.exec(new Error().stack.split('\n')[1]);
+    const info = Gio.File.new_for_path(path)
+        .query_info('standard::*', Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS, null);
+    path = info.get_is_symlink() ? info.get_symlink_target() : path;
 
     return Gio.File.new_for_path(path).get_parent().get_parent().get_path();
```
ferdnyc, line 31: Doubt it's a big deal either way, but you could save yourself a `Gio.File` instantiation in the (presumably-typical) non-symlink situation, this way:

```suggestion
    const [, path] = /@([^:]+):\d+/.exec(new Error().stack.split('\n')[1]);
    let file = Gio.File.new_for_path(path);
    const info = file.query_info(
        'standard::*', Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS, null);
    if (info.get_is_symlink()) {
        file = Gio.File.new_for_path(info.get_symlink_target());
    }
    return file.get_parent().get_parent().get_path();
```

ferdnyc (approved):
Makes sense to me!

The Gio shenanigans sort of pale in comparison to the fact that the code is _parsing source code paths out of stacktrace listings_, of course. Hooray for JavaScript. If anything was ever screaming out for an API to be exposed...

But, since it's not... `¯\_(ツ)_/¯`

(Is `imports.system.programPath` no help at all, here? Seems like the closest thing GJS has to exposing any sort of filesystem/runtime API.)

andyholmes:
Can't be bothered to fix the linter error, so I just force pushed :P

Issue #688, opened 2019-11-02
Remote Control: Scroll Down doesn't work


0x656b694d:
**Describe the bug**

The Scroll Down events are not handled with the remote control touch pad.

**Steps To Reproduce:**

1. Scroll Down in the remote control
4. See it doesn't scroll

**Expected behavior**

Should scroll

**Support Log**

The log generated while scrolling up, then scrolling down, then scrolling up again.

[gsconnect.log.txt](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/3800812/gsconnect.log.txt)

**System Details (please complete the following information):**

 - **GSConnect version:** 28
   - **Installed from:** GNOME Extensions Website
 - **GNOME/Shell version:** 3.34
 - **Distro/Release:** Ubuntu 19.10

**GSConnect environment (if applicable):**

 - **Paired Device(s):** OnePlus 3T
 - **KDE Connect app version:** 1.13.2
 - **Plugin(s):** mouse

**Additional Notes:**

Scroll Up works.

andyholmes:
Hmm, I can't reproduce this in Fedora 31. It might be a bug in Ubuntu 19.10. Does this work in Wayland or is it only broken in X11?

0x656b694d:
Remote mouse doesn't work at all in Wayland. Please find the log attached. It says it is falling back to Atspi in the end, whatever it means:
```
ноя 02 20:35:48 org.gnome.Shell.Extensions.GSConnect[12539]: [/service/components/input.js:_ensureAdapter:419]: "Falling back to Atspi"
ноя 02 20:35:49 org.gnome.Shell.Extensions.GSConnect[2692]: dbus[12539]: arguments to dbus_message_new_method_call() were incorrect, assertion "_dbus_check_is_valid_path (path)" failed in file ../../../dbus/dbus-message.c line 1366.
ноя 02 20:35:49 org.gnome.Shell.Extensions.GSConnect[2692]: This is normally a bug in some application using the D-Bus library.

```
[gsconnect.log.txt](https://github.com/andyholmes/gnome-shell-extension-gsconnect/files/3800883/gsconnect.log.txt)

andyholmes:
Hmm, that's a little hairy :S 

Atspi is the older, accessibility interface we fall back to if Mutter's RemoteDesktop is not available. Atspi pretty much just doesn't work at all on Wayland, but it shouldn't be throwing errors like that either.

[...]

PR #1724, opened 2023-12-28
fix: re-enable remote input on Ubuntu w/Wayland

Rexogamer:
Remote Input was [disabled](https://github.com/GSConnect/gnome-shell-extension-gsconnect/commit/5e98fcf79dad73ea57ee98ef2d7c88316abf370d) on Ubuntu sessions using Wayland a while back as Ubuntu and Debian disabled the `remote-desktop` flag ([1](https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/688#issuecomment-550476892), [2](https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/822#issuecomment-616723699)).

However, looking at the commits after the second comment, [they turned it back on](https://salsa.debian.org/gnome-team/mutter/-/commit/aefb8326692e0e44af18e9f21dbfdc8b9bb41081) for all Linux builds in early 2021 and haven't touched it since. I've tested Remote Input on Ubuntu 23.10 with this patch and everything seems to be working.

### DIFF
--- a/src/service/components/input.js
+++ b/src/service/components/input.js
@@ -147,18 +147,7 @@ const RemoteSession = GObject.registerClass({
     }
 
     scrollPointer(dx, dy) {
-        // NOTE: NotifyPointerAxis only seems to work on Wayland, but maybe
-        //       NotifyPointerAxisDiscrete is the better choice anyways
-        if (HAVE_WAYLAND) {
-            this._call(
-                'NotifyPointerAxis',
-                GLib.Variant.new('(ddu)', [dx, dy, 0])
-            );
-            this._call(
-                'NotifyPointerAxis',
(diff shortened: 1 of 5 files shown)


src/service/components/input.js
```diff
@@ -276,8 +276,6 @@ class Controller {
      */
     _checkWayland() {
```
Rexogamer, line 277: This check was added in the initial disabling commit - should it be removed too, or is it still needed?

andyholmes: Yeah, this whole function and all instances of `HAVE_WAYLAND` can go, actually.

Rexogamer: Should be done in 196166e

andyholmes (approved):

andyholmes:
Thanks!

PR #1725, opened 2024-01-02
File manager extensions: Collapse single-item submenu

ferdnyc:
Nautilus since v43 uses Gtk4 for its file views, and the context menus are implemented as popovers instead of "real" traditional context menus. One functionality difference is, submenus don't open on hover, they have to be clicked to enter.

That makes single-item submenus particularly tedious to deal with, as clicking "Send to Mobile Device" just to reveal a single choice "\<device_name>" feels an like unnecessary complication.

So, this PR adds logic that checks the count of devices available as "Send to Mobile Device" targets, and when there's exactly 1 such target, it collapses **Send to Mobile Device > \<device_name>** into just a top-level **Send to \<device_name>** entry, eliminating the unnecessary extra click while still only adding a single entry to the top-level context menu.

### Translations

Because this is a user-facing string and needs to be translated, the `.pot` translation master is also updated. Currently this _doesn't_ add a new translation string, because it shares the `"Send to %s"` entry with the contacts plugin (where it's used with a phone number, e.g. **Send to 555-5555**).

I somewhat question whether "Send to \<phone#>" and "Send to \<device_name>" CAN be translated the same in all languages. If they _can't_, the two strings may need to be tweaked to be different, or have context added to differentiate them. But for the moment, I'm content to take advantage of the overlap to avoid adding additional translator work.

### DIFF
--- a/nautilus-extension/nautilus-gsconnect.py
+++ b/nautilus-extension/nautilus-gsconnect.py
@@ -177,24 +177,36 @@ class GSConnectShareExtension(GObject.Object, FileManager.MenuProvider):
         if not devices:
             return ()
 
-        # Context Menu Item
-        menu = FileManager.MenuItem(
-            name="GSConnectShareExtension::Devices",
-            label=_("Send To Mobile Device"),
-        )
+        # If there's exactly 1 device, no submenu
+        if len(devices) == 1:
+            name, action_group = devices[0]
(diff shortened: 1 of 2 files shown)


andyholmes (approved):
LGTM, nice little UX enhancement!

PR #1767, opened 2024-03-26
Gnome 46 support

### LINKED ISSUE #1764: Gnome 46 support
### Describe the bug

With the update of the gnome-shell the extension broke. So far I have found that `MessageTray.NotificationBanner` got removed here https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3173 not sure yet what the replacement is or I would fix it myself.

### GSConnect version

56

### Installed from

GNOME Extensions website

### GNOME Shell version

46

Arch Linux

ymshenyu:
In https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/940b658071047fab8d1db18f301c59095b384f79, gnome maintainers decided to use `Calendar.NotificationMessage` for widget consistent.

closes #1764

### DIFF
--- a/data/metadata.json.in
+++ b/data/metadata.json.in
@@ -3,6 +3,6 @@
     "name": "GSConnect",
     "description": "GSConnect is a complete implementation of KDE Connect especially for GNOME Shell with Nautilus, Chrome and Firefox integration. It does not rely on the KDE Connect desktop application and will not work with it installed.\n\nKDE Connect allows devices to securely share content like notifications or files and other features like SMS messaging and remote control. The KDE Connect team has applications for Linux, BSD, Android, Sailfish, iOS, macOS and Windows.\n\nPlease report issues on Github!",
     "version": @PACKAGE_VERSION@,
-    "shell-version": [ "45", "46" ],
+    "shell-version": [ "46" ],
     "url": "@PACKAGE_URL@/wiki"
 }
--- a/src/shell/notification.js
+++ b/src/shell/notification.js
@@ -47,9 +47,8 @@ const NotificationBanner = GObject.registerClass({
     GTypeName: 'GSConnectNotificationBanner',
 }, class NotificationBanner extends Calendar.NotificationMessage {
 
-    _init(notification) {
-        super._init(notification);
-
+    constructor(notification) {
+        super(notification);
         if (notification.requestReplyId !== undefined)
             this._addReplyAction();
     }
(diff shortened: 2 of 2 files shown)


rpede (approved):

andyholmes (approved):
LGTM, thanks!

ferdnyc:
> In https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/940b658071047fab8d1db18f301c59095b384f79, gnome maintainers decided to use `Calendar.NotificationMessage` for widget consistent.

They also massively changed the style (this was all part of [MR !3173](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3173) which is 30 commits); I'm a bit concerned our styles won't match anymore since we're creating things ourselves, though it looks like we're _mostly_ unaffected.

Except: It [looks like](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/calendar.js?ref_type=heads#L819-836) `Calendar.NotificationMessage` buttons are still styled `notification-button`, same as before, so that's good. But the enclosing `ActionArea` container (an `St.BoxLayout`), is styled `notifications-buttons-bin`. We're styling ours `notification-actions`. `notification-actions` [was renamed](https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/d44adf3b781601c9a1b5d040e5c8ad2f864ab446) to `notification-buttons-bin` in the same PR; there's no longer a `notification-actions` style class AFAICT.

----

Presumably this change was made the way it was because `MessageTray.NotificationBanner` was considered an implementation detail. (Though it still seems weird that they didn't migrate the `Calendar` widget to `MessageTray` and have `Calendar` import _that_, instead of leaving it under `Calendar`.) ...Callers are supposed to be using [`MessageTray.Notification`](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/messageTray.js?ref_type=heads#L352) (I think).

We can't use `MessageTray.Notification`, since it doesn't provide the `.setActionArea()` method we need to implement interactive responses. That comes from `MessageList.Message`, which `Calendar.NotificationMessage` extends just like `MessageTray.NotificationBanner` used to.

So I guess we'll have to keep tracking changes made to internal APIs, which isn't fun but probably the best option.

andyholmes:
Yeah, there's a new, long overdue notification spec, with the implementation in-progress by Julian Sparber. I expect he's trying not to get side-tracked, but a lot of the notification inheritance chain is pretty eccentric (probably for historical reasons).

JalilArfaoui:
Hi ! 
Can we have a new release with this ?

ferdnyc:
> Hi ! 
> Can we have a new release with this ?

We're still finding additional changes to be made for full GNOME 46 support, and our build systems won't be testing and packaging the extension under GNOME 46 (a good sanity-check for compatibility) until my WIP PR is merged, so it's probably a bit too early for a new release. But we're making good progress.

PR #1771, opened 2024-03-27
GNOME 46 fix - NotificationBanner

### LINKED ISSUE #1768: Please Support Gnome 46.*
### Describe your request

Please support Gnome 46

### Proposed solution

_No response_

### Alternatives

_No response_

### GSConnect version

probably latest, cannot launch version (disabled / outdated)

### Installed from

GNOME Extensions website

### GNOME Shell version

GNOME 46.0

openSUSE Tumbleweed x86_64 

### Additional context

cannot upgrade gsconnect through extension website

### LINKED ISSUE #1770: Notifications synced from the mobile device isn't received on GNOME Shell
### Describe the bug

The GNOME 46 update cannot sync the notifications properly - The notifications are sent from the GNOME PC to the mobile device, but the mobile device cannot post notifications on the GNOME PC.

### Steps to reproduce

1. Install GSConnect from the Git repository.
2. Pair the mobile device with GSConnect using KDE Connect app.
3. Post a test notification on both devices.
4. Observe the cross-device notifications.

### Expected behavior

The notifications are synced from/to the GNOME PC.

### GSConnect version

56

### Installed from

other source

### GNOME Shell version

46

Arch

### Paired device(s)

WayDroid x86_64 Device, Redmi Note 10S

### KDE Connect app version

1.29.0

### Plugin(s)

N/A

### Support log

```
Mar 27 14:21:54 gnome-shell[8453]: JS ERROR: TypeError: notificationParams is undefined
                                   addNotification@file:///home/stella/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/shell/notification.js:232:13
                                   AddNotificationAsync@resource:///org/gnome/shell/ui/notificationDaemon.js:665:16
                                   _handleMethodCall@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:373:35
                                   _wrapJSObject/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:408:34
                                   @resource:///org/gnome/shell/ui/init.js:21:20
```

### Screenshots

N/A

### Notes

N/A

Skythrew:
This pull request intends to fix the notification implementation for GNOME 46, with the new change from `MessageTray.NotificationBanner` to `Calendar.NotificationMessage`.

Tested on my computer of course, everything seems to be working pretty well :+1:

Closes #1768 
Closes #1770

### DIFF
--- a/src/shell/notification.js
+++ b/src/shell/notification.js
@@ -367,8 +367,6 @@ export function unpatchGtkNotificationDaemon() {
 const _addNotification = NotificationDaemon.GtkNotificationDaemonAppSource.prototype.addNotification;
 
 export function patchGtkNotificationSources() {
-    const addNotification = Source.prototype.addNotification;
-
     // eslint-disable-next-line func-style
     const _withdrawGSConnectNotification = function (id, notification, reason) {
         if (reason !== MessageTray.NotificationDestroyedReason.DISMISSED)
@@ -410,7 +408,6 @@ export function patchGtkNotificationSources() {
         );
     };
 
-    NotificationDaemon.GtkNotificationDaemonAppSource.prototype.addNotification = addNotification;
     NotificationDaemon.GtkNotificationDaemonAppSource.prototype._withdrawGSConnectNotification = _withdrawGSConnectNotification;
 }
 
--- a/src/shell/notification.js
+++ b/src/shell/notification.js
@@ -209,11 +209,12 @@ const Source = GObject.registerClass({
     /*
      * Override to control notification spawning
      */
-    addNotification(notificationId, notificationParams, showBanner) {
+    addNotification(notification) {
         this._notificationPending = true;
 
         // Parse the id to determine if it's a repliable notification, device
         // notification or a regular local notification
+        let notificationId = notification.id;
         let idMatch, deviceId, requestReplyId, remoteId, localId;
 
         if ((idMatch = REPLY_REGEX.exec(notificationId))) {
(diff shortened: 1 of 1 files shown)


src/shell/notification.js
```diff
@@ -378,28 +367,7 @@ export function unpatchGtkNotificationDaemon() {
...
-        this._notifications[notificationId] = notification;
-
-        if (showBanner)
-            this.showNotification(notification);
-        else
-            this.pushNotification(notification);
-
-        this._notificationPending = false;
-    };
+    const addNotification = Source.prototype.addNotification;
```
andyholmes, line 371: This can also be removed, since it's no-op now.

src/shell/notification.js
```diff
@@ -229,84 +230,72 @@ const Source = GObject.registerClass({
...
-        if (notificationParams.icon) {
+        /* if (notificationParams.icon) {
             let gicon = Gio.Icon.deserialize(notificationParams.icon);
 
             if (gicon instanceof Gio.ThemedIcon) {
                 gicon = getIcon(gicon.names[0]);
                 notificationParams.icon = gicon.serialize();
             }
-        }
+        } */
```
andyholmes, line 240: This should just be removed if it's not necessary or fixable.

andyholmes (changes requested):
Thanks, just a few fix-ups and it LGTM!

src/shell/notification.js
Skythrew: Indeed, it's the legacy of my first debugging hours :smile: Just removed it with the latest commit :+1:

src/shell/notification.js
Skythrew: Just reimplemented the fix with the new API too :+1:

src/shell/notification.js
```diff
@@ -209,11 +209,12 @@ const Source = GObject.registerClass({
     /*
      * Override to control notification spawning
      */
-    addNotification(notificationId, notificationParams, showBanner) {
+    addNotification(notification) {
         this._notificationPending = true;
 
         // Parse the id to determine if it's a repliable notification, device
         // notification or a regular local notification
+        let notificationId = notification.id;
```
andyholmes, line 217:
```suggestion
        const notificationId = notification.id;
```

src/shell/notification.js
```diff
@@ -229,84 +230,72 @@ const Source = GObject.registerClass({
...
-
-        this._notificationPending = false;
-    }
-
-    /*
-     * Override to raise the usual notification limit (3)
-     */
-    pushNotification(notification) {
-        if (this.notifications.includes(notification))
+
```
andyholmes:
```suggestion

```

Skythrew: Fixed :+1:

Skythrew: Not really used to work with eslint :sweat_smile:

andyholmes (approved):

andyholmes:
Thanks!

andyholmes:
cc @jbicha, in case Ubuntu is carrying the same revision of notification code.

ymshenyu:
@Skythrew Oh, I didn't realize there was a TypeError.

src/shell/notification.js
```diff
@@ -229,84 +230,72 @@ const Source = GObject.registerClass({
...
             });
 
-            this._notifications[localId] = notification;
+            this._notifications[localId] = cachedNotification;
 
         // Service Notification
         } else {
-            notification = this._createNotification(notificationParams);
-            notification.connect('destroy', (notification, reason) => {
+            cachedNotification.connect('destroy', (notification, reason) => {
```
DaPigGuy, line 281: Shouldn't this be `notification`? `cachedNotification` is undefined in this branch

ferdnyc: I think you're probably right, but TBH that function is way too complex and should probably be like 3 functions.

The move to relying on `cachedNotification` (which may or may not be defined) instead of `notification` (which will always be defined) in so much of the code strikes me as needlessly confusing TBH.

I can see using `cachedNotification` in the first branch (the one taken if it's defined), sure. But since the other branches are _only_ followed when it's **not** defined, why reference it at all? Especially if it's necessary to first set it equal to `notification`. Just use `notification`.

andyholmes: Ah yes, sorry I missed that in review. I'm a bit surprised ESLint didn't catch it though; that should certainly be fixed.

ferdnyc: Never mind, I should've read farther. `cachedNotification` is used because it's needed on lines 287-298.

DaPigGuy: > I think you're probably right, but TBH that function is way too complex and should probably be like 3 functions.

Agreed. For example, already in this patch, `_notificationPending` is no longer correctedly reset in some cases due to the guard statement for existing notification, which from what I can tell is an attempt at reimplementing [`MessageTray.Source.addNotification`](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/messageTray.js#L555)

PR #1776, opened 2024-04-01
Updated  Greek Translation

Pyrofanis:
Hi i updated the greek translation
it needed less work that i thought for the time passed but anyway
if any problem occurs contact me

Best Regards 
Nasos

### DIFF
--- a/po/el.po
+++ b/po/el.po
@@ -5,8 +5,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: org.gnome.Shell.Extensions.GSConnect\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-03-22 10:01+0530\n"
-"PO-Revision-Date: 2023-03-24 11:53+0200\n"
+"POT-Creation-Date: 2024-04-01 04:14+0300\n"
+"PO-Revision-Date: 2024-04-01 04:16+0300\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: el\n"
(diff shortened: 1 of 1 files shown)


po/el.po
```diff
@@ -1076,6 +1065,12 @@ msgstr "Αποστολή \"%s\" σε %s"
 msgid "Sent “%s” to %s"
 msgstr "Εστάλη το \"%s\" στο %s"
 
+#. TRANSLATORS: eg. Failed to send "book.pdf" to Google Pixel
+#: src/service/plugins/share.js:321
+#, javascript-format
+msgid "Failed to send “%s” to %s"
+msgstr "Απέτυχε η αποστολή του \"%s\" στο %s"
```
ferdnyc, line 1072: @andyholmes Went to great lengths, a bunch of years ago, to "smart-quotes-ify" the translations (e3c87650726011c8f36eb3d5d5a50a29201deafc). 

While I realize that...
1. we didn't have Greek translations, at the time, and
2. escaped quotes appeared in the previous version of `el.po` as well, so it's not like it's something you introduced 

...I'm just wondering if there's some form of valid quoting in Greek that uses non-ASCII quotation marks? (Whether it's “foo”, „foo”, 「foo」, «foo», or something else?)

If not, then escaping we will go!

ferdnyc (approved):
My question about the escaped quotes is sort of tangential, this looks good to me regardless. Thanks, @Pyrofanis!

po/el.po
Pyrofanis: well the main differences i can pin point from my mind right now is that more commonly

for quotation marks we <<Κατι ειναι εδω >> but i have found and articles that contain ¨ ¨ 
there might not be ascii since these are  in the Greek keyboard layout and are located where : is

also in case you might find it in Greek we use ; as a question mark and ? doesn't exist

in texts where there is a dialog or you want to give more information about something we use : but i think that this is common in English also  
if you have any other questions feel free to ask  :)

andyholmes: Non-ASCII is totally fine for translations, so I think whatever Greek users expect is the best choice. I'd defer to your judgement as a translator for what's best :+1:

andyholmes:
Alrighty, merging as-is for for now so it gets in for v57, but feel free to submit any amendments you see fit :)

Issue #1779, opened 2024-04-03
Rewind/forward buttons on mobile cause the file playing to restart/finish


niels0n:
### Describe the bug

When playing a file on computer, pressing the rewind/forward button on mobile app should go 5 seconds back/forward.
Instead, pressing the rewind button cause the file to restart and pressing the forward one makes the player stop, or play the next file.

I'm trying to use it with mpv and with kdeconnect was working fine, so I guess it is something related to gsconnect.

### Steps to reproduce

1. Play a file on computer
2. Press rewind/forward button on KDE Connect mobile

### Expected behavior

The player going 5 seconds back/forward.

### GSConnect version

50

### Installed from

OS package manager

### GNOME Shell version

42.9

### Linux distribution/release

Pop!_OS 22.04 LTS

### Paired device(s)

Moto G32

### KDE Connect app version

1.29.0

andyholmes:
Sorry, but we don't support old versions of GNOME Shell.

Please reach out to your distribution for support, or even better, ask if they would consider sponsoring a developer with some of the income they get for LTS subscriptions.

ferdnyc:
Hmmm. With GSConnect 56, controlling VLC, I can confirm that pressing the "Rewind" button restarts the file from the beginning, same as "Skip Backwards".

Pressing "Fast Forward" seems to skip forward approximately 1 second, not 5, but otherwise works as expected.

(There is, of course, also the position slider to perform seeking within the current file. But the rewind/ff buttons should ideally work properly as well.)

AFAICT, Rewind sends:

```text
{
    "id": 1712175014041,
    "type": "kdeconnect.mpris.request",
    "body": {
        "player": "VLC media player",
        "Seek": -10000000
    }
}
```

and FF sends `"Seek": 10000000`. (Which actually should be a 10-second jump, by default. Tho that's configurable in KDEConnect Android's plugin settings.)

It looks like we're interpreting both of those wrong.

ferdnyc:
Playing with VLC's MPRIS interface in `d-spy`, sending it `Seek((10000000,))` skips 10 seconds forward, and `Seek((-10000000,))` skips 10 seconds backwards. So if nothing else, the fact that we're multiplying/dividing offset and position values appears to be wrong:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/9d274631033cf627580e274f953e8b03b55a9c08/src/service/plugins/mpris.js#L244-L250

[...]

PR #1783, opened 2024-04-06
Fix MPRIS seeking and SetPosition

ferdnyc:
Correct the math we use to perform seek operations via MPRIS, since KDEConnect uses microseconds for seek offsets even though they use milliseconds for position values. (MPRIS uses microseconds everywhere.)

Also, implement `SetPosition()` directly instead of computing the offset from the current position and calling `Seek()`, because some players don't respect arbitrary `Seek()` lengths.

Picking up the currently-playing `trackid` at the time the `SetPosition()` call is handled and assuming it's the same track it was intended for is somewhat risky, but picking up the current `Position` and making the same assumption (which is what the previous code did) carried the exact same amount of risk.

Fixes: #1779

### DIFF
--- a/src/service/plugins/mpris.js
+++ b/src/service/plugins/mpris.js
@@ -242,11 +242,22 @@ var Plugin = GObject.registerClass({
                 player.Volume = packet.body.setVolume / 100;
 
             if (packet.body.hasOwnProperty('Seek'))
-                await player.Seek(packet.body.Seek * 1000);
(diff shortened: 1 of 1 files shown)


andyholmes (approved):
LGTM, nice investigation on Chrome weirdness and bonus points for the comment.

ferdnyc:
Might as well get this one in too, before we start making release noises.

Issue #1610, opened 2023-04-25
Not able to access the mounted drive


mspatil32:
### Describe the bug

Hi,

I think the app in phone got updated recently, and I'm getting the below error when trying to access the mounted drive in linux.

Please prove solutions.

![image](https://user-images.githubusercontent.com/105443715/234292939-5e0a9c99-908e-40fb-b681-f85d59938141.png)

### GSConnect version

55

### Installed from

GNOME Extensions website

f-fouad:
You don't have permissions to access `/` of your phone.

Press `Ctrl+L` and append to the address `/storage/emulated/0` to access Android Internal storage. It should be the default.

mspatil32:
Hi @f-fouad ,

Thanks for your reply.

Please tell me where should I add the address "/storage/emulated/0".

This issue came up after the "kdeconnect" app got updated yesterday in android. As a work around, I installed the older version of the app (from Jan 2023), and it is working fine.

seanms:
I'm seeing the same problem after upgrading kdeconnect on my phone.

@mspatil32 When you're on the Nautlius page giving you the error, pressing Ctrl+L opens the address bar at the top.  Append "/storage..." to the address.

mspatil32:
@seanms 
Oh okay, understood now. Does this help? Is the entire phone's filesystem accessible by this method? Previously, the Downloads folders in the internal storage was not accessible.

seanms:
Yes, this appears to work around the problem.  I'm able to read and write the phone's file system.

mspatil32:
Hi @seanms 
Thanks a lot. Will try it.
Currently, I installed the older version of the app (Jan 2023), and it doesn't give this problem, and I can access some of the folders in the phone (not the main internal storage and the Downloads folder).

alexanderadam:
I'm seeing the same with recent versions and appending `/storage/emulated/0` helps indeed.
@andyholmes would it be possible to optionally this path via settings automatically?

andyholmes:
No, it's the responsibility of the remote device to report the correct path. This doesn't relate to any settings in GSConnect.

andyholmes:
Correct, the remote device must communicate the path. I haven't seen such a bug report yet, nope :/

mspatil32:
Hi
Will there be an update to kde connect app to solve this issue?

andyholmes:
You will have to open an upstream bug at https://bugs.kde.org

[...]

PR #1793, opened 2024-04-30
Use provided SFTP paths if enumeration fails

mavit:
In https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1610#issuecomment-1569096254, Andy Holmes wrote:

> So in fact GSConnect generally ignores the path sent by the Android device and tries to list the directories itself.
> 
> This could be fixed by either reverting that patch, or possibly falling back to the advertised paths if the listing fails. Patches for either would be okay by me.

This is that patch, to fall back to the advertised paths.

### DIFF
--- a/src/service/plugins/sftp.js
+++ b/src/service/plugins/sftp.js
@@ -150,22 +150,35 @@ const SFTPPlugin = GObject.registerClass({
 
     async _listDirectories(mount) {
         const file = mount.get_root();
-
-        const iter = await file.enumerate_children_async(
-            Gio.FILE_ATTRIBUTE_STANDARD_NAME,
-            Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
-            GLib.PRIORITY_DEFAULT,
-            this.cancellable);
-
-        const infos = await iter.next_files_async(MAX_MOUNT_DIRS,
-            GLib.PRIORITY_DEFAULT, this.cancellable);
-        iter.close_async(GLib.PRIORITY_DEFAULT, null, null);
-
         const directories = {};
 
-        for (const info of infos) {
-            const name = info.get_name();
-            directories[name] = `${file.get_uri()}${name}/`;
+        try {
+            const iter = await file.enumerate_children_async(
+                Gio.FILE_ATTRIBUTE_STANDARD_NAME,
+                Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
+                GLib.PRIORITY_DEFAULT,
+                this.cancellable);
+
+            const infos = await iter.next_files_async(MAX_MOUNT_DIRS,
+                GLib.PRIORITY_DEFAULT, this.cancellable);
+            iter.close_async(GLib.PRIORITY_DEFAULT, null, null);
+
+            for (const info of infos) {
+                const name = info.get_name();
+                directories[name] = `${file.get_uri()}${name}/`;
+            }
+        }
+        catch (e) {
+            debug(e, this.device.name);
+
+            // If no directories were found, fall back to using 'multiPaths'
+            if (this._mount_body.hasOwnProperty('multiPaths')) {
+                for (let i = 0; i < this._mount_body.multiPaths.length; i++) {
+                    let name = this._mount_body.pathNames[i];
+                    let path = this._mount_body.multiPaths[i];
+                    directories[name] = `${file.get_uri()}${path}/`;
+                }
+            }
         }
 
         return directories;
@@ -206,6 +219,7 @@ const SFTPPlugin = GObject.registerClass({
                 return;
 
             this._mounting = true;
+            this._mount_body = packet.body
 
             // Ensure the private key is in the keyring
             await this._addPrivateKey();


ferdnyc:
@mavit 

I suspect the change in #1801 might actually be more on track here — previously, GSConnect always tried to mount `sftp://user:password@a.n.i.p:port/` and get a list of the  contents. But since the android app switched over to `MANAGE_EXTERNAL_PERMISSIONS`, the `packet.body.path` won't always be `/` (previously, it _always_ was), and more importantly it appears `/` won't be accessible to us from the SFTP server — it publishes the filesystem using its on-device path, now.

So, if `packet.body.path` is something other than `/`, we should use that in the initial connection and stop trying to access the root of the SFTP filesystem.

The impression I get is that, _even if_ multiple shared locations are configured (which doesn't appear to be possible on Android 11+ anymore, the preferences for doing that are totally gone in the Android app)... but _even if_ multiple locations are configured, they're meant to be accessed directly and separately, not by mounting and listing `/`. The KDEConnect Android code makes me strongly suspect that's not supported (and maybe never really was).

mavit:
This versus #1801 is swings and roundabouts, I think.  If we mount `sftp://${host}:${packet.body.port}/${packet.body.path}/` then then it'll be nicely browsable from Nautilus, but we'll loose access to any external storage connected to the Android device.

I'd be tempted to merge this for now (since I don't think it has any drawbacks compared to what we currently have).  When someone finds time, they can look at reworking the code to mount each item in `multiPaths`.

mavit:
Although it doesn't solve all the problems, I still think this is beneficial and worth merging.

ferdnyc:
@mavit 

Here's the problem with this, I've discovered:

You **can't** mount `sftp://${host}:${packet.body.port}/${packet.body.path}/` with GVFS. It won't let you. It will only let you mount `sftp://${host}:${packet.body.port}/` — even if you try to give it a path, it will mount the root of the remote host, and you'll have to browse to `${packet.body.path}` manually to see the shared directory — potentially slamming up against permission-denied errors, attempting to browse from the root `sftp://${host}:${packet.body.port}/` directory to the shared path.

It SUUUUUUUCKS, but it's how GVFS works. It won't let us mount subpaths, and it won't let us mount multiple separate paths on the same remote host.

So I really don't know how to address this in a clean and user-friendly manner. The only thing I can think of is some sort of hybrid mount setup where we mount the remote root, then create a local directory (`$HOME/GSConnect/` maybe?) where we place symlinks to the shared paths on the mounted remote. The user can then browse that directory to navigate to the permitted folders.

ferdnyc:
(This would also be a lot easier if Nautilus still supported `.desktop` files, because we could create a `.desktop` file with a link to `sftp://${host}:${packet.body.port}/${packet.body.path}/`. Because that no longer works, and we're stuck with _real_ symlinks, we'd have to link to `/run/user/${uid}/gvfs/sftp:host=${host},user=kdeconnect/${packet.body.path}/` instead of the SFTP URL.)

mavit:
With this patch, the user can go “System menu → GSConnect → Files” to see a list of shared folders that they can click on to open in Nautilus.  This is better than nothing (but see also #1794).

So that I can access the mount-point directly from Nautilus/the GTK file selector, I have manually created a bookmark for `/run/user/<uid>/gsconnect/by-name/<name>/storage/emulated/0`.

mavit:
> The only thing I can think of is some sort of hybrid mount setup where we mount the remote root, then create a local directory (`$HOME/GSConnect/` maybe?) where we place symlinks to the shared paths on the mounted remote. The user can then browse that directory to navigate to the permitted folders.

We could replace the existing `$XDG_RUNTIME_DIR/gsconnect/by-name/<name>` symlink with a tree of subdirectories containing symlinks to the advertised directories.  But I think that's an idea for a subsequent pull request.

mavit:
> So that I can access the mount-point directly from Nautilus/the GTK file selector, I have manually created a bookmark for `/run/user/<uid>/gsconnect/by-name/<name>/storage/emulated/0`.

We could [create such bookmarks automatically by appending to `~/.config/gtk-3.0/bookmarks`](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/using_the_desktop_environment_in_rhel_8/managing-bookmarks-in-gnome_using-the-desktop-environment-in-rhel-8#bookmarks-in-gnome_managing-bookmarks-in-gnome).  For example,

```
file:///run/user/5000/gsconnect/by-name/Phone/storage/emulated/0 Phone:/storage/emulated/0
```

But, again, this is a task for a different pull request.

ferdnyc:
> With this patch, the user can go “System menu → GSConnect → Files” to see a list of shared folders that they can click on to open in Nautilus. This is better than nothing (but see also #1794).

It's better than nothing, but not much, and not being able to navigate there _in_ Nautilus is a bummer because nobody likes having to use the user menu to access files. (No matter how clunky or un-clunky we make it.)

Actually, what sucks more is having the _mount_ visible in the Nautilus sidebar, but not being able to click on it without a permission-denied error instantly popping up to slap you in the face. It's a highly-visible and excessively tempting means of _attempting_ to navigate to the remote shares... and it works exactly the way we _don't_ want it to.

vjaykrsna:
### Tired of manually mounting your phone with GSConnect?

I created a simple, lightweight script that completely automates the process. It runs in the background and automatically:

-   **Mounts** your phone's storage when it connects.
-   **Creates a bookmark and a symlink** in your home directory using your device's name.
-   **Cleans everything up** when you disconnect.

It's a "set it and forget it" solution that makes accessing your phone's files on GNOME seamless.

---

**Check it out on GitHub:**
### [https://github.com/vjaykrsna/gsconnect-mount-manager](https://github.com/vjaykrsna/gsconnect-mount-manager)

If you find it useful, please consider giving the repo a ⭐ star!

If you run into any problems or have a feature request, don't hesitate to [open an issue](https://github.com/vjaykrsna/gsconnect-mount-manager/issues).

PR #1799, opened 2024-05-04
Remove non-working shortcut editor

### LINKED ISSUE #1796: Getting keyboard Shortcut already exists for every possible key combination. FIx this shit.
### Describe the bug

I am getting Key combination already in use. I just installed ubuntu 24.04 LTS. then installing this extension. got this error. I haven't set a single keyboard shortcut. 

 
image
Screenshot from 2024-05-01 21-56-01

### Steps to reproduce

try setting keyboard shortcut

### Expected behavior

set a keyboard shortcut.

### GSConnect version

57

### Installed from

GNOME Extensions website

### GNOME Shell version

GNOME 46.0 

Ubuntu 24.04

### Paired device(s)

Redmi note 9 pro(india)

### KDE Connect app version

1.30.1

### Plugin(s)

_No response_

### Support log

_No response_

### Screenshots

Screenshot from 2024-05-01 21-56-01
Screenshot from 2024-05-01 21-57-38
Screenshot from 2024-05-01 21-55-03

### Notes

did this with ubuntu 23.something . got the same error.

ferdnyc:
AKA ["fixing this shit"](https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1796).

It would _probably be better_ if the shortcuts panel included a note about the lack of an editor, and maybe a pointer to a wiki page documenting how keybindings can be manually configured using `dconf` / `gsettings`, for the time being.

This PR does not currently contain any of that. It simply rips out the non-working component. Someone would have to write any additions to the ui/wiki.

### Potential enhancements
- [ ] Add page to the wiki documenting manual keybindings configuration
- [ ] Add link to shortcuts UI pointing to wiki information
- [ ] (For the sake of completeness:) A working keybindings editor

Fixes #1796


ferdnyc:
@andyholmes @daniellandau

(I totally forgot about the shortcut editor, in my [rundown of horribly broken components](https://github.com/GSConnect/gnome-shell-extension-gsconnect/discussions/1774#discussioncomment-8977790), but yeah it's been b0rked hard for a good long while now. Too long to leave it in there, as #1796 politely reminded us.)

andyholmes:
Unless I'm mistaken, the only thing wrong with the dialog is that the hack it used to check keybinding availability no longer works.

If not having that is reason to drop the dialog, I'd suggest dropping keybinding support entirely, since manually entering shortcuts into GSettings won't really solve anything.

ferdnyc:
> Unless I'm mistaken, the only thing wrong with the dialog is that the hack it used to check keybinding availability no longer works.

Ohh, I thought it was _only_ the **dialog's** availability detection that was broken, and causing it to refuse to save new ones. Sounds like it's worse than that.

> If not having that is reason to drop the dialog, I'd suggest dropping keybinding support entirely, since manually entering shortcuts into GSettings won't really solve anything.

Well, the reason to drop the dialog is that it doesn't _work_. No matter what you press, it reports that the shortcut is already in use, and refuses to save it.

If manually defining shortcuts in the preferences wouldn't be any more effective, then I agree, we should just drop the whole feature. It's certainly frustrating to users, that we provide a UI which _appears_ to let them do something, but Lucy-holding-the-football-style yanks it away from them at the last minute.

ferdnyc:
(The code can always be resurrected from the repo, if someone decides to fix it. But at this point, leaving it there broken is worse than removing it.)

daniellandau:
It's now working, but just without grabs so conflicts have to be resolved before trying to add a new one here.

PR #1812, opened 2024-05-24
Update Black config, AppStream template

ferdnyc:
### Python: Configure Black for Python 3.8-3.12

Our previous Black `target-version`, `py36`, reached its end-of-life date some time ago. In fact, `py37` is already unsupported, as well.

In addition, according to the Black docs, we should [configure for _all_ supported versions][1], not simply a minimum.

So, set the targeted Pythons to `py38` through `py312` inclusive.

[1]: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#command-line-options

### AppStream: Update SPDX license, developer tag

* The license for the AppStream metadata itself is `CC0-1.0`, says so right in the `metadata_license` tag. Correct the REUSE header to match.
* The `developer_name` tag [is deprecated][1], replaced by the `developer` tag which contains an `id` attribute and a `name` child. Adopt the new hotness.

[1]: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer_name


ferdnyc:
I got the new `developer` tag format slightly wrong, will update this.

andyholmes (approved):
LGTM

PR #1815, opened 2024-05-27
Resize appstream screenshot, or provide alternative

ferdnyc:
According to `appstream-util`, the maximum permitted size for the app screenshot is 1600x900.

Draft because there are two options here, and I'm looking for input on which one to go with.

I attempted to resize the screenshot _without_ scaling it (to avoid losing detail), by pulling the elements in closer to each other. But everything ended up overlapping and I wasn't very happy with the result. That version is in `gsconnect-overlapping.png` (with the `.xcf` source file also included).

Since I wasn't satisfied with that version, I also scaled the existing image to 1600x900, and saved it back in place of the previous file.

So, take a look and see what you think. If you like the `-overlapping` version, I can make that the new screenshot. If not, I can drop that commit and push the rescaled `image-01.png`.


andyholmes (approved):
Either seems okay with me. It's not terribly up to date, but it's really only a visual overview of features anyways.

ferdnyc:
I may be wrong about the max screenshot size, anyway. That info comes from `appstream-util validate-strict`, which I'm beginning to get the impression is out of date compared to `appstreamcli`. And _that_ doesn't complain about the 1920×1080 screenshot even in its most strict mode.

The current spec includes no minimum or maximum size restrictions at all (I get the impression that guidance [has been evolving](https://github.com/ximion/appstream/issues/349)), although all of the examples do still show screenshot tags for 1600px x 900px images.

ferdnyc:
I'm going to close this, as I no longer think the information about 1600x900 being the max screenshot size is correct w/r/t the actual AppStream _spec_. And since the version of the smaller screenshot that I preferred was the scaled-down one, it makes no difference if we leave it to AppStream consumers to handle resizing the image.

(Plus, if they'll be showing it _smaller_ than 1600x900 — which seems likely — then it'll look better scaled down from 1920x1080 than if it's _double_-scaled first to 1600x900, then to whatever size it's displayed at.)

Issue #1025, opened 2020-11-29
GSConnect cannot send multiple files at once


demokritos:
**Describe the bug**

GSConnect cannot send multiple files at once. If I try sending multiple files to my phone, only one file is received by my phone.
GSConnect shows "Transfer Successful" notification only for the first file, and "Transferring File" notifications for the other files remain. 

**Steps To Reproduce:**

1. Go to Nautilus
2. Select two files
3. Open context menu and click a mobile device below "Send to Mobile Device"
4. Find received files on the mobile device

**Expected behavior**

* GSConnect should notify "Transfer Successful" for two files
* The mobile device should also notify that it received two files from desktop, and its download folder should have those two files.

**Support Log**

```
GSConnect: 44 (user)
GJS:       16601
Session:   x11
OS:        Arch Linux
--------------------------------------------------------------------------------
-- Journal begins at Mon 2020-06-29 18:42:08 KST, ends at Sun 2020-11-29 21:46:50 KST. --
11월 29 21:45:55 dbus-daemon[1186]: [system] Activating via systemd: service name='org.freedesktop.hostname1' unit='dbus-org.freedesktop.hostname1.service' requested by ':1.134' (uid=1000 pid=6348 comm="gjs /home/esrevinu/.local/share/gnome-shell/extens")
11월 29 21:45:55 kernel: audit: type=1334 audit(1606653955.260:214): prog-id=35 op=LOAD
11월 29 21:45:55 kernel: audit: type=1334 audit(1606653955.260:215): prog-id=36 op=LOAD
11월 29 21:45:55 audit: BPF prog-id=35 op=LOAD
11월 29 21:45:55 audit: BPF prog-id=36 op=LOAD
11월 29 21:45:55 systemd[1]: Starting Hostname Service...
11월 29 21:45:55 dbus-daemon[1186]: [system] Successfully activated service 'org.freedesktop.hostname1'
11월 29 21:45:55 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=systemd-hostnamed comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
11월 29 21:45:55 systemd[1]: Started Hostname Service.
[...]

PR #1821, opened 2024-05-31
Feature: Add multiple file transfer, directory transfer from Nautilus

ferdnyc:
This PR adds the ability to transfer multiple files at once over the KDEConnect "share files" link, and upgrades the Nautilus/Nemo extension with the ability to both handle multi-select (sending all of the selected items), _and_ to send directory trees via a temporarily-created `.zip` file, which are deleted by the JS code after transferring the files (successfully or not).

Fixes: #1025

### DIFF
--- a/nautilus-extension/nautilus-gsconnect.py
+++ b/nautilus-extension/nautilus-gsconnect.py
@@ -11,9 +11,13 @@ developers for the sister Python script 'kdeconnect-send-nautilus.py':
 https://github.com/Bajoja/indicator-kdeconnect/blob/master/data/extensions/kdeconnect-send-nautilus.py
 """
 
-import gettext
 import os.path
+import pathlib
 import sys
+import tempfile
+import zipfile
+import typing as T
+from gettext import translation, GNUTranslations, NullTranslations
 
 import gi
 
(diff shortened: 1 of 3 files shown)


ferdnyc:
I also added a commit setting the flake8 maximum line length to 80, because there was a conflict where `black` would reformat a statement of exactly 80 characters to one line, then flake8 would complain about it and  want it wrapped, but then black would want to reformat it back to one line...

ferdnyc:
I'll have to figure out what about these changes are breaking `testSharePlugin.js`, plus it would be good to add some new tests for the new functionality.

ferdnyc:
@andyholmes I was hoping you might be able to give me some insight into why the CI is failing.

The failure is in `testSharePlugin.js`, which makes sense as that's the plugin I modified. The specific failure is test #2, "The share plugin enables its GActions when connected". Here's that portion of `gsconnect_plugins_testSharePlugin.test.txt`:

```text
not ok 2 The share plugin enables its GActions when connected
# Message: Expected false to be true.
# Stack:
#   @file:///usr/libexec/installed-tests/gsconnect/plugins/testSharePlugin.js:58:67
#   setTimeout/source<@resource:///org/gnome/gjs/modules/esm/_timers.js:72:9
#   _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34
# Message: Expected false to be true.
# Stack:
#   @file:///usr/libexec/installed-tests/gsconnect/plugins/testSharePlugin.js:58:67
#   setTimeout/source<@resource:///org/gnome/gjs/modules/esm/_timers.js:72:9
#   _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34
# Message: Expected false to be true.
# Stack:
#   @file:///usr/libexec/installed-tests/gsconnect/plugins/testSharePlugin.js:61:68
#   setTimeout/source<@resource:///org/gnome/gjs/modules/esm/_timers.js:72:9
#   _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34
# Message: Expected false to be true.
# Stack:
#   @file:///usr/libexec/installed-tests/gsconnect/plugins/testSharePlugin.js:61:68
#   setTimeout/source<@resource:///org/gnome/gjs/modules/esm/_timers.js:72:9
#   _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34
```

I did add a couple of new actions to the plugin. The only reason I can think of that they might not be activating is, both of the new actions have `kdeconnect.share.request.update` as one of their outgoing message types; that message is also added to the `outgoingCapabilities` of the plugin itself.

`testSharePlugin.js` currently configures its test rig with only the `kdeconnect.share.request` message as both incoming and outgoing:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/97a12873e4a9518ff544ae953ecd1fa24f29787b/installed-tests/suites/plugins/testSharePlugin.js#L12-L32

Do I need to add `kdeconnect.share.request.update` to the test rig, to get the new actions to auto-enable? At which endpoint(s)?

ferdnyc:
As I've said, it would also be good to add _actual tests_ for the new code. But I figure first I need to get the tests working _without_ testing the new code, before I worry about adding more tests.

andyholmes:
This is the section here that controls enabling/disabling actions:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/97a12873e4a9518ff544ae953ecd1fa24f29787b/src/service/plugin.js#L131-L143

So as long as every incoming/outgoing packet type for the action is also in the incoming/outgoing for the device, it will be enabled. So I'd guess what you need is to add `kdeconnect.share.request.update` to both sides of the test rig.

ferdnyc:
@andyholmes That seems to have done it, thanks!

Now I just have to gin up some tests for the _new_ code.

brccabral:
following issue...

Issue #1836, opened 2024-07-01
Presentation Control on Wayland unreliable


tedzards509:
### Describe the bug

On Wayland, GSConnect has to request access to perform input actions. This is accepted automatically. However, I assume that it is accepted after receiving a request, so the first request is disregarded. When using GSConnect as a presentation control, sometimes you have to press twice to get a reaction in the presentation.

I assume this also applies to remote keyboard input, but I cannot test that right now.

### Steps to reproduce

1. Start a presentation
2. Hit next on presentation remote in KDE Connect app

### Expected behavior

Presentation should go to next slide whenever the next button is hit.

### GSConnect version

57

### Installed from

GitHub releases (zip file)

### GNOME Shell version

46

### Linux distribution/release

Fedora 40

ferdnyc:
@tedzards509 

> On Wayland, GSConnect has to request access to perform input actions. This is accepted automatically.

It's not really as complicated as all that. We're not using the desktop portal, which would require user confirmation.

What happens when GSConnect receives a remote input request (which is used by both the presentation and remote input plugins) is:

1. It makes a `CreateSession` request to `org.gnome.Mutter.RemoteDesktop` over dbus. Mutter replies with a dbus session bus address which the caller (GSConnect) and only the caller can use to create input events.
2. GSConnect sends a `Start` request to that new session address, beginning a remote input session. That's when this icon appears in the Shell top bar, indicating the desktop is being remotely controlled:
    ![image](https://github.com/user-attachments/assets/c50423a1-0cf8-491c-97ec-a2c3686a4aa2)
3. GSConnect then sends `Notify{Keyboard,Pointer}*` events to the session address to create input events.

> However, I assume that it is accepted after receiving a request, so the first request is disregarded. When using GSConnect as a presentation control, sometimes you have to press twice to get a reaction in the presentation.

I don't know if the first request is _discarded_, but it does appear that when the remote input connection has to be reset, the first input event that triggers a reconnect is **lost**. And the timeout for dropping the session is _surprisingly_ short, set by GSConnect at 15 seconds.

There's no reason for it to time out that quickly, Mutter is happy to keep the remote input session open indefinitely. It's entirely GSConnect that's cutting the connection off so early.

So, as a first step towards fixing this, I propose to simply raise the timeout **significantly**, to something like 300 seconds. That should make this problem much less common, without keeping the remote input session alive for so long that it possibly interferes with powersave timeouts and the like.

[...]

PR #1844, opened 2024-07-20
Raise remote input timeout to 300 seconds

ferdnyc:
The first input event on reconnection is currently being lost, this should make reconnections much less frequent. See #1836.

I don't want to mark that issue as _fixed_ with this, since it's really just a workaround until we can figure out why input events are being lost. See my comment on the issue for more detail(/confused ramblings).

### DIFF
--- a/src/service/components/input.js
+++ b/src/service/components/input.js
@@ -10,7 +10,7 @@ import GObject from 'gi://GObject';
 import AtspiController from './atspi.js';
 
 
-const SESSION_TIMEOUT = 15;
+const SESSION_TIMEOUT = 300;
 
 
 const RemoteSession = GObject.registerClass({


andyholmes:
If the `Closed` signal works properly, you can probably remove the timeout altogether. I don't recall why I did that, it was quite some time ago now.

andyholmes (approved):

ferdnyc:
@andyholmes I think we'd want to have _some_ kind of timeout, because AFAICT Mutter doesn't have one, and is happy to leave us connected indefinitely. And there doesn't seem to be any message sent from KDE Connect that would signal the end of an input session (when the user exits the "Presentation remote" or "Remote input" interfaces, for example).

So without a timeout it seems like GSConnect would just sit around being connected... forever. Which I'm concerned could interfere with powersave and the like. (Though it's probably the other way around, entering powersave would probably forcibly disconnect us.)

PR #1860, opened 2024-09-11
Don't re-show repeated notifications

### LINKED ISSUE #1855: GSConnect notifications endlessly re-notify on GNOME 46/Ubuntu 24.04
### Describe the bug

I recently upgraded from Ubuntu 23.10 to 24.04, and this moved me to GNOME 46 (I think I was on 45, not sure). And as a result, notifications from GSConnect on the desktop just keep renotifying non-stop until I dismiss them (which I don't want to do, as that clears the notification on the phone too, which I don't want until I handle the notification).

Previously (before the upgrade and GNOME version change) I would get notified once per GSConnect notification thing (like, per E-Mail) and it would be hidden in the notification tray thinggy in Gnome. And that is exactly what I wanted.

Now, however, every, single, GSConnect notification, re-notifies after some period of time I haven't measured in the realm of 1-2 minutes or so. And when I get a lot of notifications from my phone that I want to handle _LATER_ this turns into a lot of spam on my desktop I don't want.

### Steps to reproduce

I believe I've already described how to reproduce this.

### Expected behavior

I believe I've already answered this.

### GSConnect version

57

### Installed from

GNOME Extensions website

### GNOME Shell version

46

Ubuntu Desktop 24.04

### Paired device(s)

Samsung Galaxy XCover6 Pro

### KDE Connect app version

1.32.0

### Plugin(s)

Notifications

### Support log

_No response_

### Screenshots

_No response_

### Notes

I really think this is somehow tied to GNOME 46 as this behaviour change did not happen to me in the past.

ferdnyc:
GSConnect receives a lot of duplicate notifications, whether due to reconnects, proactively requesting updates, etc.

Since GNOME 46, each of those repeats has been popped up as a new banner notification every time it's received, even if nothing has changed in the notification content. This is extremely disruptive.

To avoid that, don't blindly flag all repeated notifications as "unacknowledged", which will cause them to be popped up again. Only do so if the content has changed. (This covers re-displaying banners for SMS conversations that receive an additional message, since those are sent as updates to the existing notification.)

Fixes #1855

### DIFF
--- a/src/shell/notification.js
+++ b/src/shell/notification.js
@@ -239,19 +239,20 @@ const Source = GObject.registerClass({
         if (cachedNotification) {
             cachedNotification.requestReplyId = requestReplyId;
 
-            // Bail early If @notificationParams represents an exact repeat
             const title = notification.title;
             const body = notification.body
                 ? notification.body
                 : null;
 
+            // Bail early If @notification represents an exact repeat
             if (cachedNotification.title === title &&
                 cachedNotification.body === body)
                 return cachedNotification;
 
+            // If the details have changed, flag as an update
             cachedNotification.title = title;
             cachedNotification.body = body;
-
+            cachedNotification.acknowledged = false;
             return cachedNotification;
         }
 
@@ -304,10 +305,8 @@ const Source = GObject.registerClass({
      * notification limit (3)
      */
     _addNotificationToMessageTray(notification) {
-        if (this.notifications.includes(notification)) {
-            notification.acknowledged = false;
+        if (this.notifications.includes(notification))
             return;
-        }
 
         while (this.notifications.length >= 10) {
             const [oldest] = this.notifications;


andyholmes:
This seems reasonable, but I'm not familiar with the implementation anymore. If it works in practice, I'd say it's fine :+1:

ferdnyc:
> If it works in practice

Yeah, that's the $64,000 question. I'm hoping it'll score at least one or two testers from #1855 to confirm.

Most I can say is, "it works _for me_"... but I don't feel like that represents nearly a wide enough range of scenarios. The apps I have notifications enabled for number fewer than a dozen — _just_ enough for their constant re-displaying to annoy me into fixing it.

My limited local testing _did_ include a Google Messages notification, which no longer endlessly repeated, _but_ still popped up anew when a new SMS came in that was added to that conversation. But "saw it work exactly once" feels more like a reason not to halt further testing, than to declare unqualified success.

ferdnyc:
Reports from @michas79de after a few hours of testing match my own results:

> the same notifications that popped up repeatedly before are not popping up again.
> An SMS notification only appeared once on arrival and a second time hours later when a second SMS arrived from the same sender and the notification was updated/replaced.
> So it seems to be working fine.

It sounds like this is good to go, it at least doesn't make things appreciably **worse** and appears to solve the notification-repetition issue.

ferdnyc:
Just need an approval to be able to merge this one; reviewers?

andyholmes:
I'll approve this, given there's two confirmations.

andyholmes (approved):

BloodyIron:
How soon before we can get a new release with this baked in? :) Sorry I haven't been able to chip in on testing this. :(

ferdnyc:
@BloodyIron 

That's a question primarily for @daniellandau. Whose last activity in the GSConnect repo, I'm now seeing, was back in _April_ (!). So, that's...not ideal.

BloodyIron:
Only one person can push a release? Oof

daniellandau:
e.g.o only allows a single access, but others can do a release here too. I'll try to get around to it soon too if no one else does before me.

ferdnyc:
@daniellandau Welcome!

> others can do a release here too.

True enough, honestly the only reason I haven't put together a release PR yet is that #1863 is still unmerged, and there's absolutely no point in doing a release until that goes in.

(I question whether there's any point to doing a release at all without it getting submitted to e.g.o. But I'm sure having the release already tagged and packaged makes the submission process easier for you, so happy to help out there.)

I'll check the [v58 milestone](https://github.com/GSConnect/gnome-shell-extension-gsconnect/milestone/11) I slapped together last week (we do milestones now, **#SOORGANIZED!**) and if there's nothing other than #1863 outstanding, I'll get started on a PR for the v58 release.

ferdnyc:
Yeah, I meant #1866 there, every time.

mavit:
> I question whether there's any point to doing a release at all without it getting submitted to e.g.o.

Here's one reason.  Fedora provides this extension as an [RPM package](https://packages.fedoraproject.org/pkgs/gnome-shell-extension-gsconnect/gnome-shell-extension-gsconnect/).  The maintainer gets [notified](https://bugzilla.redhat.com/show_bug.cgi?id=2321869) to update that package [when a release happens here on GitHub](https://release-monitoring.org/project/18402/), not when it hits extensions.gnome.org.

(On this occasion, that happened too late for the update to be included in the initial release of Fedora 41 (which includes Gnome 47), so it's currently not working for early adopters)

daniellandau:
> But I'm sure having the release already tagged and packaged makes the submission process easier for you

This is a major reason too. Preparing a release is the bigger job, uploading a zip to e.g.o is just a tiny final part.

ferdnyc:
@mavit

> Here's one reason. Fedora provides this extension as an [RPM package](https://packages.fedoraproject.org/pkgs/gnome-shell-extension-gsconnect/gnome-shell-extension-gsconnect/). The maintainer gets [notified](https://bugzilla.redhat.com/show_bug.cgi?id=2321869) to update that package [when a release happens here on GitHub](https://release-monitoring.org/project/18402/), not when it hits extensions.gnome.org.
> 
> (On this occasion, that happened too late for the update to be included in the initial release of Fedora 41 (which includes Gnome 47), so it's currently not working for early adopters)

Valid, though on our end that packaging is Not Recommended™ anyway. (Especially since the maintainer [refuses to include the Nautilus/Nemo extensions by default](https://src.fedoraproject.org/rpms/gnome-shell-extension-gsconnect/pull-request/2), even as a weak dependency. A position that left me so speechless, I've failed to follow up on it for over 2 years already.)

But it also adds a 2-week delay to users receiving extension updates, vs. e.g.o — a delay that in normal Fedora packaging is justified as time for testing of the package before inflicting it on users, but I question whether that testing is more theoretical than practical, in reality.

Also, GSConnect does a lot of behind-the-scenes trickery with a backend daemon process and whatnot, and there's some evidence that updating in the user's home directory tends to be less fragile than updating a `/usr/` install via DNF, in terms of running installs. (GSConnect's `daemon.js` running from the user space even watches for modifications to its code and automatically relaunches itself. Mostly meant as a developer feature, sure, but it helps with upgrades as well.)

PR #1874, opened 2024-10-13
Set tablet device type for 30, 31(detachable) and laptop for 31(convertible)

sidevesh:


andyholmes (approved):
This LGTM. They should probably be the hex codes, but that's my fault :)

PR #1875, opened 2024-10-13
Set appropriate window titles instead of defaulting to gsconnect or GSConnect

sidevesh:


src/service/ui/legacyMessaging.js
```diff
@@ -40,6 +40,7 @@ const Dialog = GObject.registerClass({
     _init(params) {
         super._init({
             application: Gio.Application.get_default(),
+            title: _('Send SMS'),
```
ferdnyc, line 43: This is already set in the template, setting it here is redundant and makes maintenance harder:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/3758807fac3f2b3c09a30a1767cf7f551d27d915/data/ui/legacy-messaging-dialog.ui#L15

src/service/ui/messaging.js
```diff
@@ -833,7 +833,10 @@ export const Window = GObject.registerClass({
 }, class MessagingWindow extends Gtk.ApplicationWindow {
 
     _init(params) {
-        super._init(params);
+        super._init(Object.assign({
+            application: Gio.Application.get_default(),
+            title: _('Messaging'),
```
ferdnyc, line 838: This is already set in the template, setting it here is redundant and makes maintenance harder:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/3758807fac3f2b3c09a30a1767cf7f551d27d915/data/ui/messaging-window.ui#L21

src/service/ui/mousepad.js
```diff
@@ -100,6 +100,8 @@ export const InputDialog = GObject.registerClass({
 
     _init(params) {
         super._init(Object.assign({
+            application: Gio.Application.get_default(),
+            title: _('Remote Input'),
```
ferdnyc, line 104: This is set by defining the `headerbar.title`, which is the title of a GtkDialog in every way that matters.

In the application dash and overview:

image image

In the Gtk Inspector:

image

image

What issue are you trying to solve, by also passing a `title` parameter to the constructor?

src/service/ui/notification.js
```diff
@@ -45,6 +45,7 @@ const ReplyDialog = GObject.registerClass({
     _init(params) {
         super._init({
             application: Gio.Application.get_default(),
+            title: _('Reply'),
```
ferdnyc, line 48: In this case, `headerbar.title` is set to `params.notification.appName`, which will be the title of the dialog in every way that matters. What's the purpose of also setting it generically, here?

src/service/ui/service.js
```diff
@@ -53,6 +53,7 @@ export const DeviceChooser = GObject.registerClass({
         super._init({
             use_header_bar: true,
             application: Gio.Application.get_default(),
+            title: params.title,
```
ferdnyc, line 56: `title` is already set in the template, as `Select a Device`, which is why `params.title` is set as the headerbar _subtitle_:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/3758807fac3f2b3c09a30a1767cf7f551d27d915/data/ui/service-device-chooser.ui#L15

src/service/plugins/findmyphone.js
```diff
@@ -144,6 +144,8 @@ const Dialog = GObject.registerClass({
 }, class Dialog extends Gtk.MessageDialog {
     _init(params) {
         super._init({
+            application: Gio.Application.get_default(),
+            title: _('Ringing'),
```
ferdnyc, line 148: This dialog is a _mess_, and doesn't appear to have a title **at all**. So I'll probably give you this one. But really it should probably be redone with a UI template definition, like basically every other window in the code.

(Actually now that notifications can have persistent action buttons (FINALLY!), it might make **way** more sense to replace the entire dialog with an urgent notification that has a "Silence" or "Acknowledge" action button embedded.)

ferdnyc (commented):
I'm not at all sold on most of these. They appear to be solutions in search of a problem, and many are redundant with the UI template definitions. (Redundancy is bad.) -1 from me without better justification for why it's necessary.

ferdnyc:
The tests are failing in the CI because you added `Gio.Application.get_default()` calls to modules where `Gio` isn't imported, BTW.

sidevesh:
Sorry for the messy changes, I was trying to solve the issue of generic gsconnect icon and name in the dash,
I agree the addition of title in almost all the places is not useful, I will remove those,
are you okay with passing the default application to constructor so that it shows those windows as gsconnect app instead of the generic icon in the dash ?

PR #1876, opened 2024-10-13
Don't show Ringing window fullscreen and center it rather

sidevesh:


ferdnyc:
There is _so much_ wrong with that window, in every conceivable way.

As I mentioned in https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1875#discussion_r1804210555 it might even be worth considering that we just dump the window entirely in favor of a notification. Now that action buttons on notifications persist into the drawer and can still be clicked even if you miss the popup, that's not so unworkable as it used to be.

PR #1890, opened 2024-11-14
refactor(systemvolume): remove `requestSinks`

andyholmes:
Sinks are sent at connect time and updated as necessary, so drop the request command.

See:
- https://invent.kde.org/network/kdeconnect-android/-/merge_requests/500
- https://invent.kde.org/network/kdeconnect-kde/-/merge_requests/756


ferdnyc (approved):
Sorry, I'd been meaning to look at this for a while, and kept getting distracted. LGTM, anything that cuts down on the traffic noise every time a device connects is a win.

Issue #1893, opened 2024-11-27
Prevent Sending Passwords from the Clipboard


cesar19004:
### Describe your request

Auto-sharing the clipboard to a mobile device is really useful; however, when using password managers, you might not want to share passwords, as they could be stored insecurely in the mobile device's clipboard manager.

### Proposed solution

Use clipboard MIME data to determine if content should be shared. Most password managers set the appropriate MIME data to mark content as private, as demonstrated in this [example](https://github.com/keepassxreboot/keepassxc/blob/bf856d278d21be2e54d5590fb7e69e63184262d3/src/gui/Clipboard.cpp#L57-L63) from KeePassXC.

KDE Connect already [uses](https://invent.kde.org/network/kdeconnect-kde/-/blob/master/plugins/clipboard/clipboardlistener.cpp#L64) the `x-kde-passwordManagerHint` MIME data to [check for passwords](https://invent.kde.org/network/kdeconnect-kde/-/blob/master/plugins/clipboard/clipboardplugin.cpp#L40-44). It also includes a [config option](https://invent.kde.org/network/kdeconnect-kde/-/blob/master/plugins/clipboard/clipboardplugin.cpp#L49-L60) to control whether sharing passwords is permitted.

### GSConnect version

58

### Installed from

GNOME Extensions website

### GNOME Shell version

47

### Linux distribution/release

Ubuntu 24.10

cesar19004:
Thanks for taking the time.
If I'm correct, these are the MIME types accepted by GSConnect:
```js
const TEXT_MIMETYPES = [
    'text/plain;charset=utf-8',
    'UTF8_STRING',
    'text/plain',
    'STRING',
];
```

And these are the MIME types of passwords returned by the GNOME evaluator when running `St.Clipboard.get_default().get_mimetypes(Meta.SelectionType.SELECTION_CLIPBOARD)`:
```
0: text/plain;charset=utf-8
1: x-kde-passwordManagerHint
2: text/plain
```

For now, maybe a simple validation to check that `x-kde-passwordManagerHint` is not present in the MIME types should be enough, while an option in the config to control password sharing is being considered.

ferdnyc:
@cesar19004 

Mmm, it looks like currently we special-case ignore any clipboard with `text/uri-list` in its mimetypes... 

[...]

PR #1894, opened 2024-11-28
Clipboard: Block proxying of passwords

ferdnyc:
Since some password managers set a clipboard mimetype to signal when the clipboard contains password data, ignore the clipboard contents in those cases, to avoid exposing passwords over the link to the mobile device.

Fixes: #1893

### DIFF
--- a/src/service/components/clipboard.js
+++ b/src/service/components/clipboard.js
@@ -192,6 +192,14 @@ const Clipboard = GObject.registerClass({
         if (mimetypes.includes('text/uri-list'))
             return;
 
+        // Special case to avoid copying identifiable passwords,
+        // for privacy reasons.
+        // (See https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1893)
+        // TODO: Should there be a preference to re-enable password-sharing,
+        // for users who desire it?
+        if (mimetypes.includes('x-kde-passwordManagerHint'))
+            return;
+
         const text = await new Promise((resolve, reject) => {
             this._clipboard.request_text((clipboard, text) => resolve(text));
         });


daniellandau:
As I wrote in the linked issue, I feel copying passwords is a valid use case and IMO it shouldn't be blocked by default.

ferdnyc:
Heh. I happened to install KDE Connect on my laptop's Plasma session today (looking at some craziness in Messages, where the Android app seems to be sending nonsense to GSConnect — I want to see how they handle it), and sure enough, in the clipboard-sharing preferences:

Screenshot_20250112_054935

ferdnyc:
So, this becomes tricky because of the separation between our _backend_ clipboard watcher (the component responsible for watching and updating the local desktop clipboard), and our clipboard _plugin_ that runs for each connected device, interfacing between the backend and its paired device's network connection.

The way GSConnect's preferences work, with only one or two exceptions settings apply to _plugins_, and for the most part are _per-device_. Which means the decision about whether to send any given clipboard string to a paired device is made by the plugin based on its current settings. But in our current architecture, the only thing the plugin ever sees from the backend is a piece of string data — it has no way of knowing if that string represents a password. The mimetype data we're looking at to detect passwords on the clipboard is only accessible from the backend.

So, I think I'm going to have to go a route similar to what KDE did in their own Linux implementation: Have the backend assign a "content type" to each clipboard string it captures, indicating whether it's a password, a (presumed) non-password string, or... well, at least for now, those are the only two choices. `content-type` would be an additional backend property, to go with the existing `text` property that holds the clipboard string.

Whenever the plugin gets notified that the backend's `text` property has been updated, it can first look at the backend `content-type` property. If it sees "password", the plugin decides (based on the its security setting) whether or not it should grab `text` and send it to the paired device.

PR #1899, opened 2024-12-24
Changed 'connect to' to 'Add Device by IP'

realAbhi-web:
Changed the "connect to" option to "Add devices by IP"
image

Fixes #1886

PR #1902, opened 2025-01-03
Update translation files for new strings

ferdnyc:
#1899 added a new translatable string.

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: gsconnect\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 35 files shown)


ferdnyc:
Can get a +1 on this, maintainers? There's a new string that we need to push to Crowdin for translation.

daniellandau (approved):
👍

ferdnyc:
I've updated the screenshot used for that part of the interface, too, so translators have a visual reference of the new string.

(Which took me _half an hour_, and an eventual trip to their documentation — I'd love to meet the yahoo at Crowdin who decided that **double-clicking** on the screenshot was a sane choice to make the ONLY means of reaching the tagging/updating interface.)

PR #1905, opened 2025-01-09
feat: add support for protocol version 8

andyholmes:
Add support for the planned changes in protocol version 8.

See: https://invent.kde.org/network/kdeconnect-android/-/commit/454b2fd5d7b3162443fbec4a6f0b7cafa1b554ee

### DIFF
--- a/installed-tests/fixtures/backend.js
+++ b/installed-tests/fixtures/backend.js
@@ -292,7 +292,7 @@ export const ChannelService = GObject.registerClass({
                 deviceId: this.id,
                 deviceName: this.name,
                 deviceType: 'desktop',
-                protocolVersion: 7,
+                protocolVersion: 8,
                 incomingCapabilities: [],
                 outgoingCapabilities: [],
                 tcpPort: this.port,
--- a/installed-tests/fixtures/utils.js
+++ b/installed-tests/fixtures/utils.js
@@ -110,7 +110,7 @@ export function generateIdentity(params = {}) {
             'deviceId': GLib.uuid_string_random(),
             'deviceName': 'Test Device',
             'deviceType': getDeviceType(),
-            'protocolVersion': 7,
+            'protocolVersion': 8,
             'incomingCapabilities': [],
             'outgoingCapabilities': [],
         },
--- a/src/service/backends/lan.js
+++ b/src/service/backends/lan.js
@@ -730,6 +730,13 @@ export const Channel = GObject.registerClass({
                 throw new Error('missing deviceId');
 
             this._connection = await this._encryptClient(connection);
+
+            // Starting with protocol version 8, the devices are expected to
+            // exchange identity packets again after TLS negotiation
+            if (this.identity.body.protocolVersion >= 8) {
+                await this.sendPacket(this.backend.identity);
+                this.identity = await this.readPacket();
+            }
         } catch (e) {
             this.close();
             throw e;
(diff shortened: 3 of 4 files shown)


ferdnyc:
@andyholmes Yesterday Albert also [updated the protocol docs](https://invent.kde.org/network/kdeconnect-meta/-/commit/e4f93eecb876dccf1f35e6db4944d536a8b9bc53) to make it required that devices have UUID-format device IDs, the definition is now:

> * `deviceId`: **`String`** 🔒
> 
>     **`pattern`**: `/^[a-zA-Z0-9_]{32,38}$/`
> 
>     A unique ID for the device. Must be between 32 and 38 characters and made of only alphanumerical characters and underscores. A valid device ID can be generated from a random UUIDv4 string by removing the hyphens (`-`) or replacing them with underscores (`_`), such as `740bd4b9_b418_4ee4_97d6_caf1da8151be`.
> 

I feel like I still have some devices on my pairing list that use a shorter ID string. And even if not, GSConnect IDs are currently _real_ UUIDs (with hyphens instead of underscores). 

I wonder if we need code to migrate devices to new-format IDs?

andyholmes:
> @andyholmes Yesterday Albert also [updated the protocol docs](https://invent.kde.org/network/kdeconnect-meta/-/commit/e4f93eecb876dccf1f35e6db4944d536a8b9bc53) to make it required that devices have UUID-format device IDs, the definition is now:
> 
> > * `deviceId`: **`String`** 🔒
> >   **`pattern`**: `/^[a-zA-Z0-9_]{32,38}$/`
> >   A unique ID for the device. Must be between 32 and 38 characters and made of only alphanumerical characters and underscores. A valid device ID can be generated from a random UUIDv4 string by removing the hyphens (`-`) or replacing them with underscores (`_`), such as `740bd4b9_b418_4ee4_97d6_caf1da8151be`.
> 
> I feel like I still have some devices on my pairing list that use a shorter ID string. And even if not, GSConnect IDs are currently _real_ UUIDs (with hyphens instead of underscores).

I think the shorter codes are old Android hostnames, but yeah, the certificate & device ID generation needs to be updated.

> I wonder if we need code to migrate devices to new-format IDs?

I don't think true migration will be possible, since the device ID is directly tied to the certificate. It's just unavoidable that folks will have to repair some devices

I suppose detecting a certificate with an old ID is possible and then showing a popup informing the user.

ferdnyc:
@andyholmes 

> > I wonder if we need code to migrate devices to new-format IDs?
> 
> I don't think true migration will be possible, since the device ID is directly tied to the certificate. It's just unavoidable that folks will have to repair some devices

Oh, yeah, but mostly I meant that there are a lot of (GSConnect) installations out there with `real-uuid-format-deviceIDs`, that need to get their IDs regenerated as `fake_uuid_format_deviceIDs`. (And then, as you say, their certificates regenerated.) Normally, we wouldn't generate a new local device ID unless it was a first-time install.

And _then_ they'll _**also**_ have to re-pair all of their devices.

daniellandau:
Are these problems specific to gsconnect? Has the KDE side been using fake_uuids_with_underscores already?

ferdnyc:
@daniellandau 

> Are these problems specific to gsconnect? Has the KDE side been using fake_uuids_with_underscores already?

Apparently yes, Albert's comment on making the change was:

> All implementations have been using UUIDs as device IDs for a while, so we can start validating and enforcing this format.

However, while it's true GSConnect has been using _UUIDs_ as device IDs, we've been using them with hyphens instead of underscores, while KDE Connect has apparently been doing the reverse.

ferdnyc:
I opened #1910 to track the device ID format issue.

andyholmes:
I'm going to close this one up, I think it makes sense to do all of protocol v8 in #1911, just with nice small commits.

ferdnyc:
@andyholmes Anything on the branch you need to hang on to, or can it be deleted?

andyholmes:
Nope :)

ferdnyc:
Hey, wait. Nope you don't need it, or nope it can't be deleted? :smiling_imp:

andyholmes:
Nope I don't need it, I hit the delete button :)

PR #1911, opened 2025-01-18
feat: add support for protocol version 8

### LINKED ISSUE #1910: GSConnect device ID format needs to be updated
Yesterday Albert also [updated the protocol docs](https://invent.kde.org/network/kdeconnect-meta/-/commit/e4f93eecb876dccf1f35e6db4944d536a8b9bc53) to make it required that devices have UUID-format device IDs, the definition is now:

> * `deviceId`: **`String`** 🔒
> 
>     **`pattern`**: `/^[a-zA-Z0-9_]{32,38}$/`
> 
>     A unique ID for the device. Must be between 32 and 38 characters and made of only alphanumerical characters and underscores. A valid device ID can be generated from a random UUIDv4 string by removing the hyphens (`-`) or replacing them with underscores (`_`), such as `740bd4b9_b418_4ee4_97d6_caf1da8151be`.
> 

GSConnect IDs are currently _real_ UUIDs (with hyphens instead of underscores). 

Looks like we need code to generate new-format IDs, then trigger that code for all devices _not_ using the correct format, regenerate their certificates, and tell the user they'll need to re-pair all of their devices.

_Originally posted by @ferdnyc in https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1905#issuecomment-2592510772_

andyholmes:
Update the certificate generation to produce EC-based certificates
with the required device ID format and reject connections from those
with an ID not matching `/[a-zA-Z0-9_]{32,38}/`.

Migrate installations to new certificates, with device IDs
following the protocol v8 constraints.

closes #1910


src/service/daemon.js
```diff
@@ -41,10 +42,49 @@ const Service = GObject.registerClass({
...
+        // Remove the old certificate, serving as the single source of truth
+        // for the device ID
+        try {
+            Gio.File.new_build_filenamev([Config.CONFIGDIR, 'certificate.pem'])
+                .delete(null);
+            Gio.File.new_build_filenamev([Config.CONFIGDIR, 'private.pem'])
+                .delete(null);
+        } catch {
+            // Silence errors
+        }
```
ferdnyc, line 65: This all looks sensible, but I wonder if we need some sort of notification to the user, when the migration process triggers? ("Hey, we had to re-generate your local device ID, so all of your devices need to be re-paired. Lusciously sorry!")

andyholmes:
I hate this test suite so much, so much.

ferdnyc:
I don't mind admitting, it scares the bejesus out of me. I managed to wrap my brain around _JUnit_<sup>1</sup> in a day or two, a while back, at least well enough so that I could accomplish what I needed to. But jasmine? Uh-uh. (And the friction increased by a factor of about 5, once the tests could no longer be run _un_-installed after the move to ESM.)

1. (in much the manner that one wraps a car around a telephone pole)

andyholmes:
image

andyholmes:
There's still some bits left to do for protocol v8, so I'm probably going to split this up before review.

src/service/daemon.js
andyholmes: Resolved in [641f5f9](https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1911/commits/641f5f999db42e9d250c2acda12aa7f70700a3ed)

andyholmes:
How do we feel about dropping the JSDoc checks?

src/service/device.js
```diff
@@ -162,48 +173,36 @@ const Device = GObject.registerClass({
...
+        if (a.compare(b) < 0)
+            [a, b] = [b, a]; // swap
+        checksum.update(a.toArray());
+        checksum.update(b.toArray());
+
+        if (this.identity?.body.protocolVersion >= 8) {
+            const buffer = new ArrayBuffer(8);
+            const view = new DataView(buffer)
+                .setBigUint64(0, BigInt(this._pairingTimestamp));
+            checksum.update(new Uint8Array(buffer));
```
andyholmes, line 199: I need to confirm this works correctly, since GSConnect tests against itself. Specifically, `setBigUint64()` and `BigInt()` affect the endianess of how the double is cast into the buffer.

There's a debug build for kdeconnect-android in the corresponding MR, I'm just having trouble getting it to install on my phone.

andyholmes:
I think this is ready for review, or at least some testing.

You can grab a debug build for kdeconnect-android here: https://invent.kde.org/network/kdeconnect-android/-/merge_requests/526. I had to install it with `adb install`, but my Android phone is a bit old.

installed-tests/data/local-private.pem
```diff
@@ -1,52 +1,5 @@
...
-t/Pa1776/xvHBKwC1nhRP6YiB+HTbyoYrjecB4IsXYz6eyTYPzEhRF+encWenkjL
-qqkCggEBAMhz/VK+dTBoawGF5gHd7ppBmpuyvxTaFjy9ZkTq6FC5G4yZtSsOhtOQ
-ve1EF+7hZMkdUIV7B6EzTs6bXPgLaLHbpJym1gTdTGUFNWQvdcF38oSfqEmfBnze
-CbdbMtCY8+Izx1l6gfduz0J4yBQ0ME2uo8FMOjcWFQuqRRkDIE+UF+LFJuINRqT+
-N7RMZH+UYT/a6H+jolqaD+N/qUZzz3fU+GGTwJUpoI+dIUL5kUcGjIg972/XoPRX
-4gFWdEG7qKGRPd+IARhXNxfkJwHv8S89RfEOZL2f6LI3mOqHlwda14WTmH6ct9S6
-258ZGWdYdv8bBRaaLpN7nTcu8+o411E=
+MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgyI4fgKyLmxTgZpQG
+3ReQ8ipNctiJP2+NW+7PQEJGWTahRANCAASZ+sa9a474HsdgxUunA/C+Pfu0wjno
+z6YYPe8mq0JzOVtRz3dBYifHfbUN2JO+RJ6ZkxUiJo1wtRMg9WEgFxdc
```
ferdnyc, line 4: Wow, these got a whole hell of a lot shorter!

ferdnyc (approved):
LGTM! Looking forward to trying it out. I need to stash away the remaining local modifications I'm running with first, then I can install the build artifact.

I had one minor nitpick with a translator comment, but that's all.

src/service/device.js
```diff
@@ -814,15 +812,19 @@ const Device = GObject.registerClass({
             // The device thinks we're unpaired
             } else if (this.paired) {
                 this._setPaired(true);
+                this._pairingTimestamp = Math.floor(Date.now() / 1000);
                 this.sendPacket({
                     type: 'kdeconnect.pair',
-                    body: {pair: true},
+                    body: {
+                        pair: true,
+                        timestamp: this._pairingTimestamp,
```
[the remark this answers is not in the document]
andyholmes: I think they're waiting for us: https://invent.kde.org/network/kdeconnect-meta/-/merge_requests/10/diffs

andyholmes:
Here we gooo!

ferdnyc:
I guess this means we should start thinking about a new release — or several, since it would probably be good to backport this to _at least_ GNOME 46 and maybe 45. (Going past that means backporting to before the ESM conversion, which... if someone wants to do that work and submit a PR, then great, but um... **#NotIt**.) I'll give the 46/45 backports a go, though, some time this coming week.

I have a couple of other things I kind of want to try to get in, as well — a mitigation for the MPRIS spam I documented in #1919, and if I can find the time I'd really like to try and finish up #1921, but we'll see.

daniellandau:
Debian stable is on 43. Will protocol 7 stop working with the Android app at some point, and does the current version of the Android app support both?

andyholmes:
Most clients will probably retain support for protocol v7, but the device IDs will probably be what really breaks them. I expect there's a good chance this will affect some older Android devices as well, unfortunately.

ferdnyc:
@daniellandau @andyholmes 

Well, as it turned out backporting to (OUR) GNOME 45 involved crossing the ESM boundary anyway — even though 45 was the first to support ESM, we didn't make the switch until GNOME 46 — and after doing that, backporting further to GNOME 43 was a relatively minor step. So I  did end up doing both, as #1928 and #1929.

albertvaka:
Do you plan doing a new release anytime soon? It would be awesome to get a new version out with these changes :)

ferdnyc:
@albertvaka Yes, imminently. The last sticking point was the protocol 8 interactions with older devices, but now that that's fixed by 1.33.2, @daniellandau has PRs open to port protocol 8 to our back port branches (a rare exception to our "only the latest GNOME is supported" policy). 

(Due to how the GNOME extensions versioning system works, it's better to release the back ports in ascending order by GNOME version, before releasing the new mainline update.

But from my perspective, as soon as those are merged and tested we're good to go on starting release builds. Unless @daniellandau or @andyholmes has any other concerns?

daniellandau:
That's my thinking here too.

Issue #1879, opened 2024-10-24
Refine, document release process


ferdnyc:
### Describe your request

I've just merged the PRs adding GNOME 47 compatibility and v58 release details to the repo.

But now there's the process of _actually_ releasing v58.

We have a bare-bones [RELEASE_CHECKLIST.md](https://github.com/ferdnyc/gnome-shell-extension-gsconnect/blob/885697b0e18347a3ee01c3cc35e38e2465213a3c/RELEASE_CHECKLIST.md) file to follow, though it's somewhat out of date. (`meson test` no longer works since the GNOME 46 / ESM changes.)

But beyond that, @daniellandau seems to have a defined format for each of the [Releases](https://github.com/GSConnect/gnome-shell-extension-gsconnect/releases). If that's generated by a script or tool (which appears to be the case?), then it isn't part of the repo. And we don't seem to have documented it anywhere.

### Proposed solution

We should probably make sure that all tools needed to maintain the repo are **part of** the repo, and also that processes are documented so that, as said in https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1860#issuecomment-2400555244, "others can do a release here too".

In the meantime, as I don't feel like reinventing any wheels, I'll probably just tag the v58 commit and use a relatively-barebones release "announcement" for the moment.

### GSConnect version

58

### Installed from

GitHub releases (zip file)

### GNOME Shell version

47

ferdnyc:
So, turns out the tool in question is mostly just the "Generate Release Notes" button GitHub provides on the release-authoring page. Good to know. (And better to document.)

andyholmes:
The EGO password will probably need to be with one person, since we don't have any "group"/"organization" infrastructure there (i.e. single password, single email for reset).

That being said, some folks have setup GitHub actions that can push a release to EGO. It probably would be a bit of work, but a release-on-tag setup is probably possible (as long as it doesn't upload per-commit).

ferdnyc:
@andyholmes Honestly I wasn't even thinking about e.g.o — I've accepted that the submission process there is restricted to the designated "owner" of the app. Plus, I created issue #1880 about the need to submit the new release and @daniellandau had it submitted, published, and live on the extension page just under 2 hours later. IOW, so far that part of the process hasn't had any delays or speedbumps.

[...]

Issue #1897, opened 2024-12-23
Seemingly non-functional on PostMarketOS. 


Inderix:
### Describe the bug

GSConnect Settings does not open, nor does the extension appear to be doing much of anything

### Steps to reproduce

1. Install from Extension Manager/GNOME Extensions on PostmarketOS
2. Attempt to open extension settings (From Extension Manager or GNOME Extensions app)
3. Observe bug.
Quick Settings > GSConnect > Mobile Settings also does not work. 

### Expected behavior

GSConnect settings opens, error notifications about missing dependencies as needed, the usual first-run experience. 

### GSConnect version

58

### Installed from

Extension Manager

### GNOME Shell version

47

### Linux distribution/release

PostMarketOS Edge aarch64

### Paired device(s)

N/A

### KDE Connect app version

N/A

### Plugin(s)

N/A

### Support log

```plain text
N/A, as I cannot reach the button for generating the log.
```

### Notes

`gnome-extensions info gsconnect@andyholmes.github.io` returns that the extension is enabled and it's current state is "active"
when running `gnome-extensions prefs gsconnect@andyholmes.github.io`, the command returns no output and then exits to prompt.

I originally was trying to use the extension on PostmarketOS's latest stable branch, which only has Gnome 46 for the time being still. I upgraded to Edge (which provides G47) in hopes of it working, but as you can see by this report, it does not.

daniellandau:
Interesting. What form factor device is this? As in normal gnome shell or mobile?

Inderix:
I'm running the normal PC/desktop/tablet Gnome Shell, I haven't tested with Phosh.

daniellandau:
Phosh definitely wouldn't work, I was thinking about the mobile version if gnome shell. I think I could get postmarketos going on some device during the next few weeks to check.

Inderix:
Ah, I'm just using bone stock gnome shell yeah.
I'll keep an eye on this thread, if there's anything you want me to test specifically feel free to let me know

Inderix:
When I installed PostMarketOS, I ended up with OpenRC, [PmOS is currently transitioning from OpenRC to Systemd](https://postmarketos.org/blog/2024/03/05/adding-systemd/), and I'm not sure where they are in that process. Journald does not appear in my system either. PostmarketOS provides `logread` from busybox, but this showed nothing related to Gnome Shell. I checked `dmesg` as well just in case, but it turned up empty too.
[...]

PR #1920, opened 2025-02-06
Preferences: Don't rely on shebangs to launch

ferdnyc:
Since launching JS modules via shebangs relies on features of the /usr/bin/env utility that aren't universally available, avoid it where possible.

The `gsconnect-preferences` "executable" (actually a JS module with a shebang line) is renamed to `gsconnect-preferences.js`, and both `prefs.js` and `daemon.js` will launch it by running `gjs -m gsconnect-preferences.js` with the extension root dir as the working directory, using `Gio.SubprocessLauncher`.

A small replacement `gsconnect-preferences` shell script is provided to replace the old launcher, and tries to do the right thing with working directories. However, it should not be relied on too heavily.

Partially addresses: #1879 (only prefs, not daemon.js)


src/gsconnect-preferences
```diff
@@ -1,104 +1,19 @@
-#!/usr/bin/env -S gjs -m
+#!/usr/bin/sh
```
daniellandau, line 1: I don't think that's guaranteed to exist at that path.

...is what I wanted to write, and then I made the mistake of going to actually look at what POSIX says, and now I'm just sad.

image

daniellandau: On my Ubuntu system here both `/bin/sh` and `/usr/bin/sh` are symlinks to `dash`. I once set my login shell to `/usr/bin/bash` on a `/usr` merged Arch years ago and couldn't log in after that because `/bin/bash` was the canonical name even though both could execute it. I'd guess `/bin/sh` is perhaps slightly more safe.

I think an executable file with no shebang will get passed to `sh` https://utcc.utoronto.ca/~cks/space/blog/unix/EmptyFileWhyTrue, but that's also quite a silly way to do it, and then we'd perhaps need to have a non-shebang ya rly comment stating what we are doing.

daniellandau:
I think the issue reference should be https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1897 with the last two numbers flipped.

PR #1922, opened 2025-02-08
Change function call `this._ensureAdapter();` to `await this._ensureAdapter();` in input.js. fix #1836

### LINKED ISSUE #1836: Presentation Control on Wayland unreliable

Plrrb:
The async function `this._ensureAdapter();` does not initialize `this._session` from null to the new `RemoteSession();` in time for the first button press to be registered (but the second press would be). It would not be registered because calling `this._session.pressKeys();` when `this._session` is null would fail, and not press any key. I have added 'await' to functions calling `this._ensureAdapter();` so that the code waits for the `this._session` to be ready for calling methods like `this._session.pressKeys();`, etc.

I also reverted the `SESSION_TIMEOUT` back to 15 because this fix means it does not have to be 300 anymore.

Fixes #1836

### DIFF
--- a/src/service/components/input.js
+++ b/src/service/components/input.js
@@ -10,7 +10,7 @@ import GObject from 'gi://GObject';
 import AtspiController from './atspi.js';
 
 
-const SESSION_TIMEOUT = 300;
+const SESSION_TIMEOUT = 15;
 
 
 const RemoteSession = GObject.registerClass({
@@ -341,7 +341,7 @@ export default class Controller {
 
                 this._session = new AtspiController();
 
-            // Mutter is available and there isn't another session starting
+                // Mutter is available and there isn't another session starting
             } else if (this._sessionStarting === false) {
                 this._sessionStarting = true;
 
@@ -386,60 +386,60 @@ export default class Controller {
     /*
      * Pointer Events
      */
-    movePointer(dx, dy) {
+    async movePointer(dx, dy) {
         try {
             if (dx === 0 && dy === 0)
                 return;
 
-            this._ensureAdapter();
+            await this._ensureAdapter();
             this._session.movePointer(dx, dy);
         } catch (e) {
             debug(e);
         }
     }
 
-    pressPointer(button) {
+    async pressPointer(button) {
         try {
-            this._ensureAdapter();
+            await this._ensureAdapter();
             this._session.pressPointer(button);
         } catch (e) {
             debug(e);
         }
     }
 
-    releasePointer(button) {
+    async releasePointer(button) {
         try {
-            this._ensureAdapter();
+            await this._ensureAdapter();
             this._session.releasePointer(button);
         } catch (e) {
             debug(e);
         }
     }
 
-    clickPointer(button) {
+    async clickPointer(button) {
         try {
-            this._ensureAdapter();
+            await this._ensureAdapter();
             this._session.clickPointer(button);
         } catch (e) {
             debug(e);
         }
     }
 
-    doubleclickPointer(button) {
+    async doubleclickPointer(button) {
         try {
-            this._ensureAdapter();
+            await this._ensureAdapter();
             this._session.doubleclickPointer(button);
         } catch (e) {
             debug(e);
         }
     }
 
-    scrollPointer(dx, dy) {
+    async scrollPointer(dx, dy) {
         if (dx === 0 && dy === 0)
             return;
 
         try {
-            this._ensureAdapter();
+            await this._ensureAdapter();
             this._session.scrollPointer(dx, dy);
         } catch (e) {
             debug(e);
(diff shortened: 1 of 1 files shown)


src/service/components/input.js
```diff
@@ -10,7 +10,7 @@ import GObject from 'gi://GObject';
 import AtspiController from './atspi.js';
 
 
-const SESSION_TIMEOUT = 300;
+const SESSION_TIMEOUT = 15;
```
ferdnyc, line 13: Even if we were to do this, it doesn't strike me that there's any benefit to dropping the session timer back down — as discussed in the issue (or maybe my previous PR), we could arguably remove it entirely. But it certainly doesn't need to be as short as 15 seconds. Constantly stopping and starting the remote desktop session is expensive.

src/service/components/input.js
```diff
@@ -341,7 +341,7 @@ export default class Controller {
 
                 this._session = new AtspiController();
 
-            // Mutter is available and there isn't another session starting
+                // Mutter is available and there isn't another session starting
```
ferdnyc, line 344: This change misaligns the comment with the others in the function / makes it appear to be a "closing note" from the `this.connection === null` branch, instead of a description of the `this._sessionStarting === false` branch (which is what it is).

If it were moved _below_ the `else if`, THEN indenting it to this level would be appropriate, but not if it's above.

src/service/components/input.js
```diff
@@ -386,60 +386,60 @@ export default class Controller {
...
      * Pointer Events
      */
-    movePointer(dx, dy) {
+    async movePointer(dx, dy) {
         try {
             if (dx === 0 && dy === 0)
                 return;
 
-            this._ensureAdapter();
+            await this._ensureAdapter();
```
ferdnyc, line 394: Is this a good idea? Making all of our input functions into _async_ functions feels like a recipe for PHENOMENAL lag and/or out-of-order events, at least under certain circumstances.

ferdnyc (changes requested):
Thanks for the PR, @Plrrb !

I have to admit I'm a bit dubious about this one. As I note in the comments, making all of our input functions _async_ feels like a recipe for epic lag or event-ordering problems.

Here are my questions:

1. Have you tested this?
2. Have you tested this "under load", meaning with significant rapid-fire pointer events and/or keypresses? (Like, typing multiple lines of text quickly.)
3. Were the input functions changed to be `async` for a reason? 
4. Is that reason something other than, "you can't use `await` in a non-`async` function"?

"Requested changes" primarily to remove the comment indentation change, although I still think lowering the timeout is a bad idea.

Plrrb:
>     1. Have you tested this?
I have been using this modified version every day for a week, and intent to use it further. to and have not had any problems, it is the sole way I control my TV-PC to watch YouTube.
>     2. Have you tested this "under load", meaning with significant rapid-fire pointer events and/or keypresses? (Like, typing multiple lines of text quickly.)
I have used the KDE connect app with my phone's software keyboard, the 'Compose Send' functionality, and right-click, middle click, left-click and double click. without issue.
>     3. Were the input functions changed to be `async` for a reason?
Just to use `await`. But there might be a better way to fix the bug.
>     4. Is that reason something other than, "you can't use `await` in a non-`async` function"?
Yes, but as described in my pull request, this fixes the bug of the first input not registering by waiting for the `this._session` to be ready.
> "Requested changes" primarily to remove the comment indentation change, although I still think lowering the timeout is a bad idea.
That's OK I thought the timeout was changed to 300 to fix #1836, and so I though changing it back to the previous value of 15 was intended by whoever set it there in the first place.

Sorry, the comment indentation change was a typo.

src/service/components/input.js
Plrrb: I tested this using an android keyboard (HeliBoard) that can type (no paste) the clipboard contents instantly, and did not have any out of order characters. I also tested this while downloading a 3GB file over WiFi (both phone and PC), and while it did have much more lag, especially when moving the mouse, I could not tell a difference between the changes. and there were still no out of order characters.

src/service/components/input.js
Plrrb: Sorry, that was a typo.

ferdnyc:
Written that way, I don't think we technically even need `this._sessionStarting` anymore, because there can only be one `_ensureAdapter()` call running at a time — since input events are still synchronous, the next one won't be able to start and make another `_ensureAdapter` call until the current event finishes. But, can't hurt to leave it in.

(Plus, maybe two different paired devices, both sending input events, could be running on different threads and cause it to double up, or something.)

Plrrb:
Wow! you went all out, at this point the pull request is all your code.
Changing `_ensureAdapter()` to be synchronous is a good way to fix it.
I applaud that you want to be confident in the safety and reliability of your code.
I was glad I could help debug this bug. Thank you.

ferdnyc:
> Wow! you went all out, at this point the pull request is all your code.

Yeah, sorry, turned out that the rewrite needed to go the sync route was more extensive than any other fix — but I think it's still worth doing. And I never would've gotten there without your prompting, so thanks for that.

(I've never really had a very deep understanding of Promises, TBH... and I still feel like I don't, but now at least I know they can be made synchronous with `.then()`, so that's something.)

I've submitted #1925 which includes my changes (similar to the above, but with some additional tweaks), I guess we can continue the review process from there. There's a testable artifact generated from the build, if you're interested in trying it out.

Thanks again!

ferdnyc:
I just merged #1925, so this should now be fixed. Thanks again for getting the ball rolling on this, @Plrrb !

PR #1925, opened 2025-02-14
Remote input: Synchronize _ensureAdapter

### LINKED ISSUE #1836: Presentation Control on Wayland unreliable

ferdnyc:
Building on #1922 by @Plrrb,

So that we don't miss the first input event whenever a new session needs to be started, have `_ensureAdapter` return the session object it creates. Resolve the implicit Promise for that object in its callers, using `.then()`.
    
Fixes #1836
Closes #1922


andyholmes (approved):
I haven't had time to test, but LGTM.

FYI we can eradicate Atspi if wanted, since RemoteDesktop has been in all distros for awhile (debian/ubuntu was the hold-up in the past).

ferdnyc:
@andyholmes Is ATSPI possibly still of any use to _non_-GNOME GSConnect users? If not I'm happy to tear it out, but I wouldn't want to break things for CLI users any more than we already have — not if I can avoid it, anyway.

ferdnyc:
Eh, I'll rip it out then, but in a separate PR titled accordingly, to give anyone who possibly _does_ still have use for it a better chance of spotting it and raising the alarm.

andyholmes:
Yeah, there's no rush, it's just one less `if`-branch and a blob of code I doubt ever gets used.

ferdnyc:
In the meantime, let's get this one in so that's dealt with.

PR #1927, opened 2025-02-26
Since GNOME 44, we shouldn't compile our settings schema

ferdnyc:
The extension system will compile it when installing the extension.

See: https://gjs.guide/extensions/upgrading/gnome-shell-44.html#gsettings-schema


andyholmes (approved):

andyholmes:
Now that I think about it, the `glib-compile-schemas ${ZIP_DIR}/schemas` command might have to be moved to the "install" block to avoid breaking the `install-zip` build target.

ferdnyc:
@andyholmes 

> Now that I think about it, the `glib-compile-schemas ${ZIP_DIR}/schemas` command might have to be moved to the "install" block to avoid breaking the `install-zip` build target.

Ooh, that's a good point.

Or, should the `install-zip` target be using `gnome-extensions install ${ZIP_FILE}`. instead of unzipping the directory ourselves? Now that we can rely on the `gnome-extensions` command being available.

andyholmes:
> (**Edit2:** Oh, wait, `install-zip` is only used for installing into `$HOME/.local/share/gnome-shell/extensions/`. So, yeah, now I'm back to "why are we doing this manually?")

Ah, that script pre-dates the `gnome-extensions` tool, but I'm fine with whatever works :)

ferdnyc:
@andyholmes 

> Ah, that script pre-dates the `gnome-extensions` tool,

Yeah, I should've said, "why are we _still_ doing this manually?" Now that there's a tool (and I've just tested, it does indeed do the compile for us at install time), I'll add a commit to switch us over to using that.

PR #1928, opened 2025-02-26
backport #1911 to GNOME 43 (based on v54)

ferdnyc:
It will, frankly, be a miracle if I haven't missed or broken anything here, and presumably the CI will need some major adjusting just to be able to test these, but:

This PR is an attempt to bring all of the changes from #1911 to the `backports/gnome-43` branch, which is based off of tag `v54` (the last GNOME 43 release of GSConnect).

For release as `v59`, to be followed by `backports/gnome-45` as `v60`, then `main` as `v61`.


ferdnyc:
OK, now that we have CI images tagged `gnome-43` and `gnome-45`, let's try using them to test these backports...

andyholmes:
This should be okay, the only thing I might question is if the encryption algorithms are available on older systems. I expect Debian et al. would backport things like that, though.

daniellandau:
> This should be okay, the only thing I might question is if the encryption algorithms are available on older systems. I expect Debian et al. would backport things like that, though.

I ran `openssl req -newkey ec -pkeyopt ec_paramgen_cu
rve:prime256v1 -subj '/O=andyholmes.github.io/OU=GSConnect/CN=foobar' -new -x509 -nodes` on a Debian stable container and it spat out a cert, so I think it's fine.

ferdnyc:
OK... I finally got both backport branches building _without_ these changes, so now we can actually test the PRs.

ferdnyc:
Had some too-modern code in there for ES2017, whoops.

ferdnyc:
Well, I've got this one down to only `testLanBackend` failing, but it's failing pretty spectacularly. Here's the contents of `gsconnect_backends_testLanBackend.test.txt`:

```text
1..7
ok 1 A LAN channel service can be started

(gjs:523): Gjs-WARNING **: 10:32:33.647: JS ERROR: Gio.TlsError: Peer failed to perform TLS handshake: The TLS connection was non-properly terminated.
_handshake/</<@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:600:44
runTests@/usr/libexec/installed-tests/gsconnect/minijasmine:160:29
@/usr/libexec/installed-tests/gsconnect/minijasmine:176:1

not ok 2 A LAN channel service can request and accept channels
# Message: Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL) in /usr/libexec/installed-tests/gsconnect/jasmine.js (line 7049)
# Stack:
not ok 3 A LAN channel service tracks active channels
# Message: Expected undefined to be defined.
# Stack:
#   @/usr/libexec/installed-tests/gsconnect/backends/testLanBackend.js:80:30
# Message: Expected undefined to be defined.
# Stack:
#   @/usr/libexec/installed-tests/gsconnect/backends/testLanBackend.js:83:31
not ok 4 A LAN channel service produces channels that can transfer packets
# Message: TypeError: localChannel is undefined in /usr/libexec/installed-tests/gsconnect/backends/testLanBackend.js (line 94)
# Stack:
#   @/usr/libexec/installed-tests/gsconnect/backends/testLanBackend.js:94:13
not ok 5 A LAN channel service produces channels that can transfer payloads
# Message: Error: Invalid value 'undefined' for property channel in object initializer. in /usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/core.js (line 512)
# Stack:
#   _init@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/core.js:512:15
#   Transfer@/usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/core.js:509:4
#   @/usr/libexec/installed-tests/gsconnect/backends/testLanBackend.js:109:35
ok 6 A LAN channel service can be stopped
ok 7 A LAN channel service closes active channels when stopped
```

ferdnyc:
Aaand, now #1929 is failing the same way (though with more output in its logs; I'll paste them there).

ferdnyc:
@andyholmes That's what I'm thinking — while Debian 12's `openssl` may have backported support for the newer algorithms, it's almost guaranteed that neither Fedora 37 or 39 did. It's also possible that GNOME 43 wouldn't support the new certs **even on an up-to-date Debian 12**. (I'm not sure if everything usable in `openssl` would automatically be available to GLib in older versions, or if support has to be backported into GLib as well. And then, if so, whether that was done.)

Which means we can't test code using the new certs on older Fedora releases. We could theoretically test it using Debian 12, to see whether **that** GNOME 43 will take the newer crypto.

Or, I could just drop the backporting of the crypto update (which isn't actually necessary to backport)... or even lose the protocol 8 backport _entirely_. Protocol 7 is still officially supported, at least for the moment. So I could limit the backporting to just the change in device ID format. That's the only element with any real **urgency** to backport.

ferdnyc:
These PR branches have become kind of a mess anyway, since I started them before I had the _target_ branches actually working. (This PR's irreconcilable conflicts are ample evidence of that.) So I think I'll start two new PRs, based off the _working_ `backports/gnome-43` and `backports/gnome-45` branches, and backporting **just** the changes required to migrate to the new device ID format. Let's see how that goes...

andyholmes:
> It's also possible that GNOME 43 wouldn't support the new certs even on an up-to-date Debian 12. (I'm not sure if everything usable in openssl would automatically be available to GLib in older versions, or if support has to be backported into GLib as well. And then, if so, whether that was done.)

It's possible. AFAIK glib-networking links to openssl, so that should be enough, but it may be there were other changes required in `glib-networking` itself to adapt that didn't get backported.

andyholmes:
I've backed out the certificate changes from `main`, so I guess we can do the same here. I did give this a shot in an `ubuntu-22.04` container and encountered no TLS errors, so :shrug:

ferdnyc:
You know, it strikes me that we should have a `Makefile` or a Meson target to generate the key files in `installed-tests/data/`, so that the commands for doing so are part of the project code.

In principle I know it's embedded in `src/service/init.js`, and you could even make the argument that it shouldn't live in two places. So, OK, maybe a meson target that _uses_ `src/service/init.js` to generate the `installed-tests/data/` certificate files.

Problem with that is, `src/service/init.js` imports `Config`, so it can't be run from the source directory.

...So now I'm working on a new file `src/utils/crypto.js` that can contain _just_ the code for calling the OpenSSL command, with no dependencies on any other project code, on files that only exist in the build directory / installed extension package, or on being loaded via ESM.

I figure, the commands to generate files and read their data can live _there_ in a file that's runnable even from the source tree, and even in command-line interactive `gjs` sessions.

My thinking is:

1. The functions provided there can be used to generate the certificates in the `installed-tests/data` directory, when new certificates are needed.
2. `src/service/init.js` can import those functions, and install wrappers around them as `Gio.TlsCertificate.new_for_paths()`, the `Gio.TlsCertificate` methods `common_name()` and `pubkey_der()`, and any other crypto code it monkeypatches.

..That first could be done (via `Makefile` rule, Meson target, or what have you) using a script that performs the moral equivalent of:

```bash
$ cd installed-tests/data
$ gjs -I ../../src
gjs> const Crypto = imports.utils.crypto;
undefined
gjs> Crypto.generate_cert_pair('openssl', 'local-certificate.pem', 'local-private.pem');
undefined
gjs> Crypto.generate_cert_pair('openssl', 'remote-certificate.pem', 'remote-private.pem');
undefined
gjs>
```

For the second, where `src/service/init.js` used to monkeypatch `Gio.TlsCertificate.new_for_paths()` with:

```js
Gio.TlsCertificate.new_for_paths = function (certPath, keyPath, commonName = null) {
    // Check if the certificate/key pair already exists
    const certExists = GLib.file_test(certPath, GLib.FileTest.EXISTS);
    const keyExists = GLib.file_test(keyPath, GLib.FileTest.EXISTS);

    // Create a new certificate and private key if necessary
    if (!certExists || !keyExists) {
        // If we weren't passed a common name, generate a random one
        if (!commonName)
            commonName = GLib.uuid_string_random().replaceAll('-', '_');

        const proc = new Gio.Subprocess({
            argv: [
                Config.OPENSSL_PATH, 'req',
                '-newkey', 'ec',
                '-pkeyopt', 'ec_paramgen_curve:prime256v1',
                '-keyout', keyPath,
                '-new', '-x509', '-nodes',
                '-days', '3650',
                '-subj', `/O=andyholmes.github.io/OU=GSConnect/CN=${commonName}`,
                '-out', certPath,
            ],
            flags: (Gio.SubprocessFlags.STDOUT_SILENCE |
                    Gio.SubprocessFlags.STDERR_SILENCE),
        });
        proc.init(null);
        proc.wait_check(null);
    }

    return Gio.TlsCertificate.new_from_files(certPath, keyPath);
};
```

Now it would instead be...

```js
const Crypto = imports.utils.crypto;

Gio.TlsCertificate.new_for_paths = function (certPath, keyPath, commonName = null) {
    // Check if the certificate/key pair already exists
    const certExists = GLib.file_test(certPath, GLib.FileTest.EXISTS);
    const keyExists = GLib.file_test(keyPath, GLib.FileTest.EXISTS);

    // Create a new certificate and private key if necessary
    if (!certExists || !keyExists)
        Crypto.generate_cert_pair(Config.OPENSSL_PATH, certPath, keyPath, commonName);
    return Gio.TlsCertificate.new_from_files(certPath, keyPath);
};
```

ferdnyc:
Aaanyway, closing these in favor of #1939 and #1940.

daniellandau:
> I've backed out the certificate changes from `main`, so I guess we can do the same here. I did give this a shot in an `ubuntu-22.04` container and encountered no TLS errors, so 🤷

I don't think the TLS problem is on the GNU/Linux side at all.

andyholmes:
> > I've backed out the certificate changes from `main`, so I guess we can do the same here. I did give this a shot in an `ubuntu-22.04` container and encountered no TLS errors, so 🤷
> 
> I don't think the TLS problem is on the GNU/Linux side at all.

I can't think of any reason the test suite would fail then?

daniellandau:
Ok now after having read the conversation more carefully, let me rephrase. My TLS problem wasn't on the GNU/Linux side, but clearly there's some TLS problem here. Perhaps the same, but would have to dig in more carefully to try to find out if it's the same or different.

When's v7 being deprecated? If it's happening while we still care about these shell versions we probably will need to fix this. My philosophy for backporting on serious issues (such as other side deprecating a protocol version, and GSConnect fully stopping working) is two latest Ubuntu LTS, Debian stable, and distros that keep up with latest such as Fedora and Arch, i.e. latest and second-latest shell.

andyholmes:
> My TLS problem wasn't on the GNU/Linux side, but clearly there's some TLS problem here.

I wonder if we should setup a distro matrix for the CI. Distributions have been in the news lately doing odd things, and I'm not honestly sure the divide between Debian and Ubuntu anymore. Pop_!OS seems to be hanging back a little longer with LTS this "LTS cycle".

> When's v7 being deprecated? If it's happening while we still care about these shell versions we probably will need to fix this.

I would not expect for awhile, since v8 is almost a patch on top of v7. The identity is re-exchanged over TLS, after the v7 over-TCP-exchange, so I'd guess a hypothetical v9 would be required to make it worth it.

PR #1929, opened 2025-02-26
backport #1911 to GNOME 45 (based on v56)

ferdnyc:
It will, frankly, be a miracle if I haven't missed or broken anything here, and presumably the CI will need some major adjusting just to be able to test these, but:

This PR is an attempt to bring all of the changes from #1911 to the `backports/gnome-45` branch, which is based off of tag `v56` (the last GNOME 45 release of GSConnect).

For release as `v60`, after `backports/gnome-43` as `v59`, followed by `main` as `v61`.

PR #1930, opened 2025-02-27
CI: Update workflow to support container selection

ferdnyc:
Update the primary CI workflow (renamed from 'main.yml' to 'ci.yml') to also build for PRs and pushes targeting branches with names of the form `backports/gnome-*`, and select the corresponding image from `GSConnect/gsconnect-ci` tagged as `gnome-##`.

This will allow testing branches against GNOME versions other than the default/current, without requiring separate workflows with near-duplicate content like the previous `gnome-3-3[68].yml`. (Those are deleted.)

I've tested this code in my own repos, and it should work accomplish what I'm trying to do here. (Which is, specifically...)

* Test any PRs targeting `backports/gnome-45` or pushes to `backports/gnome-45` against the (already-created) `gsconnect-ci` image tagged `gnome-45`
* Test any PRs targeting `backports/gnome-43` or pushes to `backports/gnome-43` against the (already-created) `gsconnect-ci` image tagged `gnome-43`
* Test all other PRs and pushes against the `gsconnect-ci` image tagged `latest`


ferdnyc:
@andyholmes @daniellandau Can I get a whut-whut? This demonstrably already works for PRs targeting `main` (like this one), but if I get it merged in I can use it to test the backport branches in #1928 and #1929 against their appropriate GNOME versions.

(Right now there's no CI for those PRs, because they don't target `main`. These modifications will change that — both in terms of making the CI run at all, and selecting the correct "legacy" container image for testing.)

.github/workflows/ci.yml
```diff
@@ -8,18 +8,43 @@ on:
...
+        run: echo "target=${{ github.base_ref }}" >> $GITHUB_OUTPUT
+      - id: select-image
+        env:
+          TARGET_BRANCH: ${{ steps.step1.outputs.target || steps.step2.outputs.target }}
+        run: |
+          if [[ \
+            ${{ startsWith(env.TARGET_BRANCH, 'backports/') && 1 || 0 }} == 1 \
+          ]]; then
+            IMAGE_ID=$(echo "$TARGET_BRANCH" | sed -e 's,^backports/,,')
+            echo "image_id=$IMAGE_ID" >> $GITHUB_OUTPUT
```
ferdnyc, line 38: (Yes, I arguably made things unnecessarily difficult for myself by naming the branches something _different_ from the container-image tag. But, oh well.)

daniellandau (approved):
I'm not good enough with github ci to know just by reading if this will work, but looks good enough to try.

ferdnyc:
@daniellandau Thanks!

> good enough with github ci to know just by reading if this will work

TBH neither am I, which is why I test-ran all of the logic at https://github.com/ferdnyc/bug-testing/pull/7

(Which is failing, because it's half-finished code that's expected to fail. But _before_ it failed it successfully echoed `test-branch` as the ID for the targeted `backports/test-branch`, and that was all I needed to see.)

andyholmes:
I'm also not a pro with GitHub's CI, but FWIW I've used the web editor & Co-pilot for a workflow recently and its actually pretty good; I think it has extra training for GitHub stuff.

ferdnyc:
I haven't tried Co-Pilot, but I have used their web editor — agreed, it's pretty good. Though it annoys me that they STILL haven't taught it [that complex `matrix` dimensions are perfectly valid](https://github.com/github/docs/issues/26469).

PR #1933, opened 2025-03-01
Support for GNOME 48

### LINKED ISSUE #1923: Support GNOME Shell 48
### Describe the bug

Please add support for GNOME Shell 48.

I tried updating the metadata.json myself but got this error when activating the extension

```js
TypeError: class heritage Calendar.NotificationMessage is not an object or null
```

See also https://gjs.guide/extensions/upgrading/gnome-shell-48.html

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### GSConnect version

58-1

### Installed from

OS package manager

### GNOME Shell version

48

Ubuntu 25.04 or Debian Unstable

### Paired device(s)

_No response_

### KDE Connect app version

_No response_

### Plugin(s)

_No response_

### Support log

```plain text

```

### Screenshots

_No response_

### Notes

_No response_

retrixe:
Fixes #1923 

Advisable not to merge this until GNOME 48 is out, in case something breaks down the line.

Food for thought:
- Grouped notifications also don't seem to work for me (and, come to think of it, I haven't seen notifs from other apps group up either) and maybe it's not necessarily desirable for GSConnect? I suppose we might need to make use of `NotificationMessageGroup` to enable this functionality, I haven't looked too deeply into any of it so far.


jbicha:
This doesn't work for me with GNOME Shell 47.3 on Debian Testing.

```
Extension gsconnect@andyholmes.github.io: TypeError: class heritage MessageList.NotificationMessage is not an object or null
Stack trace:
  @file:///usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/shell/notification.js:54:4
  async*@resource:///org/gnome/shell/ui/init.js:21:20
```

Also, you need to update metadata.json or this won't work for 48 either.

retrixe:
@jbicha thanks for the testing! I've added GNOME 48 to the manifest.json (sorry about that) and used the GNOME version to dynamically import `NotificationMessage` from the correct file.

I don't have a machine with GNOME 46/47 on me, but it should work correctly now on those versions (it works fine for me on GNOME 48 as well)

andyholmes:
LGTM, thanks! We haven't branched for gnome-47 yet, so we should probably...

- roll a release
- branch for gnome-47
- land this

cc @daniellandau @ferdnyc

retrixe:
Not sure branching is necessary since this PR retains support for GNOME Shell 46 and 47 as well

ferdnyc:
Agreed, if we can release a new update that supports all three releases, I'd rather _not_ branch. Having just worked on two backport branches, I can confirm that it is a giant PITA and I want as little as possible to do with it in the future.

src/shell/notification.js
```diff
@@ -7,14 +7,18 @@ import GLib from 'gi://GLib';
...
-import * as Calendar from 'resource:///org/gnome/shell/ui/calendar.js';
 import * as NotificationDaemon from 'resource:///org/gnome/shell/ui/notificationDaemon.js';
 
 import {gettext as _} from 'resource:///org/gnome/shell/extensions/extension.js';
 import {getIcon} from './utils.js';
 
+const {NotificationMessage} = Number(PACKAGE_VERSION.split('.')[0]) >= 48
+    ? await import('resource:///org/gnome/shell/ui/messageList.js') // GNOME 48
+    : await import('resource:///org/gnome/shell/ui/calendar.js'); // GNOME 46/47
+
```
ferdnyc, line 21: \*grumble\* I knew notifications were never going to stay part of the `Calendar` code because that simply made not one iota of sense. So, this is a good change, and really I'm annoyed because we went through two releases where all notifications _WERE_ `Calendar.NotificationMessage`s, before that finally got fixed!

ferdnyc (approved):
If this is the only change necessary to move from GNOME 47 → 48, then great, I'm all for it.

What's this about grouped notifications; is that a new GNOME 48 thing?

Not _sure_ if we'd want it for GSConnect notifications, anyway, because I'd think that if we were going to have notifications grouped, we'd want them grouped by the source app on the remote end. And definitely by device, as well, when there are multiple remote devices we're mirroring notifications for. Having all mirrored notifications from all devices bunched up into a single "GSConnect" group sounds... sub-optimal.

So, we'd have to figure out how to convey our multiple personalities (source identities) to the notification system, so that it could do the correct grouping.

ferdnyc:
@retrixe Is this good to be un-draft-ified? I could do it for you, but that feels presumptuous.

retrixe:
> What's this about grouped notifications; is that a new GNOME 48 thing?
> 
> Not _sure_ if we'd want it for GSConnect notifications, anyway, because I'd think that if we were going to have notifications grouped, we'd want them grouped by the source app on the remote end. And definitely by device, as well, when there are multiple remote devices we're mirroring notifications for. Having all mirrored notifications from all devices bunched up into a single "GSConnect" group sounds... sub-optimal.
> 
> So, we'd have to figure out how to convey our multiple personalities (source identities) to the notification system, so that it could do the correct grouping.

Yeah, [it's a GNOME 48 feature](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3012)

It seems my Fedora 42 system is running 48.beta, whereas notification grouping was added in 48.rc which released a few days ago, so that explains why I haven't seen any grouping yet... It seems it groups based on the `Source` object, so my assumption is that all GSConnect notifications will be grouped together by default, but I'd need to test it to know for sure

This reminds me of Phone Link on Windows which I used recently, which has its notifications grouped all under Phone Link, so I'm not sure it's _exactly_ a bad thing, though it would be better to group notifications by device if possible

> @retrixe Is this good to be un-draft-ified? I could do it for you, but that feels presumptuous.

Usually I'd be fine with it, but cause of 48.rc's notification grouping changes in `messageList.js`, I'm not sure if more changes are necessary. Fedora should update to 48.rc soon enough, but in the meantime... if someone can confirm if notifications still works and check how the grouping feature affects GSConnect, it could be fine to merge

maurerle:
So the grouping works fine using the current released version enabled on Gnome 48 on Debian testing 13 trixie.
I used `gsettings set org.gnome.shell disable-extension-version-validation true` to enable this.

Using the phone and "Send ping" action creates a grouped notification as seen here:
image

And here in the extended view:

image

So this looks fine to me. LGTM :)

I think the adjustment to Gnome 48 is not even required, as the class is still available in the previous namespace..? But this is just an assumption.

retrixe:
Fedora 42 updated to GNOME 48.rc as well, and it seems notification grouping works fine like you mentioned :tada: 

<details>
<summary>Image</summary>

Screenshot From 2025-03-14 20-16-52

</details>

Only concern is if we want to make notifications grouped separately for different devices (right now they're grouped for all devices together), but for initial GNOME 48 support, it seems things work fine

I'll push a fix for [this](https://gjs.guide/extensions/upgrading/gnome-shell-48.html#st-widgets-orientation) then unmark this as draft

src/shell/gmenu.js
```diff
@@ -92,12 +93,19 @@ export class ListBox extends PopupMenu.PopupMenuSection {
...
+        this.sub = Number(PACKAGE_VERSION.split('.')[0]) >= 48
+            ? new St.BoxLayout({
+                clip_to_allocation: true,
+                orientation: Clutter.Orientation.HORIZONTAL, // GNOME 48
+                visible: false,
+                x_expand: true,
+            })
+            : new St.BoxLayout({
+                clip_to_allocation: true,
+                vertical: false, // GNOME 46/47 (afaict, these don't have orientation)
```
ferdnyc, line 105: I believe they do, actually. `Clutter.Orientation` is definitely defined in GNOME 47 (just tested in `gjs`), and `St.BoxLayout.vertical` has been [documented](https://gnome.pages.gitlab.gnome.org/gnome-shell/st/property.BoxLayout.vertical.html) as "A convenience property for the `ClutterBoxLayout:vertical` property of the internal layout for `StBoxLayout`." for as long as it's existed. (Those docs were generated in February 2024, in fact. Which is why they still don't show `St.BoxLayout.vertical` as deprecated.)

ferdnyc (approved):
This seems fine (thanks @retrixe !) and I'm willing to merge it as-is if we need to get a release out, but I think we should look into whether we really need to do the version-dependent thing, or whether we can just use `St.BoxLayout.orientation` in all versions. My admittedly-cursory investigation so far makes the latter seem like a strong possibility.

src/shell/gmenu.js
ferdnyc: Er, half-finished thought. I meant to say, it's been documented [alongside the `orientation` property](https://gnome.pages.gitlab.gnome.org/gnome-shell/st/property.BoxLayout.orientation.html) for as long as it's existed.

ferdnyc: Nope, I'm wrong. `orientation` [was only added a month ago](https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/01541929eb02557f213d22d481388896bf6fc746).

The docs were generated with the 2024-02 version _of gi-docgen_, they're not themselves a year old. Whoops.

src/shell/gmenu.js
```diff
@@ -92,12 +93,19 @@ export class ListBox extends PopupMenu.PopupMenuSection {
         this.actor.add_child(this.box);
 
         // Submenu Container
-        this.sub = new St.BoxLayout({
-            clip_to_allocation: true,
-            vertical: false,
-            visible: false,
-            x_expand: true,
-        });
+        this.sub = Number(PACKAGE_VERSION.split('.')[0]) >= 48
```
ferdnyc, line 96: Part of me wants to create a "feature-test" constant for this, instead of having to do the string-processing and comparison every time. Something like,

```js
// in src/shell/utils.js maybe
export const HAVE_ST_ORIENTATION = new Boolean(
    Number(PACKAGE_VERSION.split('.')[0]) >= 48);

// Everywhere else
import {HAVE_ST_ORIENTATION} from '../shell/utils.js';
this.sub = new St.BoxLayout(
    HAVE_ST_ORIENTATION
    ? { ... }
    : { ... }
);
```

ferdnyc: (If it does go in `src/shell/utils.js`, the `PACKAGE_VERSION` import and the definition of `HAVE_ST_ORIENTATION` will need the same conditional logic as the `St` import, to protect `prefs.js`'s use pf that mpdule.)

retrixe: @ferdnyc the latest 2 commits should be of interest, let me know what you think

I added feature flags in the first one, and the second one eliminates dynamic imports (which started to feel really icky) by moving the setup code into the `utils` folder, which makes much more sense to me, since it's common shared code

ferdnyc (approved):
Thanks, @retrixe, this is awesome!

I have some minor cosmetic nitpicks, but nothing that affects any of the code itself. That's all great. Looking forward to getting this merged!

src/shell/gmenu.js
```diff
@@ -92,12 +92,19 @@ export class ListBox extends PopupMenu.PopupMenuSection {
...
-        this.sub = new St.BoxLayout({
-            clip_to_allocation: true,
-            vertical: false,
-            visible: false,
-            x_expand: true,
-        });
+        this.sub = HAS_ST_ORIENTATION
+            ? new St.BoxLayout({
+                clip_to_allocation: true,
+                orientation: Clutter.Orientation.HORIZONTAL, // GNOME 48
```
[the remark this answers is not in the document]
ferdnyc: Ah, ESLint [decided to get out of the whitespace game entirely](https://github.com/eslint/eslint/issues/17522) for version 9. But those rules mostly migrated to the `@stylistic` plugin which we're already using.

src/extension.js
```diff
@@ -22,9 +22,8 @@ import * as Device from './shell/device.js';
 import * as Keybindings from './shell/keybindings.js';
 import * as Notification from './shell/notification.js';
 import * as Input from './shell/input.js';
-import * as Utils from './shell/utils.js';
 import * as Remote from './utils/remote.js';
-import setup from './utils/setup.js';
+import setup, * as Setup from './utils/setup.js';
```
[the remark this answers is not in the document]
retrixe: This wouldn't be called `Setup.setup` though, since it's a default export, it's name-less and the actual name it'll have is `Setup.default`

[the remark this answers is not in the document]
retrixe: If that's okay with you, I can make the change though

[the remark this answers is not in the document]
ferdnyc: @retrixe  Ah, good catch thanks. Hrm. 

We already have other weird uses of that module, like:

`src/preferences/init.js` AND `src/service/init.js`:
```js
import setup, {setupGettext} from '../utils/setup.js';
```

In fact, there's no longer _any_ place where we import only the default function. (The ones you changed were the last holdouts, the others were already weird.)

Seems to me like having it be the default export doesn't make any sense. We never need _just_ that function anywhere. So why not remove the default, and just import the entire module as `Setup` when we need to call its functions?

Then we can just call `Setup.setup()`, `Setup.setupGettext()`, `Setup.ensurePermissions()`, `Setup.installService()`, etc. as needed.

That makes more sense to me; what do you think?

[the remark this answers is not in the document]
ferdnyc: (Apologies, BTW. I realize this is entirely out of scope for the GNOME 48 changes.)

In fact, on that note, I'm going to merge this as-is, since it's in a good state. We can address cleaning up the situation with `utils/setup.js` in a future PR.

ferdnyc:
Many thanks again, @retrixe, for all the work you put into this!

Merging, so that we're GNOME-48-ready in the upcoming release.

PR #1937, opened 2025-03-06
fix(daemon): be extra sure device ID is set to a valid format

andyholmes:


andyholmes:
Oh whoops, I guess I already did handle that.

PR #1938, opened 2025-03-06
revert: backout the certificate changes

andyholmes:


ferdnyc:
I just added a commit to both of my two (new) backport branches, that creates new `installed-tests/data/` certificates that use common names (and, therefore, device IDs) of:

* `gsconnect_local_test_device_cert`
* `gsconnect_remote_test_device_cert`

...Which are valid IDs under the new rules, but are also noticeable in logs and such. I wonder if it might make sense to use the same IDs for the `main` branch as well?

ferdnyc:
(Also, looks like this is failing because your new certificates have common names of:)

* local: `d2d2117a-49f7-479e-8daf-68c89a3f7ad2`
* remote: `50d2aadf-489f-4a40-b9ee-11cbd4abdd92`

andyholmes:
> I wonder if it might make sense to use the same IDs for the main branch as well?

No strong opinion. I think it probably makes sense to use the built-in utilities, since that's what would be used in practice, but I'm generally just jumping through hoops to make the CI pass :)

ferdnyc:
> I'm generally just jumping through hoops to make the CI pass :)

Considering that's been my entire past _week_, I can definitely sympathize with that.

I think all of these CI failures have demonstrated that if we're going to validate the certificate common names as device IDs, we need _some_ kind of error reporting when that validation **fails**. It's not even _remotely_ clear from any of the failures in `testLanBackend.js` that that's what's happening.

(I haven't even explored how it manages to manifest as a **timeout** instead of some sort of failure condition, yet, but I assume it has something to do with the mocking.)

andyholmes:
> (I haven't even explored how it manages to manifest as a **timeout** instead of some sort of failure condition, yet, but I assume it has something to do with the mocking.)

I just had all the functions return early. They could be modified to throw or emit a warning, although that might depend on the function in question. Some, I expect are GSource callbacks so need to return `true` or they'll stop looping, others might be `GSocketService` whose callbacks I think also have a meaningful return value.

daniellandau:
This doesn't solve the issue with my Android 11 device, so probably it doesn't make sense to do. The Android side is still offering a P256 key

andyholmes:
Closing this one; we can always back out later if we need to.

PR #1939, opened 2025-03-07
Backport device ID changes to GNOME 43 (based on v54)

ferdnyc:
This PR is an attempt to bring the device ID changes from https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1911 to the `backports/gnome-43` branch, which is based off of tag `v54` (the last GNOME 43 release of GSConnect).

For release as `v59`, to be followed by `backports/gnome-45` as `v60`, then `main` as `v61`.

PR #1943, opened 2025-03-08
fix(device): validate device name 

andyholmes:


andyholmes:
Probably want these for #1939 & 1940 as well :see_no_evil:

daniellandau (approved):

PR #1948, opened 2025-03-09
Backport device name changes to GNOME 42

andyholmes:


daniellandau (approved):
LGTM

ferdnyc:
So, if we have gnome 42, then THIS should be `v59`, then...

* `gnome-43` = `v60`
* `gnome-45` = `v61`
* `main` = `v62`

Issue #1931, opened 2025-03-01
GSConnect doesnt recognize phone (or the other way around)


Potatovibin:
### Describe the bug

GSConnect does not recognize or connect to KDE Connect on android.It used to work for a couple of days but now it decided to stop doing that.I tried to disconnect and reconnect but KDE Connect shows fedora-linux (my gsconnect is named as fedora) and connecting to that allows me to do all the stuff GSConnect can but i can't do any of the desktop commands or do stuff like ring/ping etc.Any Fix?

### GSConnect version

58

### Installed from

Extension Manager

### GNOME Shell version

47

### Linux distribution/release

Fedora 41

### KDE Connect app version

1.33.0

### Support log

```plain text
https://pastebin.com/VkKDt95D

(I think this is log file? Instructions for this section doesn't seem to work
```

### Screenshots

![Image](https://github.com/user-attachments/assets/38e96bd8-f338-48ac-9f49-75efa9139c3c)

peppot:
Same thing here. Tried to troubleshoot with a Android 16 phone as well as a Android 14 one. The Android 14 phone can see KDE Connect on my Android 16 device, but non can see or be seen by GSConnect on my desktop.
Version 58.

darqlama:
For me sync stopped after update of KDE connect to 1.33  from Play Store, GSConnect at version 58.

peppot:
Ah, I didn't realize that kde connect had been updated. Will try to downgrade when I get home

Arbee4ever:
I have the same issue with communication between my laptop and my Android Phone, while it works completely with the iPad (both with laptop and phone)

Arbee4ever:
After uninstalling and reinstalling an older Version using the official apk release from the KDE Connect website, it works

SmartFinn:
I've also couldn't connect GSConnect with KDE Connect 1.33.0 Android app. Rollback to 1.32.11 fixes the issue.

Cabus1982:
I'm experiencing the same issue. Lately, the extension stopped finding my smartphone, both on Fedora and Arch Linux. It's worth mentioning that this only happens with the extension version, because the desktop app works fine, but its integration level is terrible compared to the extension...

devkzidane:
Same here

peppot:
Downgrade to kde connect 1.32.2

Den sön 2 mars 2025 11:23j-fraga ***@***.***> skrev:

[...]

PR #1950, opened 2025-03-12
MPRIS: Use delay timer to fix update spam

### LINKED ISSUE #1919: MPRIS plugin packet spam
### Describe the bug

I've been running with debugging switched on pretty much constantly; I have 3 GB allocated to the systemd journal, it can handle it.

Recently, not long after I'd been watching some YouTube videos, I happened to look through the packet dumps in there, and I was shocked to see how much MPRIS traffic there was. So I decided to look into it.

I extracted and reformatted one section of the logs, then spent way too much time whipping up this visualization of the traffic using Python [rich](https://github.com/Textualize/rich). Then I screencast it and converted the video into an animated GIF.

Here's how this is set up: The logs are sanitized (video metadata replaced with generic strings), then parsed in and filtered down to _just_ the dumps that contain full MPRIS state packets (no position-only updates). Actually, they're also hand-filtered so that the visualization is focused on the three transitions between the four videos I watched.

1. Each packet is pretty-printed onto the screen, and as each one is shown it's counted on the progress bar below the JSON dump. 
2. Each time a packet comes in quick succession (timestamps equal or immediately adjacent, IOW 0-1 seconds delay), it's displayed 1/10 of a second after the previous one. 
3. Whenever there's a time skip (any packet with a timestamp that's more than one full second after the previous), a second progress bar appears while the updates pause for 3 seconds.

So:
* **rapid-fire updates**: packets that came in quick succession
* **3-second "Time passes" pause**: gap in the packet timestamps.
 
Simple enough.

But, watch the updates that happen, just in the process of starting to play the first video, then switching to the second, third, and fourth:

Image

Because we don't have the signals frozen on the MPRIS class while we're updating the data, _**each property change**_ generates a new MPRIS state packet that's sent to the remote device. So, every time YouTube auto-advances its playlist from one video to another, a storm of around 20 packets is sent to my phone as the various data elements in the JSON object are filled in one-by-one!

All in all, SEVENTY-EIGHT MPRIS state packets are generated just going from video A to B to C to D! And that's not _remotely_ all of the packets that were sent over that time  — I filtered out the ones where I stopped and started the player (an action that often  generated 3-4 packets, as potentially any or all of `"isPlaying"`, `"canPause"`, and `"canPlay"` will be separately updated each time.

Clearly, we need to be freezing the object to suppress all of those notify signals while we're updating its property data. Once we're done, we can thaw it and let a single new update packet be generated with _all_ of the changes.

### Steps to reproduce

1. Activate GSConnect "Generate Support Log" (or live dangerously and just run with debugging on all the time)
2. Play any content in MPRIS
3. Perform actions that cause metadata to be updated
4. Drown in packet spam

### Expected behavior

Outlined above.

### GSConnect version

58

### Installed from

other source

### GNOME Shell version

47.3

Fedora 41

### Paired device(s)

Samsung Galaxy A53

### KDE Connect app version

1.32.11

### Plugin(s)

MPRIS

### Support log

```plain text

```

### Screenshots

_No response_

### Notes

_No response_

ferdnyc:
OK, since we have a little time to wait until KDE Connect 1.33.2 makes it to user devices (see #1931), I figure let's get a couple of other fixes in to the next release. This is a fix for the MPRIS spam issue I reported a few weeks ago, details in #1919. I've tested it, and it seems to work well enough —  I sometimes still see _two_ MPRIS packets being sent in a row, but that's still a lot better than 10-15 like before.

My only real issue with it is that the 250ms timer I'm using to deduplicate metadata updates is just so... _arbitrary_. That alone makes it feel like an imperfect solution, but it's better than nothing so I don't want to let the perfect be the enemy of the good. Thus, this PR. Commit message description follows...

Because the MPRIS plugin listens for metadata changes from players, and because it will create a new update packet for each change and immediately send it to the paired device, something like a media change (advancing in a playlist, for example) can cause 10-15 property changes in a span of milliseconds, potentially firing a barrage of 10-15 `kdeconnect.mpris` packets at the device. Worse, because each of these packets contains a _full_ metadata dump, they're 10-15 nearly identical packets, except for a single changed value.

To mitigate this spam, start a 250ms timer running whenever the `_onPlayerChanged()` listener fires, and only create the update packet when the timer expires. Ignore any other `_onPlayerChanged` signals while the timer is running.

Responses to user actions or any `kdeconnect.mpris.request` packets received will be sent immediately, aborting any running timer, so responsiveness is not impacted.

Fixes #1919


ferdnyc:
Hmm, probably should've run the tests on it, before creating the PR. (I think I know what it's mad about, I commented out the initial `_sendPlayers()` the plugin does on connect — the remote device sends a request anyway, just like we do, so I figure why not wait for that?

But I'll probably have to adjust the test accordingly.

ferdnyc:
There was an actual bug in my code, now fixed.

daniellandau:
> My only real issue with it is that the 250ms timer I'm using to deduplicate metadata updates is just so... *arbitrary*

I don't mind that. IMO these throttles/debounces are always a value judgment on what *feels* fast enough while still being nice enough about resources.

daniellandau (approved):
LGTM

ferdnyc:
@daniellandau 

> I don't mind that. IMO these throttles/debounces are always a value judgment on what _feels_ fast enough while still being nice enough about resources.

Well, the real pisser of it is, MPRIS itself signals property changes using [the `org.freedesktop.DBus.Properties.PropertiesChanged` signal](https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-properties), which carries a dictionary of the updated values. So, the state changes _ARRIVE_ on the bus bundled together into a single event.

It's only when they're being translated into GObject properties, that they get broken out into separate values. Which then trigger a separate `notify::` signal for each one.

And apparently freezing the object doesn't help -- it _delays_ the `notify::` signals, but they still queue up and get fired all at once when the object's thawed. So the code ends up compensating for a messy interface that's _proxying_ for a good interface.

PR #1953, opened 2025-03-13
Gnome 42 protocol v8

daniellandau:
Finally set up bridge networking so I can test these on virtual machines instead of trying to scrounge up old computers. Seems to work on Ubuntu 22.04, let's see what the tests say.


andyholmes (approved):
LGTM

PR #1954, opened 2025-03-15
Fix ESLint config to ignore non-configured files

ferdnyc:
ESLint 9 has a default set of file globs that match all .js, .mjs, and .cjs files anywhere in or below the current directory. This has been a [source of consternation][1] for many users.

In our case, it means that if a Meson build directory is present in the repo as `build` / `_build` / whatever, the files in that directory will be linted, in addition to the repo source files.

That's merely a waste of time when the files are the same as the ones in the source directories. It's a _problem_ when the files aren't fresh, like the build directory from an older GSConnect version. Then, ESLint will start throwing thousands of errors for files it isn't even supposed to be looking at.

To keep ESLint runtimes down and prevent it scanning unintended files, add some configs that "undo" the `'**/*.js'` default glob by ignoring it, and only explicitly unignore the directories containing files we actually want to scan.

[1]: https://github.com/eslint/eslint/discussions/18850


daniellandau (approved):
LGTM

PR #1955, opened 2025-03-16
Gnome 43 protocol v8

daniellandau:
So far I just blindly cherry picked, haven't tested yet.


ferdnyc (approved):

daniellandau:
Tested on Debian 12, seems to work. Did find another issue https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1959, but that's not a blocker.

PR #1956, opened 2025-03-16
Gnome 45 protocol v8

daniellandau:
Similar to https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1955, so far a blind cherry pick job. Will try to test ASAP.

### DIFF
--- a/installed-tests/fixtures/backend.js
+++ b/installed-tests/fixtures/backend.js
@@ -283,7 +283,7 @@ var ChannelService = GObject.registerClass({
                 deviceId: this.id,
                 deviceName: this.name,
                 deviceType: 'desktop',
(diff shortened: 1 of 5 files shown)


ferdnyc:
I mean, the tests pass, that's a good sign.

ferdnyc (approved):

daniellandau:
Hmm, does any actual currently supported distro ship gnome 45?

daniellandau:
Fedora 39 and Ubuntu 23.10 are both end of life, so I don't know if it even makes sense to release this version.

daniellandau:
Let's do 45 too for completeness. I didn't test this one, but also I don't think anyone should be using this shell version either.

PR #1957, opened 2025-03-16
Remove the default export from utils/setup

ferdnyc:
We no longer ever need _just_ that function, so having it be the default export only makes the imports more awkward. Demote it to a regular named export, and import what we need from the module when we need it.

This grew out of #1933, where @retrixe moved additional functions into the module. That highlighted the (already-existing) awkwardness in how we consume the module's exports.

### DIFF
--- a/src/extension.js
+++ b/src/extension.js
@@ -23,7 +23,7 @@ import * as Keybindings from './shell/keybindings.js';
 import * as Notification from './shell/notification.js';
 import * as Input from './shell/input.js';
 import * as Remote from './utils/remote.js';
-import setup, * as Setup from './utils/setup.js';
+import * as Setup from './utils/setup.js';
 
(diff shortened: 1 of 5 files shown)


retrixe:
I wonder if `setupConfig` might be a less ambiguous name for the function, else LGTM 👍

daniellandau (approved):
LGTM

ferdnyc:
@retrixe 

Yeah, I waffled on the name. I also thought about going with nouns (`Setup.config()`, `Setup.gettext()`, but that became inconsistent with `Setup.ensurePermissions()` -- though I suppose that could've been `Setup.permissions()`, `Setup.getResource()`, `Setup.installService()`, etc.)

In the end, I'm glad @daniellandau merged as-is, or I'd probably still be second-guessing myself over the naming.

PR #1963, opened 2025-03-19
fix(daemon): fix device migration code

### LINKED ISSUE #1960: Opening Preferences window fails
### Describe the bug

I'm using GNOME 48 on Debian unstable. I've had GSConnect installed from the official extensions repository for a long time. But I wanted to get the GNOME 48 version working so I tried installing from git. I'm at commit a54b38ae.

The extension is installed and enabled, but it doesn't recognise my old pairings. And when I try to open the Preferences window nothing happens. I see this in the logs:
```
Mar 18 13:27:16 t14g1 dbus-daemon[1366]: [session uid=1000 pid=1366 pidfd=5] Activating service name='org.gnome.Shell.Extensions.GSConnect' requested by ':1.119' (uid=1000 pid=14442 comm="/usr/bin/gapplication action org.gnome.Shell.Exten")
Mar 18 13:27:16 t14g1 systemd[1333]: Started app-gnome-org.gnome.Shell.Extensions.GSConnect.Preferences-14442.scope - Application launched by gnome-shell.
Mar 18 13:27:16 t14g1 gjs[14448]: JS ERROR: TypeError: this.manager is undefined
                                  vfunc_dbus_unregister@file:///home/richard/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:309:9
                                  vfunc_handle_local_options@file:///home/richard/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:690:18
                                  _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34
Mar 18 13:27:16 t14g1 gjs[14448]: JS ERROR: TypeError: this._removeDevice is not a function
                                  vfunc_handle_local_options@file:///home/richard/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js:690:18
                                  _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34
```

### Steps to reproduce

1. Click "Mobile settings"
2. The Preferences window does not show

### Expected behavior

The Preferences window should show.

### GSConnect version

a54b38ae

### Installed from

other source

### GNOME Shell version

48

Debian unstable

### Paired device(s)

Moto G82

### KDE Connect app version

1.33.2

### Plugin(s)

_No response_

### Support log

```plain text

```

### Screenshots

_No response_

### Notes

_No response_

andyholmes:
Fix the migration code by always checking the certificate common
name and avoiding any dependency on the manager object.

Also fix the device ID and name validation for incoming
broadcasts and the settings path, when clearing settings.

closes #1960

### DIFF
--- a/installed-tests/suites/backends/testLanBackend.js
+++ b/installed-tests/suites/backends/testLanBackend.js
@@ -23,7 +23,6 @@ describe('A LAN channel service', function () {
         );
 
         local = new Lan.ChannelService({
-            id: localCert.common_name,
             certificate: localCert,
             port: 1717,
         });
@@ -34,7 +33,6 @@ describe('A LAN channel service', function () {
         );
 
         remote = new Lan.ChannelService({
-            id: remoteCert.common_name,
             certificate: remoteCert,
             port: 1718,
         });
--- a/po/POTFILES
+++ b/po/POTFILES
@@ -23,6 +23,7 @@ src/preferences/keybindings.js
 src/preferences/service.js
 src/service/daemon.js
 src/service/device.js
+src/service/init.js
 src/service/manager.js
 src/service/backends/lan.js
 src/service/plugins/battery.js
--- a/src/service/backends/lan.js
+++ b/src/service/backends/lan.js
@@ -361,15 +361,15 @@ export const ChannelService = GObject.registerClass({
                 return;
 
             // Reject invalid device IDs
-            if (!Device.validateId(this.identity.body.deviceId))
-                throw new Error(`invalid deviceId "${this.identity.body.deviceId}"`);
+            if (!Device.validateId(packet.body.deviceId))
+                throw new Error(`invalid deviceId "${packet.body.deviceId}"`);
 
-            if (!this.identity.body.deviceName)
+            if (!packet.body.deviceName)
                 throw new Error('missing deviceName');
 
             // Reject invalid device names
-            if (!Device.validateName(this.identity.body.deviceName))
-                throw new Error(`invalid deviceName "${this.identity.body.deviceName}"`);
+            if (!Device.validateName(packet.body.deviceName))
+                throw new Error(`invalid deviceName "${packet.body.deviceName}"`);
 
             debug(packet);
 
--- a/src/service/backends/lan.js
+++ b/src/service/backends/lan.js
@@ -173,13 +173,6 @@ export const ChannelService = GObject.registerClass({
     }
 
     _initCertificate() {
-        if (GLib.find_program_in_path(Config.OPENSSL_PATH) === null) {
-            const error = new Error();
-            error.name = _('OpenSSL not found');
-            error.url = `${Config.PACKAGE_URL}/wiki/Error#openssl-not-found`;
-            throw error;
-        }
-
         const certPath = GLib.build_filenamev([
             Config.CONFIGDIR,
             'certificate.pem',
--- a/src/service/backends/lan.js
+++ b/src/service/backends/lan.js
@@ -149,6 +149,10 @@ export const ChannelService = GObject.registerClass({
         return this._channels;
     }
 
+    get id() {
+        return this.certificate.common_name;
+    }
+
     get port() {
         if (this._port === undefined)
             this._port = PROTOCOL_PORT_DEFAULT;
@@ -184,12 +188,7 @@ export const ChannelService = GObject.registerClass({
 
         // Ensure a certificate exists with our id as the common name
         this._certificate = Gio.TlsCertificate.new_for_paths(certPath, keyPath,
-            this.id);
-
-        // If the service ID doesn't match the common name, this is probably a
-        // certificate from an older version and we should amend ours to match
-        if (this.id !== this._certificate.common_name)
-            this._id = this._certificate.common_name;
+            null);
     }
 
     _initTcpListener() {
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -50,36 +50,45 @@ const Service = GObject.registerClass({
         if (!Device.validateName(this.settings.get_string('name')))
             this.settings.set('name', GLib.get_host_name().slice(0, 32));
 
-        if (Device.validateId(this.settings.get_string('id')))
+        const [certPath, keyPath] = [
+            GLib.build_filenamev([Config.CONFIGDIR, 'certificate.pem']),
+            GLib.build_filenamev([Config.CONFIGDIR, 'private.pem']),
+        ];
+        const certificate = Gio.TlsCertificate.new_for_paths(certPath, keyPath,
+            null);
+
+        if (Device.validateId(certificate.common_name))
             return;
 
         // Remove the old certificate, serving as the single source of truth
         // for the device ID
         try {
-            Gio.File.new_build_filenamev([Config.CONFIGDIR, 'certificate.pem'])
-                .delete(null);
-            Gio.File.new_build_filenamev([Config.CONFIGDIR, 'private.pem'])
-                .delete(null);
+            Gio.File.new_for_path(certPath).delete(null);
+            Gio.File.new_for_path(keyPath).delete(null);
         } catch {
             // Silence errors
         }
 
         // For each device, remove it entirely if it violates the device ID
         // constraints, otherwise mark it unpaired to preserve the settings.
-        for (const deviceId of this.settings.get_strv('devices')) {
-            if (!Device.validateId(deviceId)) {
-                this._removeDevice(deviceId);
-            } else {
-                const settings = new Gio.Settings({
-                    settings_schema: Config.GSCHEMA.lookup(
-                        'org.gnome.Shell.Extensions.GSConnect.Device',
-                        true
-                    ),
-                    path: `/org/gnome/shell/extensions/gsconnect/device/${deviceId}/`,
-                });
-                settings.set_boolean('paired', false);
+        const deviceList = this.settings.get_strv('devices').filter(id => {
+            const settingsPath = `/org/gnome/shell/extensions/gsconnect/device/${id}/`;
+            if (!Device.validateId(id)) {
+                GLib.spawn_command_line_async(`dconf reset -f ${settingsPath}`);
+                Gio.File.rm_rf(GLib.build_filenamev([Config.CACHEDIR, id]));
+                debug(`Invalid device ID ${id} removed.`);
+                return false;
             }
-        }
+
+            const settings = new Gio.Settings({
+                settings_schema: Config.GSCHEMA.lookup(
+                    'org.gnome.Shell.Extensions.GSConnect.Device', true),
+                path: settingsPath,
+            });
+            settings.set_boolean('paired', false);
+            return true;
+        });
+        this.settings.set_strv('devices', deviceList);
 
         // Notify the user
         const notification = Gio.Notification.new(_('Settings Migrated'));
@@ -288,6 +297,9 @@ const Service = GObject.registerClass({
         // GActions & GSettings
         this._initActions();
 
+        // TODO: remove after a reasonable period of time
+        this._migrateConfiguration();
+
         this.manager.start();
     }
 
@@ -295,8 +307,6 @@ const Service = GObject.registerClass({
         if (!super.vfunc_dbus_register(connection, object_path))
             return false;
 
-        // TODO: remove after a reasonable period of time
-        this._migrateConfiguration();
         this.manager = new Manager({
             connection: connection,
             object_path: object_path,
--- a/src/service/device.js
+++ b/src/service/device.js
@@ -132,11 +132,11 @@ const Device = GObject.registerClass({
     }
 
     static generateId() {
-        return GLib.uuid_string_random().replaceAll('-', '_');
+        return GLib.uuid_string_random().replaceAll('-', '');
     }
 
     static validateId(id) {
-        return /^[a-zA-Z0-9_]{32,38}$/.test(id);
+        return /^[a-zA-Z0-9_-]{32,38}$/.test(id);
     }
 
     static validateName(name) {
--- a/src/service/init.js
+++ b/src/service/init.js
@@ -355,7 +355,7 @@ Gio.TlsCertificate.new_for_paths = function (certPath, keyPath, commonName = nul
     if (!certExists || !keyExists) {
         // If we weren't passed a common name, generate a random one
         if (!commonName)
-            commonName = GLib.uuid_string_random().replaceAll('-', '_');
+            commonName = GLib.uuid_string_random().replaceAll('-', '');
 
         const proc = new Gio.Subprocess({
             argv: [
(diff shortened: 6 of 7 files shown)


ferdnyc:
@andyholmes 

Ah. I was solving this slightly differently:

```diff
diff --git a/src/service/daemon.js b/src/service/daemon.js
index 3d86067d..e3ed156f 100755
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -66,20 +66,25 @@ const Service = GObject.registerClass({
 
         // For each device, remove it entirely if it violates the device ID
         // constraints, otherwise mark it unpaired to preserve the settings.
-        for (const deviceId of this.settings.get_strv('devices')) {
+        const deviceList = this.settings.get_strv('devices');
+        const newDeviceList = deviceList.filter(deviceId => {
             if (!Device.validateId(deviceId)) {
-                this._removeDevice(deviceId);
-            } else {
-                const settings = new Gio.Settings({
-                    settings_schema: Config.GSCHEMA.lookup(
-                        'org.gnome.Shell.Extensions.GSConnect.Device',
-                        true
-                    ),
-                    path: `/org/gnome/shell/extensions/gsconnect/device/${deviceId}/`,
-                });
-                settings.set_boolean('paired', false);
+                this.manager?._removeDevice(deviceId);
+                debug(`Invalid device ID ${deviceId} removed.`);
+                return false;
             }
-        }
+            const settings = new Gio.Settings({
+                settings_schema: Config.GSCHEMA.lookup(
+                    'org.gnome.Shell.Extensions.GSConnect.Device',
+                    true
+                ),
+                path: `/org/gnome/shell/extensions/gsconnect/device/${deviceId}/`,
+            });
+            settings.set_boolean('paired', false);
+            return true;
+        });
+        this.settings.set_strv('devices', newDeviceList);
+        debug(`Device list pruned to ${newDeviceList}`);
 
         // Notify the user
         const notification = Gio.Notification.new(_('Settings Migrated'));
@@ -90,6 +95,7 @@ const Service = GObject.registerClass({
 
         // Finally, reset the service ID to trigger re-generation.
         this.settings.reset('id');
+        debug('GSConnect device ID reset and paired devices migrated.');
     }
 
     get settings() {
@@ -295,12 +301,12 @@ const Service = GObject.registerClass({
         if (!super.vfunc_dbus_register(connection, object_path))
             return false;
 
-        // TODO: remove after a reasonable period of time
-        this._migrateConfiguration();
         this.manager = new Manager({
             connection: connection,
             object_path: object_path,
         });
+        // TODO: remove after a reasonable period of time
+        this._migrateConfiguration();
 
         return true;
     }
diff --git a/src/service/manager.js b/src/service/manager.js
index 49d51694..c84b95da 100644
--- a/src/service/manager.js
+++ b/src/service/manager.js
@@ -423,7 +423,7 @@ const Manager = GObject.registerClass({
      */
     _removeDevice(id) {
         // Delete all GSettings
-        const settings_path = `/org/gnome/shell/extensions/gsconnect/${id}/`;
+        const settings_path = `/org/gnome/shell/extensions/gsconnect/device/${id}/`;
         GLib.spawn_command_line_async(`dconf reset -f ${settings_path}`);
 
         // Delete the cache
@@ -431,8 +431,10 @@ const Manager = GObject.registerClass({
         Gio.File.rm_rf(cache);
 
         // Forget the device
-        this.devices.delete(id);
-        this.settings.set_strv('devices', Array.from(this.devices.keys()));
+        if (this.devices?.has(id)) {
+            this.devices.delete(id);
+            this.settings.set_strv('devices', Array.from(this.devices.keys()));
+        }
     }
 
     /**

```

I got sidetracked by a few of other issues.

The first was discovering that `_removeDevice` has had a longstanding bug in that it resets the wrong path. (Fixed, in the diff above.)

The second was discovering that `org.gnome.Shell.Extensions.GSConnect.id` seems to be _staying_ `''` after reset. (Not fixed.)

The third was discovering that the migration notification doesn't display: "g_application_send_notification: assertion 'g_application_get_is_registered (application)' failed". (Also not fixed.)

ferdnyc:
In fact, not only does `this.settings.reset('id')` NOT cause the ID to be regenerated (or at least, not stored to GSettings), but I just re-paired my phone to GSConnect (triggered from the phone), and `id` is _still_ blank. I have no idea WHAT ID it's using. (That doesn't seem to be logged on the Android side.)

andyholmes:
> The first was discovering that _removeDevice has had a longstanding bug in that it resets the wrong path. (Fixed, in the diff above.)

:+1: 

> The second was discovering that org.gnome.Shell.Extensions.GSConnect.id seems to be staying '' after reset. (Not fixed.)

That's probably intentional; the certificate should be the single source of truth for the device ID and can't be changed once generated. Probably the `id` setting could/should be removed, but I don't think we should be reading it anymore either way.

In that case, we'll have to load the certificate and read the common name to determine whether it represents a valid device ID.

> The third was discovering that the migration notification doesn't display: "g_application_send_notification: assertion 'g_application_get_is_registered (application)' failed". (Also not fixed.)

That should be fixed if `_migrateConfiguration()` is called after `Manager.start()` is called, since that happens in `Gio.Application.startup()` after `Gio.Application.dbus_register()` returned.

ferdnyc:
@andyholmes 

> > The second was discovering that org.gnome.Shell.Extensions.GSConnect.id seems to be staying '' after reset. (Not fixed.)
> 
> That's probably intentional; the certificate should be the single source of truth for the device ID and can't be changed once generated. Probably the `id` setting could/should be removed, but I don't think we should be reading it anymore either way.
> 
> In that case, we'll have to load the certificate and read the common name to determine whether it represents a valid device ID.

Yeah, because it turns out the `id` getting written to the settings is now _DIFFERENT_ from the one stored in `$HOME/.config/gsconnect/certificate.pem`. I think because one is getting generated during the same `daemon.js` run that performs the migration, and one is getting generated in the _next_ run when it discovers `id` is blank. So, that seems bad.

If it just stays unset, that's fine, but storing a wrong value there can't be a good move. And, of course, reading that setting is what triggers the migration code, so it'll run every time if it's not set.

andyholmes:
Okay, new patches should be good to go. If that looks all good, I'll backport to the other branhces.

ferdnyc:
@andyholmes 

Can you toss in the `_removeDevice` dconf path fix, while you're in there? Then I'll test a build off this branch.

ferdnyc:
@andyholmes Never mind, I took care of it. You should pull your branch if you're going to do any more work in it.

ferdnyc:
@andyholmes 

Hmm. Nope, there's still some kind of issue.

### Good
1. The migration ran
2. The notification now displays

### Bad
1. No new certificates were written to `$HOME/.config/gsconnect/` (tho the old ones were deleted)
2. All of the GSConnect-side preferences for my paired phone were lost on migration (even though it has a valid ID)

The journal contains this, from immediately after migration:

```text
Mar 19 22:22:09 gjs[3169394]: JS ERROR: Error: invalid deviceId "f4a30b02-3662-447c-b36d-6afa37f03977"
                                     _onIdentity@file:///home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:373:23
                                     _onIncomingIdentity@file:///home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:353:18
                                     _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34
```

I'm actually confused about the`_onIdentity` code. For example, 35b954c says it's meant to "validate incoming device names, per protocol v8". But the code is validating the _local_ device name and ID:

```js
// Reject invalid device IDs
if (!Device.validateId(this.identity.body.deviceId))
    throw new Error(`invalid deviceId "${this.identity.body.deviceId}"`);

if (!this.identity.body.deviceName)
    throw new Error('missing deviceName');

// Reject invalid device names
if (!Device.validateName(this.identity.body.deviceName))
    throw new Error(`invalid deviceName "${this.identity.body.deviceName}"`);
```

(The incoming identity is in `packet.body`, not `this.identity.body`.)

...But in a sense that's good, because the error shows that `this.identity.body.deviceId` was still our old ID with the hyphens, when it triggered.

andyholmes:
> I'm actually confused about the_onIdentity code. For example, https://github.com/GSConnect/gnome-shell-extension-gsconnect/commit/35b954c134af1a4b3ae46d0a7c919be18db4b98f says it's meant to "validate incoming device names, per protocol v8". But the code is validating the local device name and ID:

Ah, that was a copy-paste error from the LanChannel code, my bad.

ferdnyc:
@andyholmes 

For starters, I think the `ChannelService` setter for `id` needs to call `this.buildIdentity()` whenever it changes the `_id` value. Otherwise, if `identity` has already been read it'll still be memoized with the old `id` value.

ferdnyc:
> For starters, I think the `ChannelService` setter for `id` needs to call `this.buildIdentity()` whenever it changes the `_id` value. Otherwise, if `identity` has already been read it'll still be memoized with the old `id` value.

Actually, that won't work, because when the `id` is _updated_ in `backends/lan.js:_initCertificate` it's done by setting `this._id` directly. The setter never gets called.

andyholmes:
Let's just drop all references to `this.id` and `this._id` and always draw from the certificate.

andyholmes:
Sigh, this is going to take a decent amount of refactoring, you'll have to give me a bit.

andyholmes:
I believe I have it this time, probably.

Basically the the migration code was removing the certificate from underneath the manager, since we had to call `Manager.start()` before we could call `Manager._removeDevice()`.

We still have to call it in `Gio.Application.startup()` so the notification works, but it's now done before the call to `Manager.start()` avoiding a bunch of refactoring I'm not keen on doing.

andyholmes:
Marked as a draft until I have time to setup some test directories and devices.

daniellandau (approved):
LGTM

daniellandau:
I tested that this now shows the notification when updated on an old install and just for my sanity tested on a fresh virtual machine that it does work when installed totally fresh too.

ferdnyc:
Good timing! I was just testing this myself.

> I tested that this now shows the notification when updated on an old install and just for my sanity tested on a fresh virtual machine that it does work when installed totally fresh too.

Hrm. I didn't see the notification, but I upgraded using the `install-zip` path so maybe that borked things.

I'm still seeing settings getting lost on BOTH sides of the pairing, which I think we'll just have to live with at this point.

ferdnyc:
There's also the confusingly-same-named redundant device on the Android side that has to be manually removed, but definitely nothing we can do about that.

ferdnyc:
(Also, I just noticed we're still defaulting SFTP Automounting to enabled. We should REALLY change that.)

daniellandau:
> I'm still seeing settings getting lost on BOTH sides of the pairing, which I think we'll just have to live with at this point.

I hope now that we rip this band-aid (and got the no underscores either change in before release) we'll be fine for a while.

My process was to first turn off the daemon from the top-right menu of the shell, then install-zip, then manually start the daemon and I did get the notification like that.

ferdnyc:
> My process was to first turn off the daemon from the top-right menu of the shell,

@daniellandau \*nod\* I didn't want to do that, since users generally won't. And I don't think the e.g.o install process is equivalent to doing that, though I could be wrong since during upgrades e.g.o basically stages the upgrade for a restart.

What I actually did was run with v58 installed, `install-zip` a build from this branch, and then immediately log out and log back in again.

What I probably SHOULD'VE done is logged out, remotely console'd in to run the `install-zip`, then logged back in graphically. That feels like the closest simulation of e.g.o / update-notification upgrades.

daniellandau:
> e.g.o basically stages the upgrade for a restart.
> 
> What I probably SHOULD'VE done is logged out, remotely console'd in to run the install-zip, then logged back in graphically. That feels like the closest simulation of e.g.o / update-notification upgrades.

Yeah I think that'd be it. That was my thinking for why I did what I did.

daniellandau:
Anyways, alea iacta est, the releases are out, and we'll soon see how the users are getting on with this semi-breaking change :crossed_fingers:

ferdnyc:
I wonder if it's worth also putting a big, fat warning at https://extensions.gnome.org/extension/1319/gsconnect/ that when installing v59-v62, all pairings will be lost and all settings reset? (Along with an apology, it's an unavoidable one-time thing, etc, etc.)

ferdnyc:
Wow, that was quick, e.g.o is already showing an upgrade icon.

ferdnyc:
In that case, since I've been reset here anyway, I think I'll downgrade my F41 install back to v58 official with a dashed id, and test the e.g.o upgrade path directly.

daniellandau:
> In that case, since I've been reset here anyway, I think I'll downgrade my F41 install back to v58 official with a dashed id, and test the e.g.o upgrade path directly.

That'd be great :pray:

I added a NOTE to the extension page on e.g.o

ferdnyc:
What. The HELL?

I'm now running v62 from e.g.o. I've logged out and logged in TWICE now. But:

```bash
$ dconf read /org/gnome/shell/extensions/gsconnect/id
'cf61ec20-4bd6-4447-a917-c5af1e6cad2d'

$ cat ~/.config/gsconnect/certificate.pem|openssl x509 -noout -subject -inform pem              
subject=O=andyholmes.github.io, OU=GSConnect, CN=cf61ec20-4bd6-4447-a917-c5af1e6cad2d
```

ferdnyc:
I'm hacking a bunch of debug logging into `_migrateSettings()` and `vfunc_startup()` now, so I can see what's going on.

daniellandau:
That's an effect from https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1963/commits/46ef8387b671d4d44b55897cd5eed7f28ca7e176, allow to keep `-`, but don't generate it anymore.

ferdnyc:
Well, shit, that doesn't help! Then migration will never run for anyone.

ferdnyc:
Yeah, we might as well remove that disclaimer. As a result of that change, dashed IDs now validate, and `_migrateSettings()` will always bail when its `Device.validateId(certificate.common_name)` check passes.

ferdnyc:
(Except that means **we're** still creating technically-invalid DBus paths like `/org/gnome/Shell/Extensions/GSConnect/Device/cf61ec20-4bd6-4447-a917-c5af1e6cad2d`, when we pair GSConnect<->GSConnect.)

ferdnyc:
> > (Except that means **we're** still creating technically-invalid DBus paths like `/org/gnome/Shell/Extensions/GSConnect/Device/cf61ec20-4bd6-4447-a917-c5af1e6cad2d`, when we pair GSConnect<->GSConnect.)
> 
> I glanced at that while I was looking at `manager.js`, and it looks like we're covered:

Oh! Neat. And nothing cares that the bus path doesn't match the device ID? Cool... unexpected, but cool. Whatever works!

PR #1964, opened 2025-03-19
fix(daemon): fix device migration code (GNOME 45)

andyholmes:
Fix the migration code by always checking the certificate common
name and avoiding any dependency on the manager object.

Also fix the device ID and name validation for incoming
broadcasts and the settings path, when clearing settings.


daniellandau (approved):
LGTM

PR #1968, opened 2025-03-24
Fix send file functionality adding a " (1)" after the extension

### LINKED ISSUE #1851: Send File functionality adds a " (1)" after the file type on duplicate file
### Describe the bug

If in my PC's downloads i already have a file "hello.png" and i send a second file "hello.png" from my phone to my PC, it saves the second file as "hello.png (1)", which changes the file type and not the file's name, as it should

### Steps to reproduce

see above

### Expected behavior

The second file is renamed to "hello_1.png" or "hello (1).png)" (first solution preferred.

### GSConnect version

57

### Installed from

Extension Manager

### GNOME Shell version

46

Fedora Linux 40 (Workstation Edition)

### Paired device(s)

Motorola edge 30 ultra, Samsung Tab S9 FE +

### KDE Connect app version

1.31.1 / 1.29.0

### Plugin(s)

_No response_

### Support log

_No response_

### Screenshots

_No response_

### Notes

_No response_

screaminglineage:
Added a new file `src/service/utils/filename.js` which has a `makeUnique` function to take in a filename and add a `(1)` after the filename but before the extension. It also handles some common double extensions like `.tar.gz` and `.<ext>.bak` and increments previously added digits at the end, as mentioned [here](https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1851#issuecomment-2307536478)

This will fix #1851.

### DIFF
--- a/src/service/plugins/share.js
+++ b/src/service/plugins/share.js
@@ -10,6 +10,7 @@ import Gtk from 'gi://Gtk';
 
 import Plugin from '../plugin.js';
 import * as URI from '../utils/uri.js';
+import * as Filename from '../utils/filename.js';
 
 
 export const Metadata = {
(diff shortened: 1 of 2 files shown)


Jai-JAP:
I think this should instead add the ' (n)' before the first period for simplicity as this method does not handle triple extensions or further eg `.pkg.tar.zst`. This is not the perfect solution but atleast doesn't mess up with the file extensions. 

Also any of these wont work with hidden files without any extensions eg: '.notes'

PR #1978, opened 2025-03-29
_debugFunc: Fix caller position formatting

ferdnyc:
This is another fix that I had to re-do, as it unfortunately got lost during my dev-machine crash a couple of weeks ago.

The format of caller lines in debug traces is:

    FUNCTION@FILE:LINE:COLUMN

where 'FILE' is a local-file URI like

     file://home/user/.local/share/gnome-shell-extensions/gsconnect@andyholmes.github.io/service/device.js

Because that contains both an `@` sign and several `:`s, the previous regexp was splitting it up incorrectly. The new regexp:
- is bound to both the start and end of the string
- expects it to end with a colon followed by digits, twice (these two sets of digits become the LINE and COLUMN captures)
- expects it to start with one or more characters that is not an `@` sign (this becomes the FUNCTION capture)
- takes everything between the first `@` sign and the next-to-last `:` sign as the FILE capture
- strips off any leading 'file:///', when building the relative file path to the script logging the error

As a result, log lines go from this:

`Mar 28 23:47:36 org.gnome.Shell.Extensions.GSConnect[5890]: [file:broadcast:///home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js]: "Broadcasting to LAN"`

to this:

`Mar 29 15:55:47 org.gnome.Shell.Extensions.GSConnect[54447]: [service/backends/lan.js:broadcast:435]: "Broadcasting to LAN"`


ferdnyc:
The `journalctl` formatting still makes the logging look terrible, with those crazy indents on multi-line log messages, but there's nothing we can do about that. It's a choice between this:

```text
Mar 28 23:58:13 org.gnome.Shell.Extensions.GSConnect[5890]: [file:_readLoop:///home/ferd/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js]: Galaxy A53 5G: {
                                                              "id": 1743220693111,
                                                              "type": "kdeconnect.connectivity_report",
                                                              "body": {
                                                                "signalStrengths": {
                                                                  "1": {
                                                                    "networkType": "5G",
                                                                    "signalStrength": 1
                                                                  }
                                                                }
                                                              }
                                                            }
```

and this:

```text
Mar 29 16:12:40 org.gnome.Shell.Extensions.GSConnect[54447]: [service/device.js:_readLoop:361]: Galaxy A53 5G: {
                                                               "id": 1743279158810,
                                                               "type": "kdeconnect.connectivity_report",
                                                               "body": {
                                                                 "signalStrengths": {
                                                                   "1": {
                                                                     "networkType": "Unknown",
                                                                     "signalStrength": 0
                                                                   }
                                                                 }
                                                               }
                                                             }
```

Issue #1864, opened 2024-09-16
Pressing volume buttons on mobile have no effect on computer's volume


lo-yak:
### Describe the bug

Controlling computer's volume through mobile's volume buttons is not working for GSconnect.
Opening the KDE Connect app on Android and going into Devices section, where the volume slides are, doesn't work with volume buttons. Manually moving the slides works, though.
Even trying to select a device from there doesn't work. Clicking on one of them just "ticks" them, but the tick is not exclusive, you can just tick them all with no effect (and to un-tick them I have to manually move the volume)

### Steps to reproduce

1. Connect the computer with the mobile through GSConnect and KDE Connect.
2. Play something on computer.
3. On mobile, open KDE Connect app, go to Multimedia Control and then to Devices.
4. Try to control the volume using the mobile's buttons.

### Expected behavior

The volume can be controlled with mobile's volume buttons, like on KDE.

### GSConnect version

57

### Installed from

OS package manager

### GNOME Shell version

46

### Linux distribution/release

Debian Sid

### Paired device(s)

Moto G32

### KDE Connect app version

1.32.2

### Notes

No logs since they were totally useless, they were just about Fragments so nothing useful at all. Let me know if I can provide other useful informations please.

ferdnyc:
OK, I see [the commit](https://invent.kde.org/network/kdeconnect-android/-/commit/7080b43d2d44bdc1052bbb719fbbb7a6668b4785) in kdeconnect-android that added volume-key monitoring.

The difference between kdeconnect and GSConnect appears to be that when we send a `sinkList` to the device, we're not marking any of the sinks as `enabled` (which selects it as the default sink). Only when there's a sink marked as the default, will KDE Connect listen for volume key presses to change its volume.

(Tapping on the radio button next to a sink causes the app to send _us_ an `enabled: true` message for that sink, after which it expects to receive a `sinkList` showing that same sink as being `enabled`. The fact that we're not doing so is why the radio buttons seem to function a bit strangely — they never receive the `enabled` sink status flags they're looking for, so they stay in a weird state until the next time we send a response that _lacks_ `enabled`, at which point they reset.

Should be an easy fix, we just need to start tracking and including `enabled` values with each `sinkList` or `sink` update we send to the remote device. (The code already has the concept of a "current sink", but in our existing design it's simply whatever sink we last received a message for. We should be checking for an `enabled` value first, before blindly switching it every time we get a message.

[...]

PR #1992, opened 2025-04-16
FIXED: #1864 Pressing volume buttons on mobile have no effect on computer's volume 

realAbhi-web:
set enabled=true

### DIFF
--- a/src/service/plugins/systemvolume.js
+++ b/src/service/plugins/systemvolume.js
@@ -138,7 +138,7 @@ const SystemVolumePlugin = GObject.registerClass({
             muted: stream.is_muted,
             volume: stream.volume,
             maxVolume: this._mixer.get_vol_max_norm(),
-            enabled:true,
+            enabled: true,
         };
 
         this._cache.set(stream, state);
--- a/src/service/plugins/systemvolume.js
+++ b/src/service/plugins/systemvolume.js
@@ -138,6 +138,7 @@ const SystemVolumePlugin = GObject.registerClass({
             muted: stream.is_muted,
             volume: stream.volume,
             maxVolume: this._mixer.get_vol_max_norm(),
+            enabled:true,
         };
 
         this._cache.set(stream, state);
--- a/src/service/plugins/systemvolume.js
+++ b/src/service/plugins/systemvolume.js
@@ -20,7 +20,7 @@ try {
     GIRepository.Repository.prepend_library_path(typelibDir);
 
     Gvc = (await import('gi://Gvc')).default;
-} catch {}
+} catch { }
 
 
 export const Metadata = {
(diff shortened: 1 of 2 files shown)


src/service/plugins/systemvolume.js
```diff
@@ -138,6 +138,7 @@ const SystemVolumePlugin = GObject.registerClass({
             muted: stream.is_muted,
             volume: stream.volume,
             maxVolume: this._mixer.get_vol_max_norm(),
+            enabled: true,
```
ferdnyc, line 141: I don't _think_ this is a correct solution in the general case.

Looking at [the protocol specification](https://invent.kde.org/network/kdeconnect-meta/-/blob/master/protocol.md?ref_type=heads#systemvolume-plugin), [the KDE plugin code](https://invent.kde.org/network/kdeconnect-kde/-/blob/master/plugins/systemvolume/systemvolumeplugin-pulse.cpp), and to a lesser extent [the Android plugin code](https://invent.kde.org/network/kdeconnect-android/-/blob/master/src/org/kde/kdeconnect/Plugins/SystemVolumePlugin/SystemVolumePlugin.java), a consistent theme is that `enabled` (which is poorly named, IMHO) is used to indicate the current _default_ output stream, of which there can only be one.

So, just blindly setting every stream to `enabled: true` may _**seem** to work_ for the simple case of only a single stream outputting to a single audio device, but as soon as GSConnect attempts to cope with a more complex scenario (a system with two audio interfaces, or two media players running at the same time) I think things will quickly fall apart because the plugin is trying to publish everything everywhere all at once as default output(s). (And "default" should never be a plural to begin with.)

In fact, both the KDE and the Android code don't use `enabled` _**anywhere**_ except inside `systemvolume` packet payloads — when an `enabled` value is read out of, or written into, one of those packets, it's used as a proxy for a sink's `isDefault()` property. (Because it really should've always been named `default` instead of `enabled`.) So, caching an `enabled` property is probably a wrong move, because sinks/streams shouldn't have an `enabled` property at all. **ONE** sink may have a _default_ property, which is merely **communicated** to paired devices using the `enabled` packet flag.

I think more complexity will need to be added to the `service/components/pulseaudio.js` code to properly implement default-sink management, before GSConnect can make correct use of `enabled`.

ferdnyc (commented):
@realAbhi-web Thanks so much for contributing!

As I explain in my review comment, I have a feeling this won't work as a general solution for system volume management, because it depends on there being only a single audio device available, and will break as soon as conditions become even slightly more complex.

src/service/plugins/systemvolume.js
realAbhi-web: Thanks should I try to open an issue regarding that and try to work for it btw I do think I have connected two phones while checking though

realAbhi-web:
Hi now i have tried to fix so more than two devices are able to change volume i tried with two phone connected and my headphones connected it was working fine. Let me know if there's any more problem in this code.

realAbhi-web:
hey can you pls tell me if this solution is wrong so i can work on it again it will be helpful for me a lot

Issue #1985, opened 2025-03-31
Device fails to connect because its name is invalid


rexendevar:
### Describe the bug

I can only assume this is the reason my other computer won't appear in the list or connect. GSConnect repeatedly throwing `JS ERROR: Error: invalid deviceName` even though it let me enter the name on my other computer.

### Steps to reproduce

Use GSConnect to put an apostrophe in a device name, then try to connect to it from another computer with GSConnect

### GSConnect version

62

### Installed from

Extension Manager

### Support log

```plain text
Mar 30 23:42:16 gjs[2993]: JS ERROR: Error: invalid deviceName "Flynn's laptop"
                           _onIdentity@file:///home/flynn/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:372:23
                           _onIncomingIdentity@file:///home/flynn/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:345:18
                           _init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34

(it's just that over and over)
```

ferdnyc:
@daniellandau We should also discuss changing the spec with upstream -- naming a device "Someone's Something" is INCREDIBLY common, and it seems excessively restrictive to disallow that.

ferdnyc:
https://invent.kde.org/network/kdeconnect-meta/-/issues/27

Flameancer:
Hello I just ran into this issue. New install of GSconnect in Gnome 48. I have an iPhone 15 Pro Max running iOS 18.5. My iPhone's name is "Mr. Meeseeks". If I have the period and space character I get the message with the invalid device name. 

![Image](https://github.com/user-attachments/assets/14b1ff0f-4a57-46b7-911f-35cadee65d76)

However when I don't include any special characters it connects just fine. 

![Image](https://github.com/user-attachments/assets/311eb8c7-a28c-4331-ba6b-1c9363b5906f)

Appears that an update has affected the use of special characters.

ferdnyc:
@Flameancer 

> Appears that an update has affected the use of special characters.

That's correct, it's mostly down to KDE Connect tightening their protocol definition for what's permissible to use in a device name. (And GSConnect expanding its enforcement of those rules.)

[...]

PR #1997, opened 2025-04-23
Sanitize invalid names instead of rejecting

ferdnyc:
Inspired by KDE Connect Android, which seems to do the same thing. If an identity request comes in with an invalid device name, instead of simply rejecting the connection outright, attempt to remove the invalid characters from the name. Unless the name is _all_ invalid characters, allow the device to connect and be identified by the sanitized version of its name.

Related to #1985

> [!WARNING]
> With my desktop offline and my laptop as my only machine capable of running GSConnect, I have no way to test this. (I tried forcing the device name of my Android phone to some invalid name, but couldn't come up with a way to accomplish that.) So, I _believe_ it should work, but I would very much appreciate if someone **would** test it.
> 
> Using two GSConnect devices should be sufficient, just force the name of one of them to an invalid string with something like:
> 
> ```bash
> GSETTINGS_SCHEMA_DIR=\
> $HOME/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/schemas/ \
> gsettings set org.gnome.Shell.Extensions.GSConnect name "(badly) named"
> ```
> 
> And then attempt to connect to another GSConnect instance. Instead of being rejected for the invalid name, the device with that name should just show up as "badly named" on the other end.


albertvaka:
Yup, I think this is the best way of dealing with "bad" names.

ferdnyc:
Well, on that endorsement, I'll put together backports for the three legacy branches and get this ready to merge.

ferdnyc:
Rebased and squashed so I have a single commit I can cherry-pick to the backports.

stdedos:
Could this be pushed into the v62 - or some "new" version?

It is merged 1.5mo ago - but it is not released :confused:

ferdnyc:
@stdedos 

v62 is released and published to e.g.o, so it can't be updated.

We'd have to roll four new releases, potentially — v63 for GNOME 42, through v66 for GNOME 46-48. (Assuming we want to continue releasing extension updates for three older GNOME versions.)

Not that that can't be done; I think the delay was in waiting to see if there was anything else that needed to be changed (in the backport versions, as well as the current release).

There was actually one other change that went in as well, @andyholmes' #1998 and friends, merged the same day.

I agree those are worth releasing, and it doesn't seem like there's been anything else submitted as a candidate for the backports, so there's nothing we're waiting on. I'll look at putting together new release builds today, unless any of the other maintainers knows of a reason to hold off?

cc: @andyholmes @daniellandau @sonnyp

stdedos:
... I hate to sound like a broken record, but "are we there yet Papa-Smurf(s)"? 😥

(PS: My hints

image

https://extensions.gnome.org/extension/1319/gsconnect/
image

)

ferdnyc:
@stdedos See #2019, which is still open.

PR #1998, opened 2025-04-24
fix(daemon): store amended device name properly

andyholmes:
When storing an amended device name, use the correct setter for the GSettings value.

cc #1985

### DIFF
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -48,7 +48,7 @@ const Service = GObject.registerClass({
 
     _migrateConfiguration() {
         if (!Device.validateName(this.settings.get_string('name')))
-            this.settings.set('name', GLib.get_host_name().slice(0, 32));
+            this.settings.set_string('name', GLib.get_host_name().slice(0, 32));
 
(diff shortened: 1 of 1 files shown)


src/service/daemon.js
```diff
@@ -48,7 +48,7 @@ const Service = GObject.registerClass({
 
     _migrateConfiguration() {
         if (!Device.validateName(this.settings.get_string('name')))
-            this.settings.set('name', GLib.get_host_name().slice(0, 32));
+            this.settings.set_string('name', GLib.get_host_name().slice(0, 32));
```
ferdnyc, line 51: B'oh! (Separately-typed functions, man I do _not_ miss C programming.)

I wonder if, instead of using the hostname, this should _also_ run `Device.sanitizeName()` on the old value? To stay as close as possible to the user's previously-set name, while still overriding it.

(If that throws because there are no valid characters, then the hostname makes total sense as a last-resort fallback.)

ferdnyc (approved):
:+1: from me, either way (re: my question/suggestion below)

PR #1999, opened 2025-04-24
fix(daemon): store amended device name properly

andyholmes:
When storing an amended device name, use the correct setter for the GSettings value.

### DIFF
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -66,6 +66,9 @@ const Service = GObject.registerClass({
     }
 
     _migrateConfiguration() {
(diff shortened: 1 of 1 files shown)


ferdnyc:
Grr, GitHub. There's no way to "Approve" these en masse from the PR list — PRs can be multiple-selected, but the only actions available are "Mark: Open" or "Mark: Closed".

I _get_ why bulk approvals would, and _should_, be **discouraged**, but there are rare cases where they're useful.

ferdnyc (approved):

PR #2009, opened 2025-05-08
Update it.po

ArtyomZorin:
Courtesy of @albanobattistella

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -3,8 +3,8 @@ msgstr ""
 "Project-Id-Version: gsconnect\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2025-01-03 13:38-0500\n"
-"PO-Revision-Date: 2025-01-03 22:06\n"
-"Last-Translator: \n"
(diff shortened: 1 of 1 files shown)


po/it.po
```diff
@@ -189,7 +189,7 @@ msgstr "Touchpad.\n"
 "Trascina su questa area per spostare il cursore del mouse.\n"
 "Premi a lungo per trascinare il cursore del mouse.\n\n"
 "Un semplice clic verrà inviato al dispositivo accoppiato.\n"
-"Sinistra, centrale, tasto destro e rotelle della ruota."
+"Pulsante sinistro, centrale, destro e rotellina di scorrimento."
```
ferdnyc, line 192: Ugh, that _source_ message is awful English to begin with, we should really rewrite that into... like... full sentences. But, Google Translate agrees your translation is more in line with the source text, referencing 'buttons' rather than 'clicks'. (And who am I to argue with Google Translate?)

ferdnyc (approved):
Thanks @ArtyomZorin!

PR #2018, opened 2025-06-12
Release v66 for GNOME 46+

ferdnyc:


ferdnyc:
(Merging this is fine, if a reviewer wants to pull the trigger — the other three PRs are targeting different branches, and don't require reviews, so it makes no difference if they're merged before or after this PR. I'm only holding off on those to get reactions to this PR / the general notion of doing a new set of releases. ...But I can also do all four merges myself, once I get an approval here.)

PR #2029, opened 2025-07-15
Removed `www.` and trailing `/` from URLs

LucasDondo:


ferdnyc:
@LucasDondo 

> Removed www. and trailing / from URLs

A: ...Why, tho?

B: That is not even _remotely_ all this PR does. There are a ton of unnecessary style changes. We use eslint for code style enforcement, and the code in this PR violates the eslint style rules in 12 places.

(You can run eslint against the branch by checking it out locally, then running these shell commands:

```console
$ cd /path/to/this/branch
$ npm install
$ npx eslint
```

Fair warning, though: Even if those 12 things are corrected, I'm personally against merging this. I see no reason why these changes need to be made, or even _should_ be made, and some of them **cannot** be made.

For starters, URLs MUST NOT be changed in license texts for any reason. Not a single character of the files in `LICENSES/` can be edited, as those files are downloaded and validated by `reuse`.

daniellandau:
Thank you for contributing, but this is not useful, as @ferdnyc mentioned above. To add, github can't render the SVGs here, since the www is part of the namespace (is some other breakage), and matching URLs with www in the regexp is very much wanted behavior.

LucasDondo:
Thanks you both for the detailed feedback, I knew nothing (clearly) about those license-related rules, but thanks again for the feedback, and sorry for the wasted time!

PR #2031, opened 2025-07-19
Version bump for the development tree

ferdnyc:
One thing I neglected to do, when rolling the latest releases, was bump the `main` branch version to be higher than the latest release (`v66`). So, `main` is now on version `67`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,7 +4,7 @@
 
 project('gsconnect',
   'c',
-  version: '66',
+  version: '67',
   meson_version: '>= 0.59.0',
 )
 


ferdnyc:
I've enabled auto-merge, so if another maintainer could give this the standard rubber-stamp approval, it'll go right in.

ferdnyc:
I also added an edit to the release checklist. Previously it started with an unconditional "bump the project version" step, implying that we'd keep `main` at the old version number until it was time to release the next version. I think that's a mistake, and we should do as I've done here: Bump the version immediately after release.

If we end up doing more backport releases, we may have to _further_ bump the `main` version when rolling the next release, to make space for lower-numbered backport releases... but that's fine, that's a release-time thing. And it doesn't matter if development builds from `main` claiming to be version `67` eventually give way to a backport release that's the _real_ `v67`, because nobody is going to be mixing backports and bleeding-edge development builds on the same system — they're inherently incompatible. (Hence the backporting.) So there's no confusion there.

But if the development code continues to pretend to be the same version as the previous release, we lose the ability to distinguish development builds from release builds. That's a bad thing.

ferdnyc:
We can actually always identify a release build (assuming we use the bump-`main`-immediately-after-release method) because only the release build will contain an `org.gnome.Shell.Extensions.GSConnect.metainfo.xml` that references itself.

Development "version 67" builds made from `main` will stop at version 66, in the metainfo `<releases>` list. Only when `v67` gets an official release, will it appear in that list.

PR #2035, opened 2025-07-24
Migration message: Change "repaired" to "re-paired"

ferdnyc:
Let's not give anyone the impression we're breaking their devices, or anything. (And let's give the translators better odds of interpreting the message correctly.)


ferdnyc:
Urk. And, once again, doesn't count unless I regenerate the `.pot` file...

Issue #1215, opened 2021-12-02
Can't assign any keys to shortcuts


MagnusLupus:
**Describe the bug**

When I try to assign a shortcut key to any actions (i.e. Clipboard pull) the app says "*key combination here* is already being used." and selected keys don't get assigned.

**Steps To Reproduce:**

1. Go to 'Keyboard Shortcuts'
2. Click on 'any action'
3. Pick a key combination

**Expected behavior**

I expected the application to assign a key combination for an action.

**Screenshots**

![Screenshot from 2021-12-02 12-43-59](https://user-images.githubusercontent.com/95414451/144407033-bbb4b262-e122-4f4a-9f22-29aeaaa67595.png)

**Support Log**

Please generate a support log ([Instructions](../wiki/Help#generate-support-log)) and paste any messages related to this issue between the two ``` lines below.

```

GSConnect: 48 (system)
GJS:       17000
Session:   x11
OS:        Manjaro Linux
--------------------------------------------------------------------------------
-- Journal begins at Wed 2021-12-01 19:59:07 EET, ends at Thu 2021-12-02 12:46:45 EET. --
dets  02 12:46:00 touchegg.desktop[107344]: Error connecting to Touchégg daemon: Could not connect: Connection refused
dets  02 12:46:00 touchegg.desktop[107344]: Reconnecting in 5 seconds...
dets  02 12:46:04 gnome-shell[107022]: Can't update stage views actor <unnamed>[<MetaWindowGroup>:0x55e6e5184350] is on because it needs an allocation.
dets  02 12:46:04 gnome-shell[107022]: Can't update stage views actor <unnamed>[<MetaWindowActorX11>:0x55e6e7092b50] is on because it needs an allocation.
dets  02 12:46:04 gnome-shell[107022]: Can't update stage views actor <unnamed>[<MetaSurfaceActorX11>:0x55e6e7094a20] is on because it needs an allocation.
dets  02 12:46:05 touchegg.desktop[107344]: Error connecting to Touchégg daemon: Could not connect: Connection refused
dets  02 12:46:05 touchegg.desktop[107344]: Reconnecting in 5 seconds...
dets  02 12:46:06 /usr/lib/gdm-x-session[106953]: (EE) client bug: timer event22 debounce short: scheduled expiry is in the past (-12ms), your system is too slow
dets  02 12:46:10 touchegg.desktop[107344]: Error connecting to Touchégg daemon: Could not connect: Connection refused
[...]

PR #2037, opened 2025-07-27
Keyboard shortcuts without grabbing

daniellandau:
It's been [almost 4 years](https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/1215), I don't have the feeling that we'll actually implement a way of checking for conflicts. I suggest we instead drop grabbing the keyboard, so if there's already a global shortcut for something, then it just won't be passed to us in the dialog. I added a note in the dialog instructing the user to first disable shortcuts in GNOME Settings.

For some reason the [pot/po updating commands](https://github.com/GSConnect/gnome-shell-extension-gsconnect/wiki/Translating#creating-a-translation) wanted to update a bunch of line breaks. Is this something that goes back and forth, or a progress moves forward type of thing? If it's back and forth then I can minimize the diff to just the actually changed parts, but if it's just old vs new tools then maybe we can rip the bandaid and have the larger diff here.

### DIFF
--- a/data/ui/preferences-shortcut-editor.ui
+++ b/data/ui/preferences-shortcut-editor.ui
@@ -111,17 +111,6 @@ SPDX-License-Identifier: GPL-2.0-or-later
                     <property name="top_attach">0</property>
                   </packing>
                 </child>
-                <child>
-                  <object class="GtkLabel" id="conflict-label">
-                    <property name="can_focus">False</property>
-                    <property name="halign">center</property>
-                    <property name="hexpand">True</property>
-                  </object>
-                  <packing>
-                    <property name="left_attach">0</property>
-                    <property name="top_attach">1</property>
-                  </packing>
-                </child>
               </object>
               <packing>
                 <property name="name">confirm</property>
--- a/data/ui/preferences-shortcut-editor.ui
+++ b/data/ui/preferences-shortcut-editor.ui
@@ -52,6 +52,20 @@ SPDX-License-Identifier: GPL-2.0-or-later
           </packing>
         </child>
         <child>
+          <object class="GtkLabel">
+            <property name="can_focus">False</property>
+            <property name="halign">center</property>
+            <property name="hexpand">True</property>
+            <property name="label" translatable="yes">Note that you have to first disable any conflicting shortucts in GNOME Settings.</property>
+            <property name="visible">True</property>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="position">2</property>
+          </packing>
+        </child>
+        <child>
           <object class="GtkStack" id="stack">
             <property name="visible">True</property>
             <property name="can_focus">False</property>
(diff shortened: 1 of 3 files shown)


ferdnyc:
> For some reason the [pot/po updating commands](https://github.com/GSConnect/gnome-shell-extension-gsconnect/wiki/Translating#creating-a-translation) wanted to update a bunch of line breaks. Is this something that goes back and forth, or a progress moves forward type of thing? If it's back and forth then I can minimize the diff to just the actually changed parts, but if it's just old vs new tools then maybe we can rip the bandaid and have the larger diff here.

Oh, I never run the PO updating tool. It has different ideas about line breaks than Crowdin, so they'll fight about that kind of stuff. It's enough to update just the `.pot` file. (Which you'll notice, itself, has no line-break differences, just updates to translatable texts.) Then that'll get picked up automatically by Crowdin and presented to the translators, whose translations will be submitted as diffs against the `.po` files.

If you check in that huge `.po` diff now, the next Crowdin update will end up being similarly huge as it puts everything back the way it wants it.

ferdnyc:
So, here's the thing about shortcuts in general:

Google Chrome has somewhat-recently developed the ability to export global shortcuts to GNOME Shell. The shortcuts in question come from some extensions, and I know about this [because of a bug](https://discourse.gnome.org/t/issue-with-global-shortcuts-setup-popup-in-gnome-48/27946): Chrome shows a dialog on startup, **which seems to come from GNOME itself, _not_ from Chrome**, and if you "Cancel" that dialog instead of "Add"-ing it (even if you want to leave all of those shortcuts Disabled), it will be re-shown on every subsequent launch until you finally think to click "Add".

Here's the dialog that comes up, for me (as I said, the contents are based on the extensions installed — something it completely fails to make clear):

image

The reason I say it seems to be from GNOME is this: As that dialog indicates, the global keyboard combos in question _are managed using `gnome-control-center`_! The Chrome app has a "Global Shortcuts" entry in its listing:

image


Which, when clicked, brings up a very familiar-feeling dialog:

image

(The fact that dialog has different contents than the G-C-C one explains why I'm again getting it at every launch.)

...It seems like we should be able to get in on that same functionality, if we just know _how_. And then we presumably wouldn't need any sort of configuration interface, we'd just ask GNOME Shell to show our global-shortcuts dialog.

ferdnyc:
(BTW, clicking the Pencil icon in either of those dialogs brings up a very familiar-looking modal, another sign this is being provided by GNOME and not Chrome:)

image

ferdnyc:
RE: Translations, it would be great if we could configure the build system to _only_ include the POT update command, but not the PO update command. I don't think there is one, but it might be worth a Meson feature request.

daniellandau:
> RE: Translations, it would be great if we could configure the build system to only include the POT update command, but not the PO update command. I don't think there is one, but it might be worth a Meson feature request.

A good start for that is changing the wiki to not recommend the PO update command. I'll do that right now

daniellandau:
I'm not overly excited about looking into chrome internals to figure out a process, that IMO doesn't even feel less confusing than having to run conflict resolution manually.

daniellandau:
@ferdnyc how about this one? I still feel like this is an improvement to the current situation and the solution from Chrome above doesn't look good IMO and it doesn't seem likely that anyone will implement that for GSConnect anytime soon.

PR #2051, opened 2025-09-04
GNOME 49 support

### LINKED ISSUE #2048: GNOME 49 support
### Describe your request

First, thank you for maintaining and developing this great extension!

I'm using GNOME 49.beta, and it looks this extension was not able to start:

```
JS ERROR: Error: Requiring GIRepository, version 2.0: Typelib file for namespace 'GIRepository', version '2.0' not found
require@resource:///org/gnome/gjs/modules/esm/gi.js:16:28
@gi://GIRepository:3:25
```

### Proposed solution

I modified the requested version:

```diff
diff --git a/src/service/daemon.js b/src/service/daemon.js
index 0b7d89c9..835874ea 100755
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -7,7 +7,7 @@
 import Gdk from 'gi://Gdk?version=3.0';
 import 'gi://GdkPixbuf?version=2.0';
 import Gio from 'gi://Gio?version=2.0';
-import 'gi://GIRepository?version=2.0';
+import 'gi://GIRepository?version=3.0';
 import GLib from 'gi://GLib?version=2.0';
 import GObject from 'gi://GObject?version=2.0';
 import Gtk from 'gi://Gtk?version=3.0';
```

But then I got:

```
Module file:///(...)/gsconnect@andyholmes.github.io/service/daemon.js threw an exception
Sep 02 12:28:49 ThinkMatt gjs[18838]: JS ERROR: TypeError: (intermediate value).Repository.get_search_path is not a function
@file://(...)/gsconnect@andyholmes.github.io/service/init.js:68:42
```

Which is strange, I don't know if it is due to a bug in Beta or not. I was not sure how to fix this, so I simply did:

```diff
diff --git a/src/service/init.js b/src/service/init.js
index ea589432..5279c542 100644
--- a/src/service/init.js
+++ b/src/service/init.js
@@ -65,9 +65,10 @@ setupGettext();
 if (Config.IS_USER) {
     // Infer libdir by assuming gnome-shell shares a common prefix with gjs;
     // assume the parent directory if it's not there
-    let libdir = GIRepository.Repository.get_search_path().find(path => {
+    /*let libdir = GIRepository.Repository.get_search_path().find(path => {
         return path.endsWith('/gjs/girepository-1.0');
-    }).replace('/gjs/girepository-1.0', '');
+    }).replace('/gjs/girepository-1.0', '');*/
+    let libdir = '/usr/lib'
 
     const gsdir = GLib.build_filenamev([libdir, 'gnome-shell']);
 
```

Which might not be a good workaround because I'm now getting this error:

```
JS ERROR: PulseAudio not found: can't access property "connect", this._mixer is undefined
```

But it is enough for me for the moment, because most of the other features seem to work!

### Alternatives

I guess such changes will break things elsewhere, but I'm sorry, I don't know well the GS extensions :)

### GSConnect version

62, 66, Git

### Installed from

other source

### GNOME Shell version

49.beta

Ubuntu Questing

### Additional context

I tried different versions: the ones installed in the Ubuntu repo, the one from extensions.gnome.org, and the one from Git.

ferdnyc:
This is the first set of changes required to support GNOME 49, which includes a new GIRepository version 3.0 that requires some adaptation.

Until GNOME 49 is out of beta, we can't be certain these changes are complete, but they're what we know we need so far.

Fixes #2048

### DIFF
--- a/data/metadata.json.in
+++ b/data/metadata.json.in
@@ -3,6 +3,6 @@
     "name": "GSConnect",
     "description": "GSConnect is a complete implementation of KDE Connect especially for GNOME Shell with Nautilus, Chrome and Firefox integration. It does not rely on the KDE Connect desktop application and will not work with it installed.\n\nKDE Connect allows devices to securely share content like notifications or files and other features like SMS messaging and remote control. The KDE Connect team has applications for Linux, BSD, Android, Sailfish, iOS, macOS and Windows.\n\nPlease report issues on Github!",
     "version": @PACKAGE_VERSION@,
-    "shell-version": [ "46", "47", "48" ],
+    "shell-version": [ "46", "47", "48", "49" ],
     "url": "@PACKAGE_URL@/wiki"
 }
--- a/src/service/components/pulseaudio.js
+++ b/src/service/components/pulseaudio.js
@@ -15,8 +15,17 @@ let Gvc = null;
 try {
     // Add gnome-shell's typelib dir to the search path
     const typelibDir = GLib.build_filenamev([Config.GNOME_SHELL_LIBDIR, 'gnome-shell']);
-    GIRepository.Repository.prepend_search_path(typelibDir);
-    GIRepository.Repository.prepend_library_path(typelibDir);
+
+    if (GIRepository.Repository.hasOwnProperty('prepend_search_path')) {
+        // GNOME <= 48 / GIRepository 2.0
+        GIRepository.Repository.prepend_search_path(typelibDir);
+        GIRepository.Repository.prepend_library_path(typelibDir);
+    } else {
+        // GNOME 49+ / GIRepository 3.0
+        const repo = GIRepository.Repository.dup_default();
+        repo.prepend_search_path(typelibDir);
+        repo.prepend_library_path(typelibDir);
+    }
 
     Gvc = (await import('gi://Gvc')).default;
 } catch {}
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -7,11 +7,17 @@
 import Gdk from 'gi://Gdk?version=3.0';
 import 'gi://GdkPixbuf?version=2.0';
 import Gio from 'gi://Gio?version=2.0';
-import 'gi://GIRepository?version=2.0';
 import GLib from 'gi://GLib?version=2.0';
 import GObject from 'gi://GObject?version=2.0';
 import Gtk from 'gi://Gtk?version=3.0';
 import 'gi://Pango?version=1.0';
+
+// GNOME 49 uses GIRepository 3.0
+import('gi://GIRepository?version=3.0').catch(() => {
+    import('gi://GIRepository?version=2.0').catch(() => {});
+});
+
+import('gi://GioUnix?version=2.0').catch(() => {}); // Set version for optional dependency
 
 import system from 'system';
 
@@ -21,8 +27,6 @@ import Config from '../config.js';
 import Device from './device.js';
 import Manager from './manager.js';
 import * as ServiceUI from './ui/service.js';
-
-import('gi://GioUnix?version=2.0').catch(() => {}); // Set version for optional dependency
 
 
 /**
--- a/src/service/init.js
+++ b/src/service/init.js
@@ -62,10 +62,22 @@ const extensionFolder = GLib.path_get_dirname(serviceFolder);
 setup(extensionFolder);
 setupGettext();
 
+
 if (Config.IS_USER) {
     // Infer libdir by assuming gnome-shell shares a common prefix with gjs;
     // assume the parent directory if it's not there
-    let libdir = GIRepository.Repository.get_search_path().find(path => {
+    let gir_paths;
+
+    if (GIRepository.Repository.hasOwnProperty('get_search_path')) {
+        // GNOME <= 48 / GIRepository 2.0
+        gir_paths = GIRepository.Repository.get_search_path();
+    } else {
+        // GNOME 49+ / GIRepository 3.0
+        const repo = GIRepository.Repository.dup_default();
+        gir_paths = repo.get_search_path();
+    }
+
+    let libdir = gir_paths.find(path => {
         return path.endsWith('/gjs/girepository-1.0');
     }).replace('/gjs/girepository-1.0', '');
 
--- a/src/service/plugins/systemvolume.js
+++ b/src/service/plugins/systemvolume.js
@@ -16,8 +16,17 @@ let Gvc = null;
 try {
     // Add gnome-shell's typelib dir to the search path
     const typelibDir = GLib.build_filenamev([Config.GNOME_SHELL_LIBDIR, 'gnome-shell']);
-    GIRepository.Repository.prepend_search_path(typelibDir);
-    GIRepository.Repository.prepend_library_path(typelibDir);
+
+    if (GIRepository.Repository.hasOwnProperty('prepend_search_path')) {
+        // GNOME <= 48 / GIRepository 2.0
+        GIRepository.Repository.prepend_search_path(typelibDir);
+        GIRepository.Repository.prepend_library_path(typelibDir);
+    } else {
+        // GNOME 49+ / GIRepository 3.0
+        const repo = GIRepository.Repository.dup_default();
+        repo.prepend_search_path(typelibDir);
+        repo.prepend_library_path(typelibDir);
+    }
 
     Gvc = (await import('gi://Gvc')).default;
 } catch {}


ferdnyc:
@matttbe 

Now that this is done building in the CI, you should be able to download the resulting artifact (https://github.com/GSConnect/gnome-shell-extension-gsconnect/actions/runs/17475549307/artifacts/3931225926) and install it using `gnome-extensions install gsconnect@andyholmes.github.io.zip` like any other extension, if you're interested in testing it out.

ferdnyc:
You might need the `-f` flag to `install`, actually, to force it to overwrite your installed GSConnect.

matttbe:
@ferdnyc great, thank you for this PR!

I used the source code and installed the new version via `ninja` (I already had everything setup). It looks like this new version is working as expected, thanks!

FYI, When the extension is started, I have the following errors in `gjs` logs:

```
$ journalctl -o cat -b0 /usr/bin/gjs
JS ERROR: My DumbPhone: Gio.IOErrorEnum: Host key verification failed
_promisify/proto[asyncFunc]/</<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:454:45
_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34
### Promise created here: ###
_handleMount@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/sftp.js:229:24
async*handlePacket@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/sftp.js:122:26
handlePacket@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:446:25
_readLoop@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:370:22
async*setChannel@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:351:18
_onChannel@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/manager.js:267:20
channel@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/core.js:436:19
_onIncomingChannel@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:245:18
async*_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34
```

Followed by this one repeated 4 extra times:

```
JS ERROR: My DumbPhone: Gio.IOErrorEnum: Unknown error 255
link /home/$USER/.ssh/known_hosts to /home/$USER/.ssh/known_hosts.old: Permission denied
_removeHostKey@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/sftp.js:286:27
async*_handleMount@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/sftp.js:239:18
async*handlePacket@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/plugins/sftp.js:122:26
handlePacket@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:446:25
_readLoop@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:370:22
async*setChannel@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/device.js:351:18
_onChannel@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/manager.js:267:20
channel@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/core.js:436:19
_onIncomingChannel@file:///home/$USER/.local/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/backends/lan.js:245:18
async*_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34
```

I also don't know if it is responsible for creating many `~/.ssh/known_hosts.<RANDOM>` files, but I guess yes when looking at the timestamps.

I don't know if I had these errors before GNOME 49, so I guess they were already there, and maybe linked to the "Filesystem Expose" option. It is possible I had it disabled before, but because I wiped KDEConnect data when I noticed GSConnect was no longer working, it got re-enabled, and caused that (still the issue with `ssh-rsa` I suppose).

If these error messages are not linked, please ignore :) (or please tell me if you prefer I open a new bug report).

ferdnyc:
> If these error messages are not linked, please ignore :) (or please tell me if you prefer I open a new bug report).

Well, they're definitely coming from GSConnect. The first one is a host key verification error, not uncommon for SSH connections. So common, in fact, that GSConnect has code to automatically attempt to address those failures.

The second error is that code failing. Normally I'd say that it looks like you don't have the `ssh-keygen` command installed, but the "permission denied" error it's hitting makes me certain that you **do** have it installed, and it's not able to do its job.

GSConnect uses `ssh-keygen` to remove host keys from the `$HOME/.ssh/known_hosts` file when they could be causing verification errors. (Which it appears yours are.) When `ssh-keygen` removes keys from the file, it writes a backup of the previous state to `$HOME/.ssh/known_hosts.old`.

It looks to me like it can't move your `$HOME/.ssh/known_hosts` file to `$HOME/.ssh/known_hosts.old`, so it's failing to clean up after itself and leaving the temporary file it made of the modified version lying around.

What do you see, permissions/ownership wise, for these two files?

```bash
ls -l $HOME/.ssh/known_hosts{,.old}
```

If either of them is owned by another user (`root`?) or non-writable, that could be causing this. Oh, you should also check the directory permissions: `ls -ld $HOME/.ssh`.

My results, for comparison:

```console
$ ls -ld $HOME/.ssh
drwx------. 4 ferd ferd 4.1k Sep  5 05:40 /home/ferd/.ssh

$ ls -l $HOME/.ssh/known_hosts{,.old}
-rw-------. 1 ferd ferd 56k Sep  5 05:40 /home/ferd/.ssh/known_hosts
-rw-------. 1 ferd ferd 57k Jul  2 22:11 /home/ferd/.ssh/known_hosts.old
```

If the permissions are any more open than `0600` (`0700` for the directory), ssh will freak out and refuse to function. But the directory and files do need to be owned by your user and writable by your user (and nobody else), in order to modify the `known_hosts` list.

Once you've got the permissions sorted out on the files it cares about, you can delete any of the temporary trash it's littered the directory with. Those temp files will be the modified versions, unsuccessfully installed, so your unmodified `known_hosts` will still be intact. Once `ssh-keygen` can access what it needs to, GSConnect will make the necessary deletions.

ferdnyc:
@matttbe 

As far as these changes go, can you check whether volume control is working, since obviously that's the plugin most affected by the GNOME 49 changes?

If you open up the Media Player controls in KDE Connect and switch to the Devices tab, you should see the volume control for your desktop's default audio out. Changing the volume there should affect the desktop. If you have the GNOME user menu open, you can even watch the volume slider move in sync.

matttbe:
> > If these error messages are not linked, please ignore :) (or please tell me if you prefer I open a new bug report).
> 
> Well, they're definitely coming from GSConnect. The first one is a host key verification error, not uncommon for SSH connections. So common, in fact, that GSConnect has code to automatically attempt to address those failures.
> 
> The second error is that code failing. Normally I'd say that it looks like you don't have the `ssh-keygen` command installed, but the "permission denied" error it's hitting makes me certain that you **do** have it installed, and it's not able to do its job.

Thank you for the explanations!

> GSConnect uses `ssh-keygen` to remove host keys from the `$HOME/.ssh/known_hosts` file when they could be causing verification errors. (Which it appears yours are.) When `ssh-keygen` removes keys from the file, it writes a backup of the previous state to `$HOME/.ssh/known_hosts.old`.

Ah yes, I can reproduce the issue when launching `ssh-keygen -f ... -R ...` command.

> It looks to me like it can't move your `$HOME/.ssh/known_hosts` file to `$HOME/.ssh/known_hosts.old`, so it's failing to clean up after itself and leaving the temporary file it made of the modified version lying around.
> 
> What do you see, permissions/ownership wise, for these two files?
> 
> ```shell
> ls -l $HOME/.ssh/known_hosts{,.old}
> ```
> 
> If either of them is owned by another user (`root`?) or non-writable, that could be causing this. Oh, you should also check the directory permissions: `ls -ld $HOME/.ssh`.
> 
> My results, for comparison:
> 
> ```
> $ ls -ld $HOME/.ssh
> drwx------. 4 ferd ferd 4.1k Sep  5 05:40 /home/ferd/.ssh
> 
> $ ls -l $HOME/.ssh/known_hosts{,.old}
> -rw-------. 1 ferd ferd 56k Sep  5 05:40 /home/ferd/.ssh/known_hosts
> -rw-------. 1 ferd ferd 57k Jul  2 22:11 /home/ferd/.ssh/known_hosts.old
> ```
>
> If the permissions are any more open than `0600` (`0700` for the directory), ssh will freak out and refuse to function. But the directory and files do need to be owned by your user and writable by your user (and nobody else), in order to modify the `known_hosts` list.

I appreciate your nice reply!

I have exactly the same permissions (indeed, SSH enforces them).

> Once you've got the permissions sorted out on the files it cares about, you can delete any of the temporary trash it's littered the directory with. Those temp files will be the modified versions, unsuccessfully installed, so your unmodified `known_hosts` will still be intact. Once `ssh-keygen` can access what it needs to, GSConnect will make the necessary deletions.

I think I found out the error:
- `strace` shows that `ssh-keygen` cannot do a `link` to create the `old` file:
  ```
  link("known_hosts", "known_hosts.old") = -1 EACCES (Permission denied)
  ```
- I can run `link ~/.ssh/known_hosts ~/.ssh/known_hosts.old` without issue
- `lsattr ~/.ssh` doesn't show anything suspicious
- but AppArmor decided to block it according to `sudo journalctl -k -f -o cat`
  ```
  audit: type=1400 audit(1757066591.115:454): apparmor="DENIED" operation="link" class="file" profile="ssh-keygen" name="/home/$USER/.ssh/known_hosts.old" pid=19750 comm="ssh-keygen" requested_mask="l" denied_mask="l" fsuid=1000 ouid=1000 target="/home/$USER/.ssh/known_hosts"
  ```

I will report this error to AppArmor devs, sorry for the noise.

matttbe:
> As far as these changes go, can you check whether volume control is working, since obviously that's the plugin most affected by the GNOME 49 changes?
> 
> If you open up the Media Player controls in KDE Connect and switch to the Devices tab, you should see the volume control for your desktop's default audio out. Changing the volume there should affect the desktop. If you have the GNOME user menu open, you can even watch the volume slider move in sync.

@ferdnyc good point, I forgot to mention I tried before, and it works. To be precise, I can see and control most devices (laptop, screen), except the main one I use (USB) :-)
But I guess that's another issue :)

matttbe:
> To be precise, I can see and control most devices (laptop, screen), except the main one I use (USB) :-)

Mmh, I just restarted the extension with debug enabled, I know I can see the main one... Sorry for the noise, I probably missed it!

ferdnyc:
> I forgot to mention I tried before, and it works. To be precise, I can see and control most devices (laptop, screen), except the main one I use (USB) :-)
> But I guess that's another issue :)

Yeah, for the purposes of this PR any control at all is good enough, it shows that Gvc is still loading properly.

The missing USB audio device is strange, though. If it's the default device and it shows up in PipeWire, I wonder why it wouldn't be visible to KDE Connect? We might need some Generate Support Logs captures, see what GSConnect is sending over in its `kdeconnect.systemvolume` packets.

But, as you say, that one's definitely a separate issue.

matttbe:
> I will report this error to AppArmor devs, sorry for the noise.

FYI, it was linked to [this issue](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2116288).

daniellandau:
I'm slightly uneasy about straying very far from the [only supports the latest stable release of GNOME](https://github.com/GSConnect/gnome-shell-extension-gsconnect/wiki/Installation#install-from-git) policy. I think this is still ok, but let's not make it a policy to try to support very many versions of the shell from the same GSConnect.

I'll try to test this on some of those versions over the weekend.

ferdnyc:
> I'm slightly uneasy about straying very far from the [only supports the latest stable release of GNOME](https://github.com/GSConnect/gnome-shell-extension-gsconnect/wiki/Installation#install-from-git) policy. I think this is still ok, but let's not make it a policy to try to support very many versions of the shell from the same GSConnect.

My take is, if we can support multiple versions in **the same** release with minimal effort, we'd be jerks _not_ to.

I definitely don't want to make any more backport releases! GNOME 42, 43, and 45 can continue on with the GSConnect they have today. But our current code, by all accounts, _is_ compatible with GNOME 46 - 49, and the requirements to keep it compatible are relatively minor. (Just the GIRepository 2.0 vs. 3.0 adjustments in this PR, at least so far.)

I don't want to bend over backwards to support older releases, but I also do actively _want_ to keep the current release compatible with as much as possible, when possible.

ferdnyc:
(I'm especially reluctant to be too draconian about the only-one-GNOME-release support policy, because it _**increases**_ the likelihood that we'll end up having to backport fixes. If the next release only supported GNOME 49, and a security issue came up that we felt we couldn't leave unfixed in earlier releases, then we'd have to do a backport update to v66 for GNOME 46-48 users. If v67 still supports those releases as well, then we just update the one release.)

daniellandau (approved):
Tested on Ubuntu 24.04, 24.10 (I know that's out of support, but that's what I had as a VM at hand) and Fedora rawhide. This doesn't cause problems with those, but I wouldn't do a release yet as GSConnect was buggy on rawhide, I'll post a separate issue.

ferdnyc:
Thanks @daniellandau, good to have confirmation!

daniellandau:
Here's the separate issue: https://github.com/GSConnect/gnome-shell-extension-gsconnect/issues/2053

daniellandau:
> I'm especially reluctant to be too draconian about the only-one-GNOME-release support policy, because it increases the likelihood that we'll end up having to backport fixes. If the next release only supported GNOME 49, and a security issue came up that we felt we couldn't leave unfixed in earlier releases, then we'd have to do a backport update to v66 for GNOME 46-48 users. If v67 still supports those releases as well, then we just update the one release.

I think we agree here almost completely. I'm advocating for reasonable case by case middle ground. My thing is, with backports you have to test separately on the old shells, but with a single version supporting multiple shells you have to test old shells for every release, or hope that nothing breaks for them on accident. Again, I also don't want to go too draconian, but not too far in the other end of the spectrum either.

PR #2059, opened 2025-09-12
Reactivate Dependabot and add npm (for eslint)

ferdnyc:
But configure it to only scan weekly, so it hopefully isn't so impatient.


ferdnyc:
Administratively merging without review, as no project files are impacted, only repo configuration.

PR #2070, opened 2025-09-13
Fix link in OpenSSL infobar

ferdnyc:
I managed to screw up the URL in the infobar twice over — I typoed the organization name as "gsconect", and used the wrong fragment for the document section.

All of the existing translations will have to be corrected as well, but I can do that myself through Crowdin.


ferdnyc:
Merging, the only change is to a URL.

ferdnyc:
Ugh, Crowdin is so not happy with me right now.

ferdnyc:
OK, I've restored the four existing translations we had of the message, with the corrected URL.

**#TIL** that Crowdin, unlike gettext, apparently _doesn't keep_ outdated translations around — at least, I couldn't find them in the interface. So the tiny change I made to the source string meant I had to re-enter the full translations from scratch. (Fortunately, the uncorrected versions were still in the repo's `.po` files.)

ferdnyc:
> **#TIL** that Crowdin, unlike gettext, apparently _doesn't keep_ outdated translations around — at least, I couldn't find them in the interface.

I did finally find the old strings, in the "Translation Memory". (Which I suppose makes sense, the name kind of gives it away.) Though I'm still not clear on how/if I could've used that to update the _new_ translations of the edited source string.

PR #2072, opened 2025-09-15
Update Russian translation

OlesyaGerasimenko:


ferdnyc:
Thanks @OlesyaGerasimenko!

I've  uploaded the new translation file to Crowdin (since it prefers that translations flow that way) and added you to the "Translator Credits" message. Just waiting for Crowdin to sync with GitHub now, to make sure it opens a PR with the updates.

Issue #2079, opened 2025-10-01
Switching profile on android while connected causes mpv to crash


lo-yak:
### Describe the bug

Okay, this is a very odd one.
I'm running GrapheneOS on my Pixel 7a, on which I use two different user profiles. KDEConnect is installed on the main profile.
The thing is, when my laptop and my Pixel are connected through GSConnect and I'm playing something on mpv, if I switch to the secondary profile on the Pixel, the moment it shows the prompt for the PIN then mpv on my laptop crashes. I can reopen mpv then and resume the video just fine, even while using the secondary profile. I can leave the profile and switch back to the main one too without any trouble, it crashes **only** if laptop and Pixel are connected and I switch from the main one, the one which the laptop is connected to, to the secondary.

### Steps to reproduce

1. Connect laptop and Pixel
2. Play something with mpv on laptop
3. Switch to the secondary profile on Pixel
4. mpv on laptop crashes

### Expected behavior

I think it is expected that it loses the connection, but I not that mpv is crashing.

### GSConnect version

67

### Installed from

GNOME Extensions website

### GNOME Shell version

48

### Linux distribution/release

Ubuntu 25.04

### Paired device(s)

Pixel 7a

### KDE Connect app version

1.33.4

### Support log

```plain text
GSConnect: 67 (user)
GJS:       18201
Session:   wayland
OS:        Ubuntu 25.04
--------------------------------------------------------------------------------
ott 02 00:27:03 org.gnome.Shell.Extensions.GSConnect[4716]: [service/device.js:_readLoop:369]: Pixel 7a: {
                                                              "id": 1759357624296,
                                                              "type": "kdeconnect.connectivity_report",
                                                              "body": {
                                                                "signalStrengths": {
                                                                  "3": {
                                                                    "networkType": "GSM",
                                                                    "signalStrength": 2
                                                                  }
                                                                }
                                                              }
                                                            }
ott 02 00:27:08 org.gnome.Shell.Extensions.GSConnect[4716]: [service/device.js:_readLoop:369]: Pixel 7a: {
                                                              "id": 1759357628628,
                                                              "type": "kdeconnect.mpris.request",
                                                              "body": {
                                                                "player": "mpv",
                                                                "action": "Stop"
                                                              }
                                                            }
ott 02 00:27:08 org.gnome.Shell.Extensions.GSConnect[4716]: [service/device.js:sendPacket:477]: Pixel 7a: {
[...]

PR #2083, opened 2025-10-08
Revert "MPRIS: Use delay timer to fix update spam"

ferdnyc:
This reverts commit 83c857fb4527b2d886d8705ad0b1ed8962a5fcc5.

The MPRIS consolidation code I added (#1950) isn't working out. Even if it's not the cause of the crash issues reported in #2079 — which is still unclear — I'm also seeing things like wrong track durations and other "confusion" in the information being reported to the paired device. So, even when it's working right, it's not working _right_.

Part of the problem is that, while my code slowed down the packet traffic from GSConnect to the paired device, it didn't change the nature of the packets being sent: Every change, even just a single property, results in a full update of all possible properties (either immediately, with the old code, or after a delay, with my code), which I think is the bigger problem.

With this revert, we'll go back to sending a full update packet **each** time a property changes in the local player. (Except for position, which is handled separately.)

It would be preferable, when there are property changes in the active player. to send a packet containing _only_ the updated property or properties. This revert doesn't address that, just eliminates the delay timer / consolidation code that hasn't been working out in practice.

### DIFF
--- a/src/service/plugins/mpris.js
+++ b/src/service/plugins/mpris.js
@@ -43,8 +43,6 @@ const MPRISPlugin = GObject.registerClass({
         this._transferring = new WeakSet();
         this._updating = new WeakSet();
 
-        this._queueTimers = new Map();
-
         this._mpris = Components.acquire('mpris');
 
         this._playerAddedId = this._mpris.connect(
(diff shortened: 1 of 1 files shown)


daniellandau (approved):
I haven't tested or read through, but I guess a plain git revert to old functionality should be fine.

ferdnyc:
I've been testing for a while, everything seems fine. No errors in the logs, and more correct metadata in the KDE Connect control interface. 

I had to reconcile two or three conflicts, but in both cases that involved just restoring the version from before the reverted commit.

I'll keep banging on it a bit more, and maybe wait to see if @lo-yak is able to try it out to see if it fixes that crash issue. (Though even if it doesn't, I'll still merge it and then try to find another angle on that issue.)

ferdnyc:
@lo-yak reports no change to the crashes in #2079, but I'm still going to merge this anyway while we continue to hunt that one down. As I said, the consolidation code has its own issues.

PR #2087, opened 2025-10-20
Adding a "Log exchanges" settings to enable or disable packet logs

neutrino38:
I started to twick the GSConnect extension and noticed that all packets exchanged with KDECOnnect were in journalctl by default. A lot of sensitve and private info can be recorded that way without user's consent.

I added a device setting that allows the user to enable or disable the logging of theses exchanges. By default, they are disabled.
Capture d’écran du 2025-10-20 23-59-17

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -23,7 +23,7 @@ libexecdir = join_paths(prefix, get_option('libexecdir'))
 localedir = join_paths(prefix, get_option('localedir'))
 sysconfdir = get_option('sysconfdir')
 
-gschemadir = join_paths(datadir, 'glib-2.0', 'schemas')
+gschemadir = join_paths(extdatadir, 'schemas')
 
 
 # GNOME Shell LIBDIR


andyholmes:
Hi, the debug logging can already be controlled by the global `debug` setting:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/a10961d895840984f7be05e6211434044278b8d4/data/org.gnome.Shell.Extensions.GSConnect.gschema.xml#L35-L37

This is disabled by default and only enabled temporarily by the `Generate Support Log` feature:

https://github.com/GSConnect/gnome-shell-extension-gsconnect/blob/a10961d895840984f7be05e6211434044278b8d4/src/preferences/service.js#L420-L432

Unless you've turned this on manually (or found a bug), this should never print any private information to the log. In the past, there was a switch in the UI, but it was removed as users would sometimes turn it on and forget about it.

Unless there's a good reason for someone to turn this *on* manually, I'm not sure it's a good idea to offer a switch in the UI again. It's possible to just turn it *off* everytime the service starts, but that would prevent debugging from startup.

neutrino38:
ok then let's drop this. I only started to play with GSconnect to tinker it and got debug activated. So I agree that this is not relevant.

PR #2095, opened 2025-11-05
Fix sensitivity state for "Share When Active" button

### LINKED ISSUE #1599: Notifications > "Share When Active" always initializes enabled
### Describe the bug

So, this is the _dumbest_ little display bug, a silly little thing that is the epitome of a "good first issue" if anyone is looking to dive in and make a fix.

Whenever a device's Notifications preferences view is first opened, the "Share When Active" switch always initializes as sensitive even if "Share Notifications" is _disabled_ (which is supposed to make "Share When Active" insensitive, the same way it does all the applications in the list:

### First arrival at "Notifications" after opening "Mobile Settings"

image

It's purely a setup bug — if you cycle "Share Notifications" on and then back off again, it'll become insensitive, and it'll stay that way as long as the Settings window is still open. (The bug reoccurs each time the window is newly constructed, though.)

### After cycling "Share Notifications"
image

### Steps to reproduce

1. Open "Mobile Settings"
2. Click a device (connected or not)
3. Switch to "Notifications"
4. If "Share Notifications" is on, turn it off, close the window, and start back from step 1 again.

### Expected behavior

"Share When Active" never appears sensitive (able to be interacted with) unless "Share Notifications" is switched on.

### GSConnect version

54 & 55

### Installed from

GNOME Extensions website

### GNOME Shell version

43.3 & 44

Fedora 37 – 38b

### Paired device(s)

_No response_

### KDE Connect app version

_No response_

### Plugin(s)

_No response_

### Support log

_No response_

### Screenshots

_No response_

### Notes

_No response_

Vict0rTesla:
This PR fixes the sensitivity state of the `Share When Active` button.

Previously. the button's sensitivity was controlled using the UI bind method.
which caused the initial state to be incorrect when opening the window.
I updated the implementation to use the same sensitivity handling pattern as `notifications-apps`, ensuring the correct state is applied consistently.

Screenshot From 2025-11-05 19-16-40
Screenshot From 2025-11-05 19-17-36

Fixes #1599

### DIFF
--- a/src/preferences/device.js
+++ b/src/preferences/device.js
@@ -281,7 +281,7 @@ export const Panel = GObject.registerClass({
 
         // Notifications
         'notification', 'notification-page',
-        'notification-list', 'notification-apps',
+        'notification-list', 'notification-apps', "share-notification-active",
 
         // Telephony
         'telephony', 'telephony-page',
@@ -787,6 +787,13 @@ export const Panel = GObject.registerClass({
         settings.bind(
             'send-notifications',
             this.notification_apps,
+            'sensitive',
+            Gio.SettingsBindFlags.DEFAULT
+        );
+        
+        settings.bind(
+            'send-notifications',
+            this.share_notification_active,
             'sensitive',
             Gio.SettingsBindFlags.DEFAULT
         );


Vict0rTesla:
@ferdnyc this PR aims to fix #1599. Could you please review it?

ferdnyc:
Not until next week at the earliest, I'm afraid.

Vict0rTesla:
@ferdnyc  Hi, just following up on the PR review when you get a moment. No rush — just making sure it’s still on your radar.

PR #2110, opened 2025-12-16
Added name option for CLI

snaCW:
This PR lets you write commands using only the device name.
```bash
./daemon.js --name "MyDeviceName" --ring
./daemon.js -n "MyDeviceName" --ring
```
This adds readability and ease of use to the CLI.

### DIFF
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -773,7 +773,6 @@ const Service = GObject.registerClass({
                 const name = options.lookup_value('name', null).unpack();
                 id = this._findDeviceID(name); // May return null if no match found
             }
-            
             if (id === null)
                 return -1;
             // Pairing


src/service/daemon.js
```diff
@@ -731,10 +767,19 @@ const Service = GObject.registerClass({
 
             // We need a device for anything else; exit since this is probably
             // the daemon being started.
-            if (!options.contains('device'))
+            let id = null;
+            if (options.contains('device')) {
+                id = options.lookup_value('device', null).unpack();
+            }
+            else if (options.contains('name')) {
```
andyholmes:
```suggestion
            } else if (options.contains('name')) {
```

snaCW:
I removed `Object.freeze(id)` as suggested and fixed the else-if formatting.

src/service/daemon.js
```diff
@@ -710,6 +719,33 @@ const Service = GObject.registerClass({
...
+        let device;
+
+        for (let object of Object.values(variant)) {
+            object = object.recursiveUnpack();
+            device = object['org.gnome.Shell.Extensions.GSConnect.Device'];
+
+            if (name === device.Name)
+                return device.Id;
+        }
+
```
andyholmes:
```suggestion

```

andyholmes (changes requested):
Sorry, I should run the test suite first 🤦 

Just some linter fixes, then it's good to merge.

andyholmes (approved):

andyholmes:
Thanks!

cc @ferdnyc The last test error looks like `dbus-run-session` got dropped from the CI somehow:

```
/usr/sbin/xvfb-run: line 181: dbus-run-session: command not found
Error: Process completed with exit code 127.
```

Did this maybe happen when the base image was bumped?

ferdnyc:
@andyholmes Most likely. `dbus-run-session` is part of the `dbus-daemon` package, which I doubt we're installing explicitly. It was probably part of the default package set for Fedora <= 42, but removed from F43 because we're all in on `dbus-broker` now. Should just be a matter of adding it to the Dockerfile as an explicit install.

ferdnyc:
GSConnect/gsconnect-ci#36 should sort it.

ferdnyc:
Hm. Maybe we should add a CI job to the `gsconnect-ci` repo. It could check out the GSConnect `main` sources and attempt to build & test them using the PR branch's image.

ferdnyc:
[Re-ran the failed job](https://github.com/GSConnect/gnome-shell-extension-gsconnect/actions/runs/20364249574/job/58580856698), and it's all good now.

ferdnyc:
Hm. I wonder if we should be using [`dbus-test-runner`](https://packages.fedoraproject.org/pkgs/dbus-test-runner/dbus-test-runner/) instead of `dbus-run-session`, anyway.

andyholmes:
> Hm. I wonder if we should be using [`dbus-test-runner`](https://packages.fedoraproject.org/pkgs/dbus-test-runner/dbus-test-runner/) instead of `dbus-run-session`, anyway.

Never heard of that before, actually. Looks like it still requires `dbus-daemon` (and/or `dbus-run-session`) either way, but might be interesting to look at one day.

ferdnyc:
@andyholmes Having looked at it some more, for us I don't think it really makes a difference
 Its biggest feature is the ability to run multiple tasks in the same session. (And to start tasks after some interface has appeared on the bus, that kind of thing.) But we only have the one task.

PR #2115, opened 2026-01-07
Adds inital support for mDNS discovery via the dbus interface of avahi

### LINKED ISSUE #1654: Add support for mDNS discovery
### Describe your request

KDE Connect recently added preliminary support for mDNS discovery with DNS-SD, which may fix several problems with UDP broadcasting.

### Proposed solution

* [`kdeconnect-android`: Add MDNS discovery](https://invent.kde.org/network/kdeconnect-android/merge_requests/375)
* [`kdeconnect-kde`: Add MDNS discovery](https://invent.kde.org/network/kdeconnect-kde/merge_requests/555)
* [`kdeconnect-ios`: Add MDNS discovery](https://invent.kde.org/network/kdeconnect-ios/merge_requests/108)

### Alternatives

_No response_

### GSConnect version

All

### Installed from

other source

### GNOME Shell version

44+

All

### Additional context

_No response_

lub-dub:
Inital mdns support so it works on enterprise WiFi deployments that only allow mdns based discovery and block broadcast.

I'm not too familiar with the gio/gjs dbus behavoir and if in the current setup this will gracefully not work on systems that don't have dbus running or available. Otherwise some extra fencing needs to be added if there is a easier way to detect if dbus is available on the system.

Feedback/ pointers are welcome. I personally have not a lot of JavaScript experience so if there are any obvious things that should be changed/improved feel free to let me know.

This closes #1654

### DIFF
--- a/src/service/backends/avahi/avahi-dbus.js
+++ b/src/service/backends/avahi/avahi-dbus.js
@@ -1,0 +1,345 @@
+// SPDX-FileCopyrightText: GSConnect Developers https://github.com/GSConnect
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+import Gio from 'gi://Gio';
+
+const xmlDef = `
+<node>
+  <interface name="org.freedesktop.Avahi.Server2">
+
+    <method name="GetVersionString">
+      <arg name="version" type="s" direction="out"/>
+    </method>
+
+    <method name="GetAPIVersion">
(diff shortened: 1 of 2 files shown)


lub-dub:
Fixed the failed pipeline, and forced pushed over, still had a wrong header left over wrong modifying some example code. Should now pass with no issue.

PR #2128, opened 2026-02-04
feat: Add Virtual Camera plugin

adymany:
Adds a new plugin to receive camera streams from Android.

Uses v4l2loopback and GStreamer to create a virtual webcam device (/dev/videoX).
- Auto-loads v4l2loopback on stream start
- Unloads module on stream stop
- Supports MJPEG streaming from Android
- this works with my fork of the kde connect app with virtual camera plugin [link to my kde-connect fork](https://invent.kde.org/adymany/kdeconnect-android)

### DIFF
--- a/src/service/plugins/index.js
+++ b/src/service/plugins/index.js
@@ -18,6 +18,7 @@ import * as share from './share.js';
 import * as sms from './sms.js';
 import * as systemvolume from './systemvolume.js';
 import * as telephony from './telephony.js';
+import * as virtualcamera from './virtualcamera.js';
 
 export default {
     battery,
@@ -36,4 +37,5 @@ export default {
     sms,
     systemvolume,
     telephony,
+    virtualcamera,
 };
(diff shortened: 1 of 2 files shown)


andyholmes:
Aside from the linter errors, this seems pretty awesome to me :slightly_smiling_face:

Given that the lint errors are fixed, and it's merged upstream in kdeconnect-android, I'm happy to give my :+1: to this.

daniellandau:
This seems to be stalled on the KDE side with another proposed camera streaming MR being favored, but also stalled. We can keep this open for now until the other side becomes clearer

adnantabrezi:
> This seems to be stalled on the KDE side with another proposed camera streaming MR being favored, but also stalled. We can keep this open for now until the other side becomes clearer

this method is still not practical as it depends on v4l2loopback which is not officially signed driver , it can use pipewire camera input just like obs virtual camera

PR #2134, opened 2026-02-11
Add Device.validateLocalId for migration

### LINKED ISSUE #2133: GSConnect is not working with KDE Connect 1.35
### Describe the bug

Since the last update on the android App to 1.35, device is not detected anymore.
I had to downgrade to 1.34.4 and now is back.

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### GSConnect version

71

### Installed from

GNOME Extensions website

### GNOME Shell version

_No response_

Ubuntu 22.04

### Paired device(s)

Pixel 9 Pro

### KDE Connect app version

1.35.2

### Plugin(s)

_No response_

### Support log

```plain text

```

### Screenshots

_No response_

### Notes

_No response_

ferdnyc:
The `Device.validateId()` function was made more liberal just prior to the release of v61 (or was it v66?), breaking its use in `_migrateConfiguration()` to trigger resetting of invalid device IDs. (It accepts hyphenated UUIDs, which we no longer want to consider valid local device IDs.)

KDE Connect 1.35 has stopped accepting hyphenated IDs, so we really need to force migration of those. To facilitate that, add a new `Device.validateLocalId()` that accepts only UUID strings without separators, or with underscore separators (no hyphens), and use that in `_migrateConfiguration()`.

Fixes #2133

### DIFF
--- a/src/service/daemon.js
+++ b/src/service/daemon.js
@@ -66,7 +66,7 @@ const Service = GObject.registerClass({
 
         const certificate = Gio.TlsCertificate.new_for_paths(certPath, keyPath, null);
 
-        if (Device.validateId(certificate.common_name))
+        if (Device.validateLocalId(certificate.common_name))
             return;
 
         // Remove the old certificate, serving as the single source of truth
--- a/src/service/device.js
+++ b/src/service/device.js
@@ -139,6 +139,10 @@ const Device = GObject.registerClass({
         return /^[a-zA-Z0-9_-]{32,38}$/.test(id);
     }
 
+    static validateLocalId(id) {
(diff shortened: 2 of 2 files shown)


andyholmes (approved):
LGTM, sorry for the long delay.

daniellandau (changes requested):
The IDs with dashes are explicitly allowed by the protocol and the recent breakage reports seem to have been about IP address restrictions so I don't think doing this makes any sense.

andyholmes:
I might have misunderstood, per my approval, but I think this is intended to only apply to device IDs generated by GSConnect.

PR #2156, opened 2026-04-09
Follow XDG Base Directory Specification for `.mozilla` (fixes #2121)

### LINKED ISSUE #2121: Gsconnect preventing /.mozilla folder from  being removed
### Describe the bug

the only file keeping /.mozilla directory in home is this org.gnome.shell.extensions.gsconnect.json file from gsconnect located in /native-messaging-hosts folder after firefox starting to support xdg-desktop protocol , so it should be changed .

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### GSConnect version

71

### Installed from

Extension Manager

### GNOME Shell version

49.2

Cachyos

### Paired device(s)

Redmi note 10 pro

### KDE Connect app version

_No response_

### Plugin(s)

_No response_

### Support log

```plain text

```

### Screenshots

_No response_

### Notes

_No response_

Renarde-dev:
Uses $HOME/.mozilla if it already exist, if not use configDir/mozilla
This will allow user who still have their config in `.mozilla` to keep it.

I checked syntax with `eslint` and tested this on gnome 50 (ArchLinux) but I do not have a lot of experience with GJS and gnome extension dev in general.


daniellandau (approved):
LGTM, thanks!

ferdnyc:
Hmm. I wonder if this fix should look a bit deeper, instead.

Any of our users will have a `$HOME/.mozilla/native-messaging-hosts/` directory, because _we_ have been keeping it in place. But as the reporter in #2121 said, we may be the only thing preventing them from deleting `$HOME/.mozilla`.

Maybe it would be better to look for a <code>$HOME/.mozilla/{firefox,icecat,<var>others?</var>}</code> dir, vs. a <code>$HOME/.config/mozilla/{firefox,icecat,<var>others?</var>}</code> dir, when deciding where to put the NMH? Just so that _our own_ installs don't throw off the logic.

PR #1814, opened 2024-05-27
Add linkable shell script launchers

ferdnyc:
This PR adds or replaces the runnable JS modules with shell scripts that can be executed from any directory, even symlinked.

So, a user interested in operating GSConnect in CLI mode can symlink `gsconnect.sh` and `gsconnect-preferences.sh` into `/usr/local/bin/` or `$HOME/.local/bin/` (as either their original names, or as `gsconnect` and `gsconnect-preferences`), and use them from the command line.

This is functionality that was formerly available, but broke in the migration to ES modules.

Specific changes:

* Rename `gsconnect-preferences` to `preferences-app.js` and remove its shebang line.
* Add a `gsconnect-preferences.sh` shell script that will launch `preferences-app.js` in `gjs` correctly, even if symlinked.
* Add a new `gsconnect.sh` that will do the same for `service/daemon.js`.
* Use `gsconnect-preferences.sh` to launch `preferences-app.js` from `prefs.js`.
* Make tools launchable without `runAsync`, for environments without latest Gtk.

Issue #1787, opened 2024-04-21
Filepaths in Wiki: CLI usage without Gnome environment


nebulosa2007:
### Describe your request

After this commit addda66 (I've done [git bisect](https://github.com/GSConnect/gnome-shell-extension-gsconnect/files/15051692/gitbisect.log)) links, as described in Wiki, stopped working:

```console
$ ls -l /usr/bin/gsconnect*
lrwxrwxrwx 1 root root 82 Apr 21 13:38 /usr/bin/gsconnect -> /usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/daemon.js
lrwxrwxrwx 1 root root 86 Apr 21 13:38 /usr/bin/gsconnect-preferences -> /usr/share/gnome-shell/extensions/gsconnect@andyholmes.github.io/gsconnect-preferences

$ /usr/bin/gsconnect

(gjs:45130): Gjs-WARNING **: 14:27:59.041: JS ERROR: ImportError: Unable to load file from: file:///usr/bin/init.js (Error opening file /usr/bin/init.js: No such file or directory)


(gjs:45130): Gjs-CRITICAL **: 14:27:59.041: Failed to resolve imports for module: 'file:///usr/bin/gsconnect'

$ gsconnect

(gjs:45944): Gjs-WARNING **: 14:37:07.248: JS ERROR: ImportError: Unable to load file from: file:///sbin/init.js (Error opening file /sbin/init.js: No such file or directory)


(gjs:45944): Gjs-CRITICAL **: 14:37:07.248: Failed to resolve imports for module: 'file:///sbin/gsconnect'

$ /usr/bin/gsconnect-preferences 

(gjs:45173): Gjs-WARNING **: 14:28:01.902: JS ERROR: ImportError: Unable to load file from: file:///usr/bin/preferences/init.js (Error opening file /usr/bin/preferences/init.js: No such file or directory)


(gjs:45173): Gjs-CRITICAL **: 14:28:01.902: Failed to resolve imports for module: 'file:///usr/bin/gsconnect-preferences'

$ gsconnect-preferences 

(gjs:46019): Gjs-WARNING **: 14:37:47.066: JS ERROR: ImportError: Unable to load file from: file:///sbin/preferences/init.js (Error opening file /sbin/preferences/init.js: No such file or directory)


[...]

PR #2157, opened 2026-04-13
Improve CLI support for non-GNOME environment

JingMatrix:
As explained in PR #1814, we need a wrapper to fix the path issue in #1787. Since PR #1814 becomes inactive, we re-create a wrapper script that is consistent with `gsconnect-preference`.

Moreover, the custom debug function is not available for early CLI initialization since a caller couldn't be found from error stack.


JingMatrix:
The mentioned PR and issue could be closed once this PR is merged.

P.S.: I was the author of the CLI support for non-GNOME environment and tutorial https://github.com/GSConnect/gnome-shell-extension-gsconnect/wiki/CLI-usage-without-Gnome-environment.

daniellandau (approved):
LGTM

PR #2150, opened 2026-03-14
Fix handle-method-call signature for GNOME 49 compatibility

Austrum-lab:
The callback signature for GjsPrivate.DBusImplementation::handle-method-call changed in GNOME 49 from (iface, name, parameters, invocation) to (iface, name, invocation, parameters).
In the current version clipboard not working for gnome 49.

Add runtime detection to support both GNOME 48 and earlier, and GNOME 49+.

Fixes: #2048 (GNOME 49 support)
Affected components:
- shell/clipboard.js
- wl_clipboard.js
- service/utils/dbus.js
- service/components/notification.js

PR #2165, opened 2026-06-08
shell/clipboard.js: wrap async handler in safety-net try/catch (no unhandled rejections)

vipinus:
handle-method-call is a sync signal from GObject's perspective. The async handler's returned Promise isn't awaited, so any throw past the inner try/catches becomes an unhandled rejection.

The escape hatch in practice is `invocation.return_*` itself throwing when the peer connection is gone or the invocation has already been answered.

On my machine (gnome-shell 50.1, Wayland, no paired devices):

- `journalctl --user-unit gnome-shell --since '1 hour ago' | grep -c '_onHandleMethodCall@.*clipboard.js'` → 10
- libsoup runtime check failures from the daemon side at the same rate: `soup_message_queue_item_destroy: soup_message_get_connection (item->msg) == NULL`

Patch:
- top-level try/catch around the whole body, `logError` on the way out
- wrap each `invocation.return_*` call in its own try/catch (peer-gone safe)
- guard `e.name` in case `e` is not an `Error`

No behavior change on the success path. After patching + restarting the daemon, the unhandled-rejection lines and the libsoup warnings both drop to ~0/hour.

Keeps the param1/param2 detection from #2150.


daniellandau:
Thank you, this was merged in https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/2177

PR #2171, opened 2026-07-11
Update Polish translation.

gszy:
Add missing, improve existing strings. This patch series rebuilds the PO Template file.


ferdnyc (approved):
Thanks @gszy !

PR #2181, opened 2026-07-31
Add ESlint configuration to limit line-length, and correct all newly-flagged issues

ferdnyc:
This PR adds configuration for the `@stylistic/max-len` ESLint plugin, which sets limits on the maximum length of source lines in the codebase. Specifically, the configuration applied is:

```js
{
            '@stylistic/max-len': ['error', {
                code: 100,
                comments: 80,
                ignoreComments: false,
                ignorePattern: 'TRANSLATORS:',
                ignoreTrailingComments: false,
                ignoreRegExpLiterals: true,
                ignoreStrings: true,
                ignoreTemplateLiterals: true,
                ignoreUrls: true,
            }],
```

Which limits source lines to 100 characters, comment lines to 80 characters, and makes all of the exceptions listed there, so that we won't be forced to split long URLs, or strings, or TRANSLATOR: comments, etc.

The second commit corrects all of the newly-flagged errors resulting from running `eslint` with that added configuration.

### Motivation

I noticed some recent changes added excessively-long comment lines, when there's no reason not to wrap comments (in particular) to fit readably in an 80-character width.

But I didn't want to unduly limit our ability to write long string literals, or regular expressions, or URLs, etc. And I didn't want to constrain, or force any changes to, our `TRANSLATOR:` comments. So I made exceptions for all of those.

The worst of the changes this forced were in `src/shell/notification.js`, which has several interactions with the unwieldy-named `NotificationDaemon.GtkNotificationDaemonAppSource` object. So, for example, the diff contains:

```diff
-    NotificationDaemon.GtkNotificationDaemonAppSource.prototype._withdrawGSConnectNotification = _withdrawGSConnectNotification;
+    NotificationDaemon.GtkNotificationDaemonAppSource.prototype
+    ._withdrawGSConnectNotification = _withdrawGSConnectNotification;
```

and

```diff
-    NotificationDaemon.GtkNotificationDaemonAppSource.prototype.addNotification = _addNotification;
-    delete NotificationDaemon.GtkNotificationDaemonAppSource.prototype._withdrawGSConnectNotification;
+    NotificationDaemon.GtkNotificationDaemonAppSource.prototype
+    .addNotification = _addNotification;
+
+    delete NotificationDaemon.GtkNotificationDaemonAppSource
+           .prototype._withdrawGSConnectNotification;
```

I don't _love_ that, but the original source lines really WERE excessively long. And now they're readable in the GitHub web interface, which was part of the intent with this.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -113,6 +113,18 @@ export default defineConfig([
                 max: 5,
             }],
 
+            '@stylistic/max-len': ['error', {
+                code: 100,
+                comments: 80,
+                ignoreComments: false,
+                ignorePattern: 'TRANSLATORS:',
+                ignoreTrailingComments: false,
+                ignoreRegExpLiterals: true,
+                ignoreStrings: true,
+                ignoreTemplateLiterals: true,
+                ignoreUrls: true,
+            }],
+
             '@stylistic/max-statements-per-line': 'error',
             '@stylistic/new-parens': 'error',
             'no-array-constructor': 'error',
--- a/installed-tests/fixtures/utils.js
+++ b/installed-tests/fixtures/utils.js
@@ -259,7 +259,9 @@ function isolateDirectories() {
  * Patch in the mock components for plugin tests.
  */
 export async function mockComponents() {
-    const {functionOverrides} = await import(`file://${Config.PACKAGE_DATADIR}/service/components/index.js`);
+    const {functionOverrides} = await import(
+        `file://${Config.PACKAGE_DATADIR}/service/components/index.js`
+    );
     const MockComponents = await import('./components/index.js');
 
     functionOverrides.acquire = function (name) {
--- a/installed-tests/suites/components/testClipboardComponent.js
+++ b/installed-tests/suites/components/testClipboardComponent.js
@@ -9,7 +9,9 @@ import Gtk from 'gi://Gtk';
 import GLib from 'gi://GLib';
 
 import Config from '../config.js';
-const {default: Clipboard} = await import(`file://${Config.PACKAGE_DATADIR}/service/components/clipboard.js`);
+const {default: Clipboard} = await import(
+    `file://${Config.PACKAGE_DATADIR}/service/components/clipboard.js`
+);
 
 
 describe('The Clipboard component', function () {
@@ -55,4 +57,3 @@ describe('The Clipboard component', function () {
         clipboard.text = text;
     });
 });
-
--- a/installed-tests/suites/components/testMprisComponent.js
+++ b/installed-tests/suites/components/testMprisComponent.js
@@ -8,7 +8,9 @@ import '../fixtures/utils.js';
 import MockPlayer from '../fixtures/mpris.js';
 
 import Config from '../config.js';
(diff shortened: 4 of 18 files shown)


ferdnyc:
BTW, if the consensus on things like this:

```js
NotificationDaemon.GtkNotificationDaemonAppSource.prototype
._withdrawGSConnectNotification = _withdrawGSConnectNotification;
```
Is that we prefer it indented, i.e.
```js
NotificationDaemon.GtkNotificationDaemonAppSource.prototype
    ._withdrawGSConnectNotification = _withdrawGSConnectNotification;
```

...then I can do that, and see about adding ESLint rules to formalize that as a style requirement. Currently our configs don't include any expectations about that sort of formatting, and personally I find the indented form slightly weirder-looking than the non-indented form.

But, we could also go with something like...
```js
NotificationDaemon
.GtkNotificationDaemonAppSource
.prototype
._withdrawGSConnectNotification = _withdrawGSConnectNotification;
```
(OK, I already hate that), or

```js
NotificationDaemon.GtkNotificationDaemonAppSource
                  .prototype
                  ._withdrawGSConnectNotification = _withdrawGSConnectNotification;
```

...Although I think that still exceeds the 100-char limit. (But that could be raised to 120 chars, or something.)

daniellandau (approved):
Assuming the diff is just `eslint --fix` after the config change this is fine for me.

ferdnyc:
> Assuming the diff is just `eslint --fix` after the config change this is fine for me.

Actually it was all manual, as `eslint --fix` doesn't fix line-length issues surprisingly!

daniellandau:
For the indentation I usually prefer whatever my editor or a prettier-js or whatever wants to do over a manually written indentation policy. From your examples the second is IMO better than first

ferdnyc:
The second being this?

```js
NotificationDaemon.GtkNotificationDaemonAppSource.prototype
    ._withdrawGSConnectNotification = _withdrawGSConnectNotification;
```

daniellandau:
> > Assuming the diff is just `eslint --fix` after the config change this is fine for me.
> 
> Actually it was all manual, as `eslint --fix` doesn't fix line-length issues surprisingly!

Oh. It's always dangerous to fix linter issues manually in bulk. I've introduced many bugs like that in various code bases. Adding line feeds seems like one of the safer things to do so I'm leaving my approval in place but know that I didn't read the full diff.

daniellandau:
> The second being this?
> 
> ```js
> NotificationDaemon.GtkNotificationDaemonAppSource.prototype
>     ._withdrawGSConnectNotification = _withdrawGSConnectNotification;
> ```

Yes, sorry. Writing from phone so didn't quote properly. In any case stylistically I don't have a strong preference. It's the tooling part that is more important for me and something like that is what my editor would produce there.

ferdnyc:
Ohh-kay, I went back through and fired off "Format selection" in VSCode for all of the code changes, and committed anything where it altered my manual reformatting. So at least there's some standardization to the changes, tho they're still heavily influenced by the JS/TS extension's Formatting configuration. 

(Which I did adjust, since by default it was configured to add spaces inside curly braces, and I disliked this rather a lot:)
```js
    const { functionOverrides } = await import(
        `file://${Config.PACKAGE_DATADIR}/service/components/index.js`
    );
```

PR #2184, opened 2026-08-05
Fix two issues with the phone number conversion methods, and add tests

ferdnyc:
This PR contains a set of tests, and two fixes, for the phone-number conversion methods we monkeypatch into `String.prototype`.

The first commit adds a new Jasmine suite, `unit`, containing `testPhoneNumber.js`, which tests those methods in isolation.

The second commit fixes both issues that would otherwise cause those tests to fail:

1. `String.prototype.toPhoneNumber()` was using a bad regex to clean up phone numbers, such that a number like `+1 (212) 555-1212` would come out as `+12125551212`, and `+01-212-555-1212` would come out as `+012125551212`. Fixed by replacing with a chained double `.replace()` that holds off stripping leading zeroes until AFTER all of the other unwanted characters are cleaned up.

2. `String.prototype.equalsPhoneNumber()` was returning the result of some comparisons directly, in a way that caused the function to return `0` for some failure cases. Wrapped the return value in a `Boolean()` to ensure that it's always exactly `true` or `false`, as specified in the jsdoc metadata.

### DIFF
--- a/installed-tests/suites/meson.build
+++ b/installed-tests/suites/meson.build
@@ -6,4 +6,4 @@ subdir('core')
 subdir('backends')
 subdir('components')
 subdir('plugins')
-
+subdir('unit')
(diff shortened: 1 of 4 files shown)


daniellandau (approved):
LGTM

Issue #2185, opened 2026-08-06
IPv6 addresses are mangled by the LAN backend, breaking direct reconnect


ihubanov:
Hey folks,

I hit a couple of IPv6 address-handling bugs in the LAN backend while getting GSConnect
working over a Bluetooth PAN link. They look like real bugs rather than something about my
setup, so I thought I'd write them up.

Short version: GSConnect happily *accepts* an IPv6 connection, but it can't parse an IPv6
address back out of its own strings. So it never reconnects directly to an IPv6 peer — it
silently falls back to a 255.255.255.255 broadcast instead, with no error anywhere.

Two spots, both in `service/backends/lan.js` (v72):

**1. `broadcast()`, line 427:**

```js
const [host, portstr] = address.split(':');
```

That splits on the first colon, so `fe80::de6a:e7ff:fe1e:a86a:1716` gives you host `fe80`.
`Gio.InetSocketAddress.new_from_string('fe80', port)` then returns null.

**2. `Channel.address` getter, line 555:**

```js
get address() {
    return `lan://${this.host}:${this.port}`;
}
```

No brackets around the host, so any IPv6 address comes out ambiguous and unparseable.

The two feed each other, because that value round-trips through GSettings:

- `service/device.js:279` — `this.settings.set_string('last-connection', this.channel.address)`
- `service/manager.js:465` — `this.identify(device.settings.get_string('last-connection'))`

So for an IPv6 peer the sequence is: the getter writes `lan://fe80::de6a:e7ff:fe1e:a86a:1716`
into `last-connection`, `identify()` splits off the scheme, `broadcast()` splits on `:` and
gets host `fe80`, `new_from_string()` returns null, the `address instanceof
Gio.InetSocketAddress` check fails, and it drops into the else branch and broadcasts to the
whole LAN. Quiet fallback, no warning. With debug on you just get:

```
[service/backends/lan.js:broadcast]: could not parse address "fe80"
```

To be clear about what does work: the inbound path is fine. I sent a `kdeconnect.identity`
packet to an Android peer over IPv6 link-local and it dialled back and completed a full TLS
handshake against GSConnect's listener, unpatched. Confirmed with tcpdump. It's only the
outbound/reconnect direction that's broken.

Environment: GSConnect v72, GNOME Shell 50.1, gjs 1.88.0, Ubuntu 26.04, Wayland.

Happy to send a PR if you'd like one. Rough plan would be bracket-aware parsing of
[...]

PR #2186, opened 2026-08-06
lan: fix IPv6 address parsing and formatting

ihubanov:
Hey,

This fixes the two IPv6 address-handling bugs I wrote up in #2185.

Quick recap of what was wrong:

- `broadcast()` parsed `<host>:<port>` with `address.split(':')`, which splits on the first
  colon and so shreds any IPv6 literal — `fe80::1:1716` comes out as host `fe80`.
- `Channel.address` built `lan://${this.host}:${this.port}` with no brackets, so IPv6
  addresses were emitted in an unparseable form.

Those compound, because the value round-trips through GSettings (`device.js:279` writes
`channel.address` into `last-connection`, `manager.js:465` feeds it back to `identify()`).
The net effect is that GSConnect never reconnects directly to an IPv6 peer — the parse
returns null, the `instanceof` check fails, and it quietly falls back to a 255.255.255.255
broadcast with no warning.

What's in here:

- `_formatAddress()` brackets IPv6 hosts.
- `_parseSocketAddress()` handles `[addr]:port`, `[addr%zone]:port`, the plain IPv4 form, and
  bare IPv6 literals — the last one so that values already written unbracketed into
  `last-connection` by earlier versions still parse after upgrading.
- A zone is resolved to a `scope_id` by reading `/sys/class/net/<iface>/ifindex`, since GJS
  exposes no `if_nametoindex()` and `Gio.InetSocketAddress.new_from_string()` has no
  parameter for it. Returns 0 where there's no sysfs, which is correct for everything except
  link-local addresses.
- `broadcast()` now sends on a socket that can carry the address family. Worth a look during
  review: when `_udp6.speaks_ipv4()` is true, `_initUdpListener()` deliberately leaves
  `_udp4` as null, so IPv4 targets have to fall through to `_udp6`. The naive version of this
  change silently breaks all IPv4 broadcast.

No protocol changes — this only touches how addresses are stringified and parsed locally, so
it shouldn't have any KDE Connect compatibility implications.

Added 12 tests to `testLanBackend.js` covering formatting, parsing, zones, round-tripping and
the unparseable case. All pass.

Verified end to end as well, not just in unit tests. With the patch running I triggered a
directed identify at an Android peer's link-local address over a Bluetooth PAN link:

- the identity packet goes out over IPv6 (`fe80::…6d41b > fe80::…a86a … UDP, length 1778`,
  fragmented, which is worth knowing if you go looking for it — a `port 1716` capture filter
  won't match the fragments)
- the peer connects back and the channel is established
- `Channel.address` now yields `lan://[fe80::de6a:e7ff:fe1e:a86a]:1716`
- that gets written to `last-connection` in a form that parses back correctly, which is the
  round-trip that was broken

Heads up for anyone testing an upgrade path: once the fixed getter has written a bracketed
address into `last-connection`, downgrading to an unpatched version leaves a value it can't
parse. The reverse direction is handled — the bare-IPv6 branch in `_parseSocketAddress()`
exists so old unbracketed values still work after upgrading.

One thing to flag honestly: the 7 pre-existing `A LAN channel service` tests fail in my
environment, but they fail identically on a clean checkout with none of my changes applied,
so I don't believe they're related. Would appreciate a second pair of eyes on that.

Tested on GSConnect v72, GNOME Shell 50.1, gjs 1.88.0, Ubuntu 26.04, Wayland. ESLint clean.

### DIFF
--- a/installed-tests/suites/backends/testLanBackend.js
+++ b/installed-tests/suites/backends/testLanBackend.js
@@ -145,3 +145,93 @@ describe('A LAN channel service', function () {
     // TODO: restarting stopped services
 });
 
+
+describe('LAN address formatting', function () {
+    it('leaves IPv4 addresses unbracketed', function () {
+        expect(Lan._formatAddress('192.168.1.10', 1716))
+            .toBe('lan://192.168.1.10:1716');
+    });
+
+    it('brackets IPv6 addresses', function () {
+        expect(Lan._formatAddress('2001:db8::1', 1716))
+            .toBe('lan://[2001:db8::1]:1716');
+    });
+
+    it('brackets IPv6 addresses carrying a zone', function () {
+        expect(Lan._formatAddress('fe80::1%wlan0', 1716))
+            .toBe('lan://[fe80::1%wlan0]:1716');
+    });
+});
+
+
+describe('LAN address parsing', function () {
+    it('parses an IPv4 address with a port', function () {
+        const addr = Lan._parseSocketAddress('192.168.1.10:1716', 1716);
+
+        expect(addr).not.toBeNull();
+        expect(addr.address.to_string()).toBe('192.168.1.10');
+        expect(addr.port).toBe(1716);
+    });
+
+    it('falls back to the default port', function () {
+        const addr = Lan._parseSocketAddress('192.168.1.10', 1716);
+
+        expect(addr).not.toBeNull();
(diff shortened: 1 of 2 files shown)


daniellandau (approved):
Looks good to me, thanks! The test suite is notoriously difficult to make pass locally.

PR #2187, opened 2026-08-09
Separate manifest v3 extension with reduced functionality for chrome

daniellandau:
Supersedes https://github.com/GSConnect/gnome-shell-extension-gsconnect/pull/1977

As per the discussion in the previous pull request, the popup menu didn't work so I took it out from this. As it still does work in Firefox I decided to separate these into two folders with reduced and full functionality.

image

Current situation is that Chrome has no webextension at all, so this is an improvement on that.


daniellandau:
If no major objections, I'll go with this approach soon

PR #2188, opened 2026-08-09
Fix 'shortucts' typo error to 'shortcuts'

mlorier:
Hi!
I saw in the last commit that a typo error has been introduced: so here is the fix!


daniellandau (approved):
Whoops, good catch!

PR #2189, opened 2026-08-10
Better conditional GioUnix loading

ferdnyc:
Sorry for the churn, found this after-the-fact.

Taken from some GNOME internal code, but actually also found in our own `installed-tests/fixtures/backend.js` (I subsequently realized). Hopefully those facts will render it maximally future-proof.


daniellandau (approved):
Works on Ubuntu 26.04, LGTM

PR #2190, opened 2026-08-14
Phone number comparisons: Use .equalsPhoneNumber

### LINKED ISSUE #2176: Contact with all-alphabetic phone number ("sender ID") can hijack lookups for any real number
## Summary

`Contacts.query()` matches a real phone number to the wrong contact when another
contact in the address book has a `TEL` value containing **no digits at all**
(e.g. an alphanumeric SMS sender ID like a company name, imported from a vCard).

## Root cause

In `query(query)`:

```js
const cnumber = num.value.toPhoneNumber();
...
if (qnumber.endsWith(cnumber) || cnumber.endsWith(qnumber)) {
```

`toPhoneNumber()` strips all non-digit characters. A `TEL` value with **zero digits**
(e.g. a vCard entry like `TEL;TYPE=WORK,PAGER:Amazon`, used to give a nice
name/avatar to SMS received from an alphanumeric sender ID) normalizes to an
**empty string**.

In JavaScript, every string `.endsWith('')` is `true`. So once `cnumber === ''`,
the condition `qnumber.endsWith(cnumber)` is true for **any** `qnumber`,
regardless of its actual value. If such a contact is reached before the real
match while iterating `this.contacts`, `query()` returns it immediately (the
`!query.name` early-return path), silently misattributing an SMS/call from a
real phone number to the wrong contact.

## Fix

- File: `service/components/contacts.js` — `query()` method
  (GSConnect repo: same code, `src/service/components/contacts.js`)

Only apply the `endsWith` shortcut when both numbers actually contain digits;
keep matching two empty-normalized numbers to each other (harmless, and
preserves the "cosmetic" alphanumeric-sender-ID use case):

```diff
-                if (qnumber.endsWith(cnumber) || cnumber.endsWith(qnumber)) {
+                if ((qnumber !== '' && cnumber !== '' && (qnumber.endsWith(cnumber) || cnumber.endsWith(qnumber))) ||
+                    (qnumber === '' && cnumber === '')) {
```

Tested locally (Zorin Connect fork, same code): stops a real phone number from
ever matching an empty-normalized contact, while leaving normal number
matching and the alphanumeric-sender-ID cosmetic case unchanged.

### Steps to reproduce

1. Add a contact whose only `TEL` value is non-numeric, e.g.:
   ```
   TEL;TYPE=WORK,PAGER:Amazon
   ```
   (this is a real, if malformed, vCard field some companies use for their
   "add to contacts" SMS shortcodes — it also happens intentionally, as a way
   to get a nice name/logo for SMS from alphanumeric sender IDs).
2. Sync contacts so this entry is cached with an empty-string number.
3. Receive an SMS (or otherwise call `query({number: '+33612345678'})`) from a
   real contact whose entry happens to be ordered *after* the alphanumeric one
   in the contacts cache.
4. The SMS/call is displayed under the wrong contact's name.

### Expected behavior

An SMS or call from a real phone number should always be matched to the
  contact that actually owns that number. It must never be matched to an
  unrelated contact whose vCard has a non-numeric/empty TEL value (e.g. a
  placeholder like "Amazon" used as an alphanumeric SMS sender-ID label),
  regardless of the order in which contacts happen to be cached/iterated.

### GSConnect version

Zorin Connect 1.33.4 - /e/OS 3.0.4 community - Android 13

### Installed from

OS package manager

### GNOME Shell version

46.0 (46.0-0ubuntu6~24.04.14+zorin1)

Zorin OS 18.1 / (Ubuntu 24.04 noble)

### Paired device(s)

Xiaomi Redmi 8 (Mi439)

### KDE Connect app version

Zorin Connect 71

### Plugin(s)

_No response_

### Support log

```plain text

```

### Screenshots

_No response_

### Notes

_No response_

ferdnyc:
Now that `String.prototype.equalsPhoneNumber()` is fixed to perform comparisons the way we want, replace all of the ad-hoc comparison code with calls to `.equalsPhoneNumber()`. This eliminates a lot of explicit temporary `.toPhoneNumber()` creation (because it's implicit in the call to `.equalsPhoneNumber()`), and it means that comparisons are automatically protected from matching when one side is empty (since `.equalsPhoneNumber()` always checks for that condition).

Along the way, add a few `?.` property accesses to better protect against possible null dereferences.

Fixes #2176

### DIFF
--- a/src/service/components/contacts.js
+++ b/src/service/components/contacts.js
@@ -362,15 +362,12 @@ const Store = GObject.registerClass({
         // First look for an existing contact by number
         const contacts = this.contacts;
         const matches = [];
-        const qnumber = query.number.toPhoneNumber();
 
         for (let i = 0, len = contacts.length; i < len; i++) {
             const contact = contacts[i];
 
             for (const num of contact.numbers) {
-                const cnumber = num.value.toPhoneNumber();
-
-                if (qnumber.endsWith(cnumber) || cnumber.endsWith(qnumber)) {
+                if (query.number.equalsPhoneNumber(num.value)) {
                     // If no query name or exact match, return immediately
                     if (!query.name || query.name === contact.name)
                         return contact;
--- a/src/service/plugins/sms.js
+++ b/src/service/plugins/sms.js
@@ -493,15 +493,10 @@ const SMSPlugin = GObject.registerClass({
     }
 
     _threadHasAddress(thread, addressObj) {
-        const number = addressObj.address.toPhoneNumber();
-
         for (const taddressObj of thread[0].addresses) {
-            const tnumber = taddressObj.address.toPhoneNumber();
-
-            if (number.endsWith(tnumber) || tnumber.endsWith(number))
+            if (addressObj.address?.equalsPhoneNumber(taddressObj.address))
                 return true;
         }
-
         return false;
     }
 
--- a/src/service/ui/contacts.js
+++ b/src/service/ui/contacts.js
@@ -167,15 +167,10 @@ function getNumberTypeLabel(type) {
  * @returns {string} A (possibly) better display number for the address
  */
 export function getDisplayNumber(contact, address) {
-    const number = address.toPhoneNumber();
-
     for (const contactNumber of contact.numbers) {
-        const cnumber = contactNumber.value.toPhoneNumber();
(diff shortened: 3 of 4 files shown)


daniellandau (approved):
LGTM

PR #2191, opened 2026-08-17
notification: suppress duplicate re-sends from reconnecting devices

YappLeCunt:
## Summary

Mobile OSes with aggressive battery management (ColorOS, MIUI, etc.) regularly kill background services. When GSConnect's service on the phone is killed and restarts (or the connection drops and reconnects), the device re-sends its **complete list of active notifications** — so the same notifications keep re-appearing on the desktop over and over, even if they were already shown (or dismissed).

This adds a small cache to the notification plugin that suppresses these re-send bursts **only during the reconnection handshake**, so legitimate notifications are never swallowed.

## Changes

- `src/service/plugins/notification.js`
  - Per-plugin `_recentNotifications` Map (initialized in `_init`), keyed by the packet's `id` + `appName` + `title` + `text` + `requestReplyId`
  - `connected()` records `_lastConnectedAt`; `_isRecentNotification()` only treats a re-send as a duplicate if it arrives within `RECENT_NOTIFICATION_BURST_WINDOW` (60s) of a (re)connection
  - `RECENT_NOTIFICATION_TTL` (30 min) and `RECENT_NOTIFICATION_MAX` (200) bound cache size; expired entries are purged and oldest entries evicted
  - `debug()` line when a duplicate is suppressed (gated by the existing `debug` setting)
- `installed-tests/suites/plugins/testNotificationPlugin.js`
  - Three new cases: identical re-send inside the burst window is suppressed; changed content is shown again; identical re-send **outside** the burst window is shown again

## Design notes

**Why is `id` part of the key?** On Android, notification IDs are assigned by the *source app*, not by GSConnect's relay, so they are stable across GSConnect service restarts — the reconnect re-send carries the same `id`, which is exactly what gets deduped. Including `id` deliberately means a genuinely **new** notification instance with identical content (a recurring alarm, an app that re-posts the same text) is still shown: the key only collapses re-sends of the *same instance*.

**Why gate on the reconnection burst rather than a bare TTL?** The desktop uses the packet `id` as the Gio notification ID, so same-ID re-sends are normally collapsed by the Shell — but that breaks when (a) the user has dismissed the notification (a re-send re-pops it), or (b) the Shell's in-place update doesn't apply. Content-keying alone risks suppressing *legitimate* repeats (recurring alarms, repeated reminders) within the window. Restricting suppression to the 60s following a (re)connection means:

- the reconnect-burst spam (the actual bug) is silenced,
- a notification re-posted while the connection is stable is always shown,
- the "refresh on suppression" behavior only extends across actual reconnect churn — once the device stays connected, the window closes and nothing is hidden.

`RECENT_NOTIFICATION_TTL` is now purely a memory bound for the cache, not a behavioral window, so it is intentionally not exposed as a setting; happy to add one if maintainers prefer.

## Verification

- The jasmine suite in this PR covers the behavior; CI runs ESLint and the full installed-tests suite
- Logic additionally exercised during development with a standalone 45-assertion harness (TTL expiry, eviction, reply-ID variants, burst-window boundaries, reconnect-churn refresh); not part of the repo
- Live reproduction on a ColorOS device (realme GT 7T): after patching the extension and toggling airplane mode, a reconnect burst of 17 re-sent notifications (Facebook, Discord, Notion Calendar, ChatGPT, Telegram, ...) was suppressed in the journal; the first cycle was shown as expected

### DIFF
--- a/installed-tests/suites/plugins/testNotificationPlugin.js
+++ b/installed-tests/suites/plugins/testNotificationPlugin.js
@@ -281,6 +281,12 @@ describe('The notification plugin', function () {
     });
 
     describe('suppresses duplicate re-sends', function () {
+        beforeEach(function () {
+            // Simulate a just-reconnected device so the suppression
+            // burst window is active
+            remotePlugin._lastConnectedAt = Date.now();
+        });
+
         it('shows a notification only once when re-sent identically',
             async function () {
                 const packet = {
@@ -308,6 +314,24 @@ describe('The notification plugin', function () {
                 expect(remotePlugin.device.showNotification.calls.count()).toBe(1);
 
                 packet.body = {...Notifications.withoutIcon, text: 'Changed Body'};
+
+                await remotePlugin._receiveNotification(packet);
+                expect(remotePlugin.device.showNotification.calls.count()).toBe(2);
+            });
+
+        it('does not suppress re-sends outside the reconnection burst',
+            async function () {
+                const packet = {
+                    type: 'kdeconnect.notification',
+                    body: Notifications.withoutIcon,
+                };
+
+                await remotePlugin._receiveNotification(packet);
+                expect(remotePlugin.device.showNotification.calls.count()).toBe(1);
+
+                // The burst window has elapsed; the same notification is a
+                // new event now and must be shown again
+                remotePlugin._lastConnectedAt = Date.now() - (60 * 1000 + 1000);
 
                 await remotePlugin._receiveNotification(packet);
                 expect(remotePlugin.device.showNotification.calls.count()).toBe(2);
--- a/installed-tests/suites/plugins/testNotificationPlugin.js
+++ b/installed-tests/suites/plugins/testNotificationPlugin.js
@@ -280,6 +280,40 @@ describe('The notification plugin', function () {
         expect(remotePlugin._handleNotificationRequest).toHaveBeenCalled();
     });
 
+    describe('suppresses duplicate re-sends', function () {
+        it('shows a notification only once when re-sent identically',
+            async function () {
+                const packet = {
+                    type: 'kdeconnect.notification',
+                    body: Notifications.withoutIcon,
+                };
+
+                await remotePlugin._receiveNotification(packet);
+                expect(remotePlugin.device.showNotification.calls.count()).toBe(1);
+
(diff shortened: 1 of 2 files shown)


alkaiyali:
Updated in bff25b3

**1. Suppression is now gated to reconnection bursts.** Duplicates are only suppressed within a 60s window after a device (re)connection (new `RECENT_NOTIFICATION_BURST_WINDOW`), since the full notification-list re-send happens as part of the reconnection handshake. Notifications re-posted while the connection is stable are never suppressed — recurring alarms, repeated reminders and re-posted messages all still show. New jasmine case covers the outside-burst path.

**2. `id` stays in the key, deliberately.** Android notification IDs are assigned by the source apps and remain stable across GSConnect service restarts (verified live: the reconnect burst carried stable IDs and was fully deduped), so the key collapses re-sends of the *same instance* while a genuinely new instance with identical content (new id) still displays.

**3. TTL is now a memory bound, not a behavior window.** The refresh-on-suppression only extends across actual reconnect churn; once the device stays connected, the window closes and nothing is hidden.

CI workflows are waiting on the fork-PR approval gate; Merge conflicts check already passed.

alkaiyali:
Author's confirmation on the reconnect-path edge case:

**Covered — kill/restart and network-drop scenarios.** A killed phone-side process closes its sockets at the OS level; the desktop's `_readLoop` hits EOF and goes through `setChannel(null)` → disconnect → the phone's fresh connection then fires `connected()` on the plugin, arming the burst window. This is exactly the path exercised in the live reproduction (the cycle-2 suppression at 11:21:42 required `connected()` to have fired on the new channel).

**Narrow gap — acknowledged.** There is one path where the window stays closed: `setChannel()` (device.js) replaces a *live* channel with a new one before the stale socket's EOF is processed (fast-restart race or half-open socket after sudden WiFi loss). In that case `connected === !!this.channel` is unchanged and the early return at device.js:355-356 skips `_triggerPlugins()`, so no burst window is armed and a re-sent list would slip through. The clean fix is to treat a channel swap as a reconnection (fire `_triggerPlugins()` on swap) — that's a core device.js behavior change with blast radius across all plugins, so I'd rather land it as its own PR than bundle it into this notification fix. Happy to open that follow-up if maintainers agree with the semantics.

PR #2192, opened 2026-08-18
GNOME 51

DaPigGuy:
Not much broke in GNOME 51 except `GjsPrivate.DBusImplementation` was marked as a final type in GJS, so no more subclassing. Took this as an opportunity to switch to the `wrapJSObject` convenience helper from GJS instead of us handling method calls and property get/set ourselves.

https://gjs.guide/guides/gio/dbus.html#services
https://gitlab.gnome.org/GNOME/gjs/-/blob/master/modules/core/overrides/Gio.js?blame=1#L437

`key-press-event`, `button-press-event` were also deprecated and replaced with `Clutter.KeyController` / `Clutter.ClickGesture`. Trying to test the KeyController changes led to finding a lot of issues with the Commands submenu which I've included fixes for.

https://gjs.guide/extensions/upgrading/gnome-shell-51.html#clutter-controllers


daniellandau (approved):
Thank you! I tested on 46, 50 and 51, hopefully the versions in between also work.

